Issue #64, opened 2013-12-09
Allow setting up async methods easily


jdom:
It would be nice to have better usability for methods that return Task.

For example, an option for doing something like:

myMock.Setup(x => x.ReadAsync()).ThrowsAsync(new Exception());

which should return a faulted task with that exception.

Also, for methods that return Task that are not explicitly set up, it would be good that instead of defaulting to return a null Task, that they return a completed Task (or completed Task<T>) instead.

Didn't consider the best way to address this yet, but currently it can be a little cumbersome especially when mocking code where >70% of its methods are async.

stakx:
Closing this issue since all that it asked for has been implemented since.

PR #66, opened 2013-12-10
Avoid NullReferenceException with async methods

alextercete:
When using async methods and loose Mock Behavior, calls to `Task` methods give `NullReferenceException`, since `default(Task)` is `null`. Calls to `Task.Wait()` and `Task<T>.Result` should follow the loose behavior: do nothing and return the default value for `T`, respectively.

Related to #64.

### DIFF
--- a/Source/EmptyDefaultValueProvider.cs
+++ b/Source/EmptyDefaultValueProvider.cs
@@ -43,7 +43,9 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


iskiselev:
TPL presents in Silverlight 5. So, if Moq will provide separate SL4 and SL5 version or drop SL4 support, it can use auto async proxy creation for Silverlight.

kzu:
I'm happy to drop SL4 unless there is a massive outcry to keep it ;)

/kzu from mobile

PR #81, opened 2013-12-24
Proxy generator added inside moq

marianor:
This is the proxy generator that I was developing some time ago. It can save the dependency of Castle project. All tests were "Passed".  This version supports override ToString(), GetHashCode() and Equals().


kzu:
Hey @marianor, sorry for taking so long to answer :(

I think the long-term solution is more like moq/moq.proxy, compile-time generated proxies, so that they also work on Xamarin :)

PR #104, opened 2014-03-27
Add support for SetupAnyArg and VerifyAnyArg

olegoga:
We use Moq a lot in our team and I find that I spend a lot of time writing the correct It.IsAny<T>() parameters for Verify and Setup methods. I really wanted a simple way match any parameters and created two methods SetupAnyArg and VerifyAnyArg. When they are used, the underlying logic ignores any expression parameters and just matches the method by the signature. 

I realize that my current implementation might not be up to the standards of the Moq code-base and maybe I missed some important design principle. If you are interested in integrating this feature, give me some comments and I will gladly make the adjustments.

### DIFF
--- a/Source/ConditionalContext.cs
+++ b/Source/ConditionalContext.cs
@@ -59,12 +59,12 @@ namespace Moq
 
 		public ISetup<T> Setup(Expression<Action<T>> expression)
 		{
-			return Mock.Setup<T>(mock, expression, this.condition);
+			return Mock.Setup<T>(mock, expression, this.condition, false);
 		}
 
 		public ISetup<T, TResult> Setup<TResult>(Expression<Func<T, TResult>> expression)
 		{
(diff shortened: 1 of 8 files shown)


7Pass:
I've love to have this too. I implemented it as an extension method (https://gist.github.com/7Pass/1c6b329e85ca29071f42). I've tried to create a pull request but could not get the build to work, some error with xdoc:
```
Mock.Generic.cs(87,7): error CS1735: XML comment on 'Mock<T>.Mock(params object[])' has a typeparamref tag for 'T', but there is no type parameter by that name [Z:\moq4\Source\Moq.csproj]
```

kzu:
I think it's more idiomatic to try to implement this as an extension method on the ISetup interface that fetches the last setup and replaces the matchers. Something like:

```
mock.Setup(x => x.Execute(null, null, null)).IgnoreArgs().Verifable();
```

PR #109, opened 2014-05-11
Expose IDefaultValueProvider

jamesfoster:
I've added the ability for users to define their own default value strategy without affecting the external interface. There should be no braking changes, just update and recompile. I wanted to rename a few properties to better reflect their function but decided to stick with not braking the API.

This will be great for things like http://github.com/AutoFixture/AutoFixture where the defaults can be set to reasonable random values.


Source/AsInterface.cs
```diff
@@ -83,7 +86,7 @@ public override bool CallBase
 			set { this.owner.CallBase = value; }
 		}
 
-		public override DefaultValue DefaultValue
+		public override Func<IDefaultValueProvider> DefaultValue
```
kzu: Why a Func and not an IDefaultValueProvider directly?

Source/DefaultValue.cs
```diff
@@ -40,18 +40,20 @@
...
-	public enum DefaultValue
+	public static class DefaultValue
 	{
 		/// <summary>
 		/// Default behavior, which generates empty values for 
 		/// value types (i.e. default(int)), empty array and 
 		/// enumerables, and nulls for all other reference types.
 		/// </summary>
-		Empty,
+		public static readonly Func<IDefaultValueProvider> Empty = () => new EmptyDefaultValueProvider();
```
kzu: This would look better in intellisense as a static property, initialized from the static ctor

Source/IDefaultValueProvider.cs
```diff
@@ -45,12 +45,12 @@ namespace Moq
 {
 	/// <include file='IDefaultValueProvider.xdoc' path='docs/doc[@for="IDefaultValueProvider"]/*'/>
 	[EditorBrowsable(EditorBrowsableState.Advanced)]
```
kzu: Since EmptyDefaultValueProvider is already public, I think it would be fine to make this interface discoverable too.

Source/IMock.cs
```diff
@@ -20,6 +22,6 @@ public interface IMock<out T> where T : class
 		bool CallBase { get; set; }
 
 		/// <include file='Mock.xdoc' path='docs/doc[@for="Mock.DefaultValue"]/*'/>
-		DefaultValue DefaultValue { get; set; }
+		Func<IDefaultValueProvider> DefaultValue { get; set; }
```
kzu: Wondering why this isn't an `IDefaultValueProvider` too...

Source/Mock.cs
```diff
@@ -169,7 +169,11 @@ private object GetObject()
 
         internal virtual Interceptor Interceptor { get; set; }
 
-        internal virtual Dictionary<MethodInfo, Mock> InnerMocks { get; private set; }
+        /// <summary>
+        /// Contains a list of all dynamically created mocks when the DefaultValue is set to DefaultValue.Mock
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public virtual Dictionary<MethodInfo, Mock> InnerMocks { get; private set; }
```
kzu: Not sure this needs to be made public, does it?

kzu:
This looks good, aside from a couple minor comments, could you rebase on top of master?

Thanks

stakx:
There are a few things that must happen before this can be looked at:

1. We need to normalize whitespace (see #364). Here, I tried to rebase so I could look into the proposed changes, but the mixed tabs & spaces situation leads to very large diffs; it's hard to actually see what's changing.

2. Like just mentioned, this needs to be rebased before it can be judged any further. Several things have changed since this PR was last edited (e.g. the `MockFactory` &rarr; `MockRepository` rename, or the replacement of `Dictionary` with `ConcurrentDictionary`, to name just a few).

3. Like mentioned in #310, I'm not sure whether it's wise to change the public API (turning `DefaultValue` from an `enum` into a `static class`).

Source/AsInterface.cs
stakx: I suppose the idea of using `Func` here might be lazy instantiation, in case that the provider is a "heavy" object. Given that the default value provider is probably used all the time, I'm not positively sure that lazy instantiation would have practical benefits, though.

stakx:
This PR has a few issues:

1. It introduces a breaking change.

    * `enum DefaultValue` is redefined as a `static class`.

2. At least when rebased onto `develop`, it no longer compiles:

    * `SerializableTypesValueProvider` is lacking an implementation for `object IDefaultValueProvider.ProvideDefault(MethodInfo member, Mock owner)`.

    * The `Mock.GetInitialValue` method must be changed since `IDefaultValueProvider.ProvideDefault` now has an additional parameter. Same in `SerializableTypesValueProvider.ProvideDefault`.

    * There's a missing `null` check in `MockDefaultValueProvider.ProvideDefault` (this could be called with `owner == null`).

3. Once the above is fixed, one test (`StubsAllHierarchy`) fails. Apparently Moq can no longer stub whole object graphs. That would be a major problem.

stakx:
Superseded by #366. Sorry for closing this, I really did give it a try (see [stakx/moq4:jamesfoster-defaultvalueprovider](https://github.com/stakx/moq4/tree/512d3ca9cdc24116e3452a7625dbb9894478a106)), but couldn't get everything to work. It's far easier to reimplement the feature cleanly on top of the latest version of Moq.

stakx:
@jamesfoster: It's been a while. Just in case you're still interested, I just merged #533, which exposes a new abstract `DefaultValueProvider` base class and a `Mock[Repository].DefaultValueProvider` property. This will be in the next release of Moq (4.8).

stakx:
@jamesfoster - A pre-release version of Moq (`4.8.0-rc1`) is now available on NuGet. It includes support for custom `DefaultValueProvider`s. Feel free to test away.

PR #134, opened 2014-10-07
Issue 135:	Fix equality check bug in ExpressionKey

kostadinmarinov:
The equality check in ExpressionKey is wrong. The check must be conjunction (&), but not disjunction(|), because then the values don't matter in that check and they are just skipped. So, invocations:

aMock.Setup(m => m.Foo(i1, i2));
aMock.Setup(m => m.Foo(i2, i1));

have the same expression key, but this is not correct.


kzu:
If someone volunteers some unit tests that assert the new behavior, I'm happy to merge this too :)

kostadinmarinov:
Done

kzu:
Damn, now it needs rebasing on top of master :(

stakx:
There's one problem with the added unit test: it passes even when the suggested code change is not applied. Therefore it doesn't really tell us anything.

I also noted that the test appears almost identical in its intent to [this other existing test](https://github.com/kostadinmarinov/moq4/blob/9ad408355e381bbde7ec68c7ad1b9933f9a04c52/UnitTests/Regressions/IssueReportsFixture.cs#L1071-L1086).

stakx:
I've rebased and cleaned up this PR. Please see #363 instead.

Issue #128, opened 2014-08-19
NotImplementedException DynamicProxy2 error when CallBase = true while mocking interfaces


MichaelLogutov:
This code:

```
[TestClass]
public class Tests
{
	public interface IDataService
	{
		IList<int> GetData ();
	}


	[TestMethod]
	public void Test ()
	{
		// arrange
		var data_service_mock = new Mock<IDataService> ();
		data_service_mock.DefaultValue = DefaultValue.Mock;
		data_service_mock.CallBase = true;
			

		// act
		var data = data_service_mock.Object.GetData ();
		var result = data.Sum ();


		// assert
		Assert.IsTrue (result == 0);
	}
}
```

Throws:

```
Test method Bug.Tests.Test threw exception: 
System.NotImplementedException: This is a DynamicProxy2 error: The interceptor attempted to 'Proceed' for method 'System.Collections.Generic.IEnumerator`1[System.Int32] GetEnumerator()' which has no target. When calling method without target there is no implementation to 'proceed' to and it is the responsibility of the interceptor to mimic the implementation (set return value, out arguments etc)
    at Castle.DynamicProxy.AbstractInvocation.ThrowOnNoTarget()
   at Castle.DynamicProxy.Internal.CompositionInvocation.EnsureValidTarget()
   at Castle.Proxies.Invocations.IEnumerable`1_GetEnumerator.InvokeMethodOnTarget()
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Moq.Proxy.CastleProxyFactory.CallContext.InvokeBase()
   at Moq.InvokeBase.HandleIntercept(ICallContext invocation, InterceptorContext ctx, CurrentInterceptContext localctx)
   at Moq.Interceptor.Intercept(ICallContext invocation)
   at Moq.Proxy.CastleProxyFactory.Interceptor.Intercept(IInvocation invocation)
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.IList`1Proxy.GetEnumerator()
   at System.Linq.Enumerable.Sum(IEnumerable`1 source)
   at Bug.Tests.Test() in Tests.cs: line 28
```

Moq v 4.2.1408.0717

The exception is not thrown when CallBase set to false which is strange because it should has no side effects when mocking interfaces.

It seems that this issue started to appear beginning version 4.2.1408.619.

LeonidLevin:
This seems to have been fixed.
Please, close the issue!

PR #145, opened 2014-11-26
Fix: #128 NotImplementedException DynamicProxy2 error when CallBase = true 

rjasica:
Fix: #128 NotImplementedException DynamicProxy2 error when CallBase = true  while mocking interfaces


MichaelLogutov:
This is a serious issue. I don't understand how it's been unsolved for so long. Are there no collaborators anymore?

kzu:
It's not a serious issue, it's a user error. Why would someone specify CallBase on an interface?

memark:
But never the less it has now been fixed by accepting the merge, right? That's probably the thing the other posters and myself care most about. :)

kzu:
merged and published to nuget.org ;)


/kzu

memark:
Great, thanks!

PR #155, opened 2015-02-16
Added MockBehavior overload to Mock.Of<T>()

JohanLarsson:
Removed MockBehavior parameter from InterceptorContext ctor + tests.

Added overloads with MockBehavior to Mock.Of<T>(...) I could not figure out what problem CreateMockQuery et al. are solving. Wrote a couple of overloads with MockBehaviour. Threw in a couple of tests but don't know what all relevant testcases are.

### DIFF
--- a/Source/IInterceptStrategy.cs
+++ b/Source/IInterceptStrategy.cs
@@ -28,21 +28,26 @@ namespace Moq
 
 	internal class InterceptorContext
 	{
-		private Dictionary<string, List<Delegate>> invocationLists = new Dictionary<string, List<Delegate>>();
-		private List<ICallContext> actualInvocations = new List<ICallContext>();
-		private List<IProxyCall> orderedCalls = new List<IProxyCall>();
+		private readonly Dictionary<string, List<Delegate>> invocationLists = new Dictionary<string, List<Delegate>>();
+		private readonly List<ICallContext> actualInvocations = new List<ICallContext>();
+		private readonly List<IProxyCall> orderedCalls = new List<IProxyCall>();
 
-		public InterceptorContext(Mock Mock, Type targetType, MockBehavior behavior)
+		public InterceptorContext(Mock mock, Type targetType)
 		{
-			this.Behavior = behavior;
-			this.Mock = Mock;
+			this.Mock = mock;
 			this.TargetType = targetType;
 		}
-		public Mock Mock { get; private set; }
-		public Type TargetType { get; private set; }
-		public MockBehavior Behavior { get; private set; }
 		
-		#region InvocationLists
+        public Mock Mock { get; private set; }
+		
+        public Type TargetType { get; private set; }
+
+	    public MockBehavior Behavior
+	    {
+	        get { return Mock.Behavior; }
+	    }
+
+	    #region InvocationLists
 		internal IEnumerable<Delegate> GetInvocationList(EventInfo ev)
 		{
 			lock (invocationLists)
--- a/Source/Linq/Mock.cs
+++ b/Source/Linq/Mock.cs
@@ -48,28 +48,54 @@ using Moq.Linq;
 
 namespace Moq
 {
-	public partial class Mock
-	{
-		/// <summary>
-		/// Creates an mock object of the indicated type.
-		/// </summary>
-		/// <typeparam name="T">The type of the mocked object.</typeparam>
-		/// <returns>The mocked object created.</returns>
-		public static T Of<T>() where T : class
-		{
-			return Mocks.CreateMockQuery<T>().First<T>();
-		}
+    public partial class Mock
+    {
+        /// <summary>
+        /// Creates an mock object of the indicated type.
+        /// </summary>
+        /// <typeparam name="T">The type of the mocked object.</typeparam>
+        /// <returns>The mocked object created.</returns>
+        public static T Of<T>() where T : class
+        {
+            return Mocks.CreateMockQuery<T>().First<T>();
+        }
 
-		/// <summary>
-		/// Creates an mock object of the indicated type.
-		/// </summary>
-		/// <param name="predicate">The predicate with the specification of how the mocked object should behave.</param>
-		/// <typeparam name="T">The type of the mocked object.</typeparam>
-		/// <returns>The mocked object created.</returns>
-		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
-		public static T Of<T>(Expression<Func<T, bool>> predicate) where T : class
-		{
-			return Mocks.CreateMockQuery<T>().First<T>(predicate);
-		}
-	}
+        /// <summary>
+        /// Creates an mock object of the indicated type.
+        /// </summary>
+        /// <param name="predicate">The predicate with the specification of how the mocked object should behave.</param>
+        /// <typeparam name="T">The type of the mocked object.</typeparam>
+        /// <returns>The mocked object created.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
+        public static T Of<T>(Expression<Func<T, bool>> predicate) where T : class
+        {
+            return Mocks.CreateMockQuery<T>().First<T>(predicate);
+        }
+
+        /// <summary>
+        /// Creates an mock object of the indicated type.
+        /// </summary>
+        /// <param name="predicate">The predicate with the specification of how the mocked object should behave.</param>
+        /// <param name="behavior">Behavior of the mock.</param>
+        /// <typeparam name="T">The type of the mocked object.</typeparam>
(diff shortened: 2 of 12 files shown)


JohanLarsson:
Not sure what the status is on this. Feedback would be nice in case something needs fixing.

kzu:
A Mock.Of<T> without an expression is just a dumb helper over new Mock<T>().Object. Not very useful.

Mock.Of<T> enables non other than [Linq to Mocks](https://github.com/Moq/moq4/wiki/Quickstart#linq-to-mocks), a unique and super powerful feature.

JohanLarsson:
Not sure I understand. You can do:
    var foo = Mock.Of<IFoo>(x => x.BoolProperty1, MockBehavior.Strict);
https://github.com/Moq/moq4/pull/155/files#diff-a01c95ad546c3bcacfcc8fef5c3936aeR38

UnitTests/Linq/BehaviorFixture.cs
```diff
@@ -0,0 +1,102 @@
...
+        public void NoQuery_Loose()
+        {
+            var target = Mock.Of<IFoo>(MockBehavior.Loose);
+            Assert.Equal(MockBehavior.Loose, Mock.Get(target).Behavior);
+        }
+
+        [Fact]
+        public void WithQuery_Strict()
+        {
+            var target = Mock.Of<IFoo>(x => x.BoolProperty1, MockBehavior.Strict);
```
kzu: Test should assign a value in the expression:

x => x.BoolProperty1 == true

And verify that it is indeed true.

kzu:
Tests should also verify that the strict and loose behaviors remain (i.e. Throw on non setup member, and default value, respectively)

JohanLarsson:
Sure, I'll add more tests & asserts.

JohanLarsson:
I finally got some time to do this. Came across an issue:
```
[Test]
public void ShouldThrowWhenStrictRight()
{
    var mock = new Mock<Dto>(MockBehavior.Strict);
    Assert.IsNull(mock.Object.Value); // This should throw right?
    mock.Object.Value = "test"; // This should throw right?
    Assert.AreEqual("test", mock.Object.Value); // This should throw right?
}

public class Dto
{
	public string Value { get; set; }
}
```
Suggested fix:
```
public Mock(MockBehavior behavior, params object[] args)
{
    if (behavior == MockBehavior.Strict && !typeof(T).IsInterface)
    {
        var message = string.Format("Cannot use MockBehavior.Strict if T is not an interface. T was: {0}", typeof(T).FullName);
        throw new ArgumentException(message, "behavior");
    }
    ...
```
The fix breaks a couple of tests, want to check before touching too much.

JohanLarsson:
Also, while Mock.Of<IFoo>(MockBehavior.Strict) is trivial I don't agree that it is not useful.

kzu:
MockStrict can totally be used with classes, not only interfaces. I'd remove that "fix" from the rest. 

The error is in the test: your Value property isn't virtual, therefore, it's not being intercepted by Moq.

JohanLarsson:
Imo it is not very elegant that a strict mock can be touched without throwing but I'll remove it as it is a breaking change. Should I delete this and make new pull requests? I introduced some noise in the diffs by formatting.

kzu:
@JohanLarsson I appreciate the time you're putting into this, but the PR is virtually unmergable with all the noise from formatting changes that are totally unrelated to the actual new behavior :(

You don't need to delete it (otherwise, the whole conversation will go away). You can just make more changes locally, and force-push to your fork. That will automatically update this PR

kzu:
> Imo it is not very elegant that a strict mock can be touched without throwing but I'll remove it as it is a breaking change.

The fact of the matter is that we cannot know for sure if that's a user error or not. If something isn't virtual, we have no way of knowing. If you revert that behavior, you'll certainly break a ton of users that are currently using strict mocks with abstract classes or even non-abstract classes where they know what members to make virtual for testing/extensibility.

JohanLarsson:
It is probably less work to make new clean forks. Thanks for taking the time to read it. I'll clean it up some way.

kzu:
Hey @JohanLarsson, did you end up getting around to cleaning this up? this looks like a useful feature.

You don't need to keep the fork/branch around in order for this conversation to remain. But a new PR will not have this history. If you recreate it locally, and force-push to the same branch in your fork, you can have the best of both worlds: a clean PR/branch, and preserve the discussion ;). But it's totally up to you, whichever is easier. I don't mind either way.

JohanLarsson:
I meant to do it but forgot about it, been reminded at times when I have missed the feature but been busy with other stuff then. 

Had a quick look at the diff now and have a couple of questions/comments:
1. Are tabs and not spaces used in the rest of the code base? If so I guess I'll have to change all rows. If not maybe we can review the diff line-by-line and merge it with spaces, reviewing a line that only differs by indentation should be pretty fast.
2. There are a couple of typos fixed in comments and internal stuff. Those should of course ideally be separate PRs.
3. The two commits should be squashed to one.
I'll try to fix it this weekend.

JohanLarsson:
I squashed them now, maybe we can have a look at the diff?

Source/IInterceptStrategy.cs
```diff
@@ -28,21 +28,26 @@ internal interface IInterceptStrategy
 
 	internal class InterceptorContext
 	{
-		private Dictionary<string, List<Delegate>> invocationLists = new Dictionary<string, List<Delegate>>();
-		private List<ICallContext> actualInvocations = new List<ICallContext>();
-		private List<IProxyCall> orderedCalls = new List<IProxyCall>();
+		private readonly Dictionary<string, List<Delegate>> invocationLists = new Dictionary<string, List<Delegate>>();
+		private readonly List<ICallContext> actualInvocations = new List<ICallContext>();
+		private readonly List<IProxyCall> orderedCalls = new List<IProxyCall>();
```
JohanLarsson: Changing to readonly is perhaps noisy.

Source/Linq/Mock.cs
```diff
@@ -48,28 +48,54 @@
...
+    {
+        /// <summary>
+        /// Creates an mock object of the indicated type.
+        /// </summary>
+        /// <typeparam name="T">The type of the mocked object.</typeparam>
+        /// <returns>The mocked object created.</returns>
+        public static T Of<T>() where T : class
+        {
+            return Mocks.CreateMockQuery<T>().First<T>();
+        }
```
JohanLarsson: Would be nice if the github diff tool had an ignore whitespace option. Do we really need tabs here badly?

Source/Linq/MockSetupsBuilder.cs
```diff
@@ -208,7 +208,7 @@ private static Expression ConvertToSetupProperty(Expression targetObject, Expres
 			// which also allows the use of this querying capability against plain DTO even 
 			// if their properties are not virtual.
 			var setPropertyMethod = typeof(Mocks)
-				.GetMethod("SetPropery", BindingFlags.Static | BindingFlags.NonPublic)
+				.GetMethod("SetProperty", BindingFlags.Static | BindingFlags.NonPublic)
```
JohanLarsson: Fixed typo on internal method.

MpDzik: I think this fix should be in a separate changeset.

Source/IInterceptStrategy.cs
```diff
@@ -28,21 +28,26 @@ internal interface IInterceptStrategy
...
-			this.Mock = Mock;
+			this.Mock = mock;
 			this.TargetType = targetType;
 		}
-		public Mock Mock { get; private set; }
-		public Type TargetType { get; private set; }
-		public MockBehavior Behavior { get; private set; }
 		
-		#region InvocationLists
+        public Mock Mock { get; private set; }
```
kzu: This is all noise :(. I have provided a .editorconfig file that will be picked up by the EditorConfig extension in VS (available from the VS gallery) to make this easier.

Source/Linq/MockSetupsBuilder.cs
kzu: :+1: this is a minor improvement independent of this.

kzu:
Just to resume the conversation here: I like this feature, I think it will be useful.

If you could adjust the spacing stuff (it's currently very hard to read) by using the provided EditorConfig support, I'll merge it.

It needs rebasing on top of master too, I'm afraid.

Thanks for taking the time to look at this!

stakx:
Since this PR currently has merge conflicts and needs to be rebased anyhow, would it be much trouble retargeting this PR at the (brand-new) `develop` branch (which can be done via the *Edit* button on the right side of the PR title) and also rebasing onto `develop` instead of `master`?

JohanLarsson:
I'll close it and try to find time to split it up into more focused pull requests.

stakx:
I'm looking a the whitespace issue now (should take about an hour or so), I'll convert all indentation back to tabs (as it should be per `.editorconfig`), so if you make sure to use tab indentation you shouldn't have any problems with noisy diffs next time you open a PR.

Issue #90, opened 2014-02-07
NullReferenceException when mocked method has a nullable value type argument, but It.IsAny constraint is not nullable


ashmind:
Example:

    public interface ISomeInterface {
        void DoSomething(Guid? guid);
    }

    mock.Setup(x => x.DoSomething(It.IsAny<Guid>()))
    mock.Object.DoSomething(null);

Result:

    System.NullReferenceException : Object reference not set to an instance of an object.
        at Moq.Match`1.Matches(Object value)
        at Moq.MethodCall.Matches(ICallContext call)
        at System.Linq.Enumerable.LastOrDefault(IEnumerable`1 source, Func`2 predicate)
        at Moq.ExtractProxyCall.HandleIntercept(ICallContext invocation, InterceptStrategyContext ctx)
        at Moq.Interceptor.Intercept(ICallContext invocation)
        at Castle.DynamicProxy.AbstractInvocation.Proceed()
        at Castle.Proxies.ISomeInterfaceProxy.DoSomething(Nullable`1 guid)

anonymous112233:
This seem to back as of latest v. 4.2.1506.2016

yrshaikh:
For those who are coming here in 2019. Try this

`mock.Setup(x => x.DoSomething(It.IsAny<Guid?>()))`

stakx:
@yrshaikh: The difference between `It.IsAny<Guid>()` or `It.IsAny<Guid?>()` is whether or not `null` arguments should be matched. Either way, in up-to-date versions of Moq, you should no longer see a `NullReferenceException` (which is what this issue was about).

PR #160, opened 2015-03-10
Fix NullReferenceException when passing null to a nullable argument but trying to match it with a non-nullable IsAny.

benjamin-hodgson:
This fixes issue #90.

The problem was that `Matches()` has to cast its input to the type parameter `T` in order to call the `Condition` function.
Since `T` is a reference type, it does not accept nulls and throws a `NullReferenceException` at the site of the cast.

We get around this by treating value types as a special case inside `Matches`. If `T` is a value type and the input is null, it clearly doesn't match (since value types are not nullable).

### DIFF
--- a/Source/Match.cs
+++ b/Source/Match.cs
@@ -128,7 +128,18 @@ namespace Moq
 			{
 				return false;
 			}
-
+			if (value == null && typeof(T).IsValueType)
+			{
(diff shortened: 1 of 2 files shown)


anonymous112233:
This seem to back as of latest v. 4.2.1506.2016

PR #179, opened 2015-06-17
Fixed infinite loop when calling SetupAllPropertis

vik-borisov:
Fixed infinite loop when calling SetupAllPropertis for entities with the hierarchy. Written test StubsAllCyrcularDependency


vik-borisov:
As you can see this a fix in nuget repository?

kzu:
pushed!

vik-borisov:
thanks!

PR #187, opened 2015-06-30
Update build.proj

scott-xu:
Fix https://github.com/castleproject/Core/issues/97


kzu:
Fixed already in https://github.com/Moq/moq4/commit/adf6b6ad4e2aeae011bbd733646f953ec119b848 as soon as I noticed :).

scott-xu:
Thanks! When will it be released to nuget?

kzu:
Pushed!

Issue #92, opened 2014-02-11
Can't setup protected methods with nullable parameters


RobSiklos:
It seems that it's not possible to set up a protected method if it has nullable parameters.

Example:

``` csharp
public abstract class MyService
{
    protected abstract void MyMethod(int? val);
}

class Program
{
    static void Main(string[] args)
    {
        var mock = new Mock<MyService>();

        mock.Protected().Setup("MyMethod", (int?)1);
    }
}
```

The output of running this program is an `ArgumentException` with the following message:

Expression of type 'System.Int32' cannot be used for parameter of type 'System.Nullable`1[System.Int32]' of method 'Void MyMethod(System.Nullable`1[System.Int32])'

RobSiklos:
@kzu Ok - I finally got around to doing this :) Pull request #199.

RobSiklos:
Pull request #200 (single commit).

RobSiklos:
@kzu Any reason why the pull request hasn't been merged yet?

stakx:
I guess we can close this issue, now that #200 has been merged.

PR #199, opened 2015-09-14
Issue 92: Can't setup protected methods with nullable parameters

RobSiklos:
Implemented solution as per https://github.com/Moq/moq4/issues/92#issuecomment-37041881


kzu:
Could you squash both commits into a single one so as not to pollute history with tabs related changes?

Thanks for implementing this, btw! Good catch :+1:

stakx:
I'm closing this PR since it is was obsoleted by #200, which has been merged.

PR #232, opened 2015-12-24
Added Beginner's Guide to Mocking using Mock to README.md

sethdorris:


stakx:
Hi @sethdorris, thanks for taking the time to put this together. I think this would be a great addition in the project Wiki. I think it would be good to keep the `README.md` fairly brief and only add a link to this tutorial in the Wiki (similar to how it's done for the quickstart guide).

Probably we can set up a new Wiki section, e.g. "User contributions" or "Tutorials". What is your opinion on that?

(Btw. there are still a few formatting issues with headings (you need an empty line before/after headings or they'll be rendered incorrectly.)

stakx:
I am going to close this for the time being. @sethdorris, don't take this to mean that your content isn't welcome. I still think it would be nice to have something like this in the Wiki (in a contrib section), but IMHO the README document isn't quite the right place for such extensive content.

PR #238, opened 2016-01-05
VerityMocks and VerifyAllMocks Static Helper Methods

RehanSaeed:
I tend to write a method that does multiple verifications a lot, it saves a few lines:

```
mock1.VerifyAll();
mock2.VerifyAll();
mock3.VerifyAll();

// Versus

Mock.VerifyAll(mock1, mock2, mock3);
```

### DIFF
--- a/Source/Mock.cs
+++ b/Source/Mock.cs
@@ -122,8 +122,8 @@ namespace Moq
 			throw new ArgumentException(Resources.ObjectInstanceNotMock, "mocked");
 		}
 		
-		/// <include file='Mock.xdoc' path='docs/doc[@for="Mock.VerifyMocks"]/*'/>
-		public static void VerifyMocks(params Mock[] mocks)
+		/// <include file='Mock.xdoc' path='docs/doc[@for="Mock.Verify"]/*'/>
+		public static void Verify(params Mock[] mocks)
         	{
             		foreach (var mock in mocks)
             		{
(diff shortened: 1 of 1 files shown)


Source/Mock.cs
```diff
@@ -121,6 +121,24 @@ protected Mock()
 
 			throw new ArgumentException(Resources.ObjectInstanceNotMock, "mocked");
 		}
+		
+		/// <include file='Mock.xdoc' path='docs/doc[@for="Mock.VerifyMocks"]/*'/>
+		public static void VerifyMocks(params Mock[] mocks)
```
kzu: I think I like the original proposal naming better: `Verify(mocks)`

Source/Mock.cs
```diff
@@ -121,6 +121,24 @@ protected Mock()
...
+        	{
+            		foreach (var mock in mocks)
+            		{
+                		mock.Verify();
+            		}
+        	}
+		
+		/// <include file='Mock.xdoc' path='docs/doc[@for="Mock.VerifyAllMocks"]/*'/>
+		public static void VerifyAllMocks(params Mock[] mocks)
```
kzu: Ditto here, I like `VerifyAll(mocks)` better

kzu:
If these two get renamed to Verify and VerifyAll, I'll merge it ;)

RehanSaeed:
@kzu Methods renamed.

Issue #163, opened 2015-03-20
Stack overflow when setting up mock with default value set to mock.


psmolinsky:
Following code used to work in 4.2.1409.1722, but does not in 4.2.1502.0911. It throw a `StackOverflowException`. Please fix it. This looks similar to https://github.com/Moq/moq4/issues/161

```
var httpContext = new Mock<HttpContextBase> { DefaultValue = DefaultValue.Mock };
httpContext.SetupGet(c => c.Request.Cookies).Returns(new HttpCookieCollection());
```

Issue #161, opened 2015-03-13
Stack overflow when setting up a property on a mock that references another instance of the same type


ArildF:
The following code gives a StackOverflowException in Moq 4.2.1502.911, but not in 4.1.1309.1617:

``` C#
public interface IFoo
{
    IBar Bar { get; }
}

public interface IBar
{
    IBar SomeOtherBar { get; }
    string ATC { get; }
}

[TestMethod]
public void SomeTest()
{
    var factory = new MockRepository(MockBehavior.Strict){DefaultValue = DefaultValue.Mock};
    var foo = factory.Create<IFoo>();
    foo.SetupGet(r => r.Bar.ATC).Returns("asjkhsd"); // boom
}
```

This appears to be a similar problem to https://code.google.com/p/moq/issues/detail?id=332 , although in this case the result is a stack overflow and not an infinite loop as it appears in that issue.

Using DefaultValue.Empty avoids the problem; however, this may not be a practical solution in our case. The value of MockBehavior does not matter.

stakx:
This appears to have been fixed. The error can no longer be reproduced in Moq 4.7.25.

PR #245, opened 2016-01-29
Fixing issues #161 and #163

vladonemo:
This contains fix for the two issues causing StackOverflowException when using DefaultValue.Mock.
The idea of the fix is the following:
* SetupAllProperties avoids loops in the graph of the properties. For example, if a mocked type contains a property of the same type, then this property is mocked with default value (null for reference types) instead. In fact, this was already addressed, but only in the direct child level.
* The test from issue #163 was failing even after the fix due to the fact, that some of the .NET framework types do not implement ISerializable properly (e.g. System.Reflection.Module, System.Reflection.Assembly). So in practice, you are not able to mock IFoo with DefaultValue.Mock if the IFoo contains a property of type Assembly (or Module, or even Type) anywhere in the property graph, because Castle throws ArgumentException. To fix this one I introduced IsSerialiableMockable extension method and use it in the SetupAllProperties (via SerializableTypesValueProvider decorator)


kzu:
Could you fix the conflicts by rebasing on master so I can take another look at merging it?

Thanks!

vladonemo:
should be ok now, thanks.

kzu:
Hi @vladonemo sorry I broke again the merge-ability :(

I do see a large amount of changes that are just whitespace changes. Please make sure you have the EditorConfig VS extension and auto-format again from VS the files, all whitespace changes should be gone after that since the repo (master) now contains an .editorconfig file with the right settings.

Let me know if that works, thanks!

vladonemo:
I have rebased, but now the AppVeyor build fails. I can't build it locally either. Missing Castle in the references. Any idea why?

kzu:
Castle reference should be fixed now, but another rebase is needed now :(

vladonemo:
done. I had to fix the build error ince the XUnit was updated to v2. And removed one dummy file which I checked in by coincidence.

kzu:
The whitespaces changes are still there :(.

Please make sure you have the EditorConfig VS extension and auto-format again from VS the files, all whitespace changes should be gone after that since the repo (master) now contains an .editorconfig file with the right settings.

vladonemo:
oh, I'm sorry. I used different PC. I applied the EditorConfig rules now to all 6 files that I've touched. I hope it's fine now. I also have Resharper, so not sure whether it fights with its formatting rules ...

kzu:
Looks good. I'm sorry you have to use R# ;)

Thanks!

PR #251, opened 2016-03-09
Add an helper class to simplify parameter capture

ocoanet:
It is frequently required to capture a mock parameter to perform complex assertions. Parameter capture is [a regular request from new Moq users](https://www.google.com/#q=moq+capture+argument).

It can be done with setups and callbacks, but the syntax can be tricky, especially for methods with many parameters.

This helper class simplifies parameter capture:
```cs
// Arrange
var capture = new MockCapture<string>();
mock.Setup(x => x.DoSomething(capture.CaptureAny()));

// Act
// ...

// Assert
Assert.Equal("World!", capture.Last);
```

### DIFF
--- a/Source/Capture.cs
+++ b/Source/Capture.cs
@@ -1,0 +1,118 @@
+//Copyright (c) 2007. Clarius Consulting, Manas Technology Solutions, InSTEDD
+//http://code.google.com/p/moq/
+//All rights reserved.
+
+//Redistribution and use in source and binary forms, 
+//with or without modification, are permitted provided 
+//that the following conditions are met:
+
+//    * Redistributions of source code must retain the 
+//    above copyright notice, this list of conditions and 
+//    the following disclaimer.
+
+//    * Redistributions in binary form must reproduce 
+//    the above copyright notice, this list of conditions 
+//    and the following disclaimer in the documentation 
+//    and/or other materials provided with the distribution.
+
+//    * Neither the name of Clarius Consulting, Manas Technology Solutions or InSTEDD nor the 
+//    names of its contributors may be used to endorse 
+//    or promote products derived from this software 
+//    without specific prior written permission.
+
+//THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+//CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+//INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+//MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+//DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR 
+//CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+//SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+//BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+//SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+//INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+//WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+//NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+//OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+//SUCH DAMAGE.
+
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace Moq
+{
+    /// <summary>
+    /// Allows to create parameter captures in setup expressions.
+    /// </summary>
+    public static class Capture
+    {
+        /// <summary>
+        /// Creates a parameter capture that will store values in a collection.
+        /// </summary>
+        /// <typeparam name="T">The captured object type</typeparam>
+        /// <param name="collection">The collection that will store captured parameter values</param>
+        /// <example>
+        /// Arrange code:
+        /// <code>
+        /// var parameters = new List{string}();
+        /// mock.Setup(x => x.DoSomething(Capture.In(parameters)));
+        /// </code>
+        /// Assert code:
+        /// <code>
+        /// Assert.Equal("Hello!", parameters.Single());
+        /// </code>
+        /// </example>
+        public static T In<T>(ICollection<T> collection)
+        {
+            var match = new CaptureMatch<T>(collection.Add);
+            return With(match);
+        }
+
+        /// <summary>
+        /// Creates a parameter capture that will store specific values in a collection.
+        /// </summary>
+        /// <typeparam name="T">The captured object type</typeparam>
+        /// <param name="collection">The collection that will store captured parameter values</param>
+        /// <param name="predicate">A predicate used to filter captured parameters</param>
+        /// <example>
+        /// Arrange code:
+        /// <code>
+        /// var parameters = new List{string}();
+        /// mock.Setup(x => x.DoSomething(Capture.In(parameters, p => p.StartsWith("W"))));
+        /// </code>
+        /// Assert code:
+        /// <code>
+        /// Assert.Equal("Hello!", parameters.Single());
+        /// </code>
+        /// </example>
+        public static T In<T>(IList<T> collection, Expression<Func<T, bool>> predicate)
+        {
+            var match = new CaptureMatch<T>(collection.Add, predicate);
+            return With(match);
+        }
+
+        /// <summary>
+        /// Creates a parameter capture using specified <see cref="CaptureMatch{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The captured object type</typeparam>
+        /// <example>
+        /// Arrange code:
+        /// <code>
+        /// var capturedValue = string.Empty;
+        /// var match = new CaptureMatch{string}(x => capturedValue = x);
+        /// mock.Setup(x => x.DoSomething(Capture.With(match)));
+        /// </code>
+        /// Assert code:
+        /// <code>
+        /// Assert.Equal("Hello!", capturedValue);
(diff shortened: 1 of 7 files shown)


ocoanet:
It was not easy to name the class members, for example the `AsParameter` method can be simply called `Capture`. The `RunOnCapture` and `CopyOnCapture` methods can also be renamed or changed to delegate properties.

I am open to any suggestion or comment.

kzu:
Need to think a bit more about this, but it looks interesting... Maybe others can chime in too for feedback?

hahn-kev:
This would be really cool, I would like to see matching support with this though, currently it just matches all, while that's the typical use case I've had cases for other captures. You might also want to get the argument for a specific call, so provide index access to the underlying list.

ocoanet:
@hahn-kev I updated the PR to include match support. The new syntax is : `mock.Setup(x => x.DoSomething(capture.CaptureAny()));` or `mock.Setup(x => x.DoSomething(capture.Capture(p => p.StartsWith("W"))));`.

Source/MockCapture.cs
```diff
@@ -0,0 +1,131 @@
...
+    public class MockCapture<T>
+    {
+        private readonly List<T> _capturedValues = new List<T>();
+        private Action<T> _callback;
+        private Func<T, T> _copier;
+
+        /// <summary>
+        /// Gets all the captured parameter values.
+        /// </summary>
+        public IEnumerable<T> All { get { return _capturedValues; } }
```
kzu: I'd expose IList<T> to make this more flexible, maybe wrapping a ReadOnlyList<T>?

ocoanet: Done

ocoanet:
I removed the underscores on the private field names to follow the naming rule of the project, sorry about that.

Source/MockCapture.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <summary>
+        /// Gets the single captured parameter value.
+        /// </summary>
+        public T Single { get { return _capturedValues.Single(); } }
+
+        /// <summary>
+        /// Gets the parameter to use in the setup expression.
+        /// </summary>
+        /// <returns>The setup expression parameter</returns>
+        public T CaptureAny()
```
kzu: I think `capture.Any()` reads better, and since there is already a Linq `Enumerable.Any()` and an overload `Enumerable.Any(func predicate)`, I'd propose to make this method and `Capture` both named `Any`

Source/MockCapture.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            return Match.Create(value => CaptureParameter(value), () => It.IsAny<T>());
+        }
+
+        /// <summary>
+        /// Gets the parameter to use in the setup expression.
+        /// </summary>
+        /// <returns>The setup expression parameter</returns>
+        public T Capture(Expression<Func<T, bool>> match)
```
kzu: Rename the parameter to `predicate` to match the enumerable `Any` overload

UnitTests/MockCaptureFixture.cs
```diff
@@ -0,0 +1,163 @@
...
+            mock.Object.DoSomething("World!");
+
+            Assert.Throws<InvalidOperationException>(() => capture.Single);
+        }
+
+        [Fact]
+        public void CanRunCallbackOnParameter()
+        {
+            var capture = new MockCapture<FooParam>();
+            capture.RunOnCapture(x => x.Value = 42);
```
kzu: This seems to be somewhat overlapping with a `Callback` before returns, which would also allow to mutate the value before returning it. 

The test even uses it only for that, never actually using the captured parameter. It's almost like a callback based on the parameter type, rather than a particular setup. Not sure about mixing that behavior with the rest.

Maybe make the MockCapture could have virtual `Any` implementations that could be further extended to add that if needed for your particular usage? I almost feel the same way about the copier stuff...

Source/MockCapture.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private bool CaptureParameter(T item)
+        {
+            var capturedValue = _copier != null ? _copier.Invoke(item) : item;
+            _capturedValues.Add(capturedValue);
+
+            if (_callback != null)
+                _callback.Invoke(item);
+
+            return true;
```
kzu: Since we always return true from this, maybe it should just be `void`?

Source/MockCapture.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <summary>
+        /// Provide an action to run on captured parameters.
+        /// </summary>
+        /// <paramref name="callback">an action to run on the captured parameters</paramref>
+        public void RunOnCapture(Action<T> callback)
+        {
+            _callback = callback;
+        }
+
+        private bool CaptureParameter(T item)
```
kzu: If this method were, say, `protected virtual OnCapture(T item)`, a derived class in your test project may provide additional behavior on the parameter as needed, including creating a copier and/or the run on capture stuff.

Source/MockCapture.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            var capturedValue = this.copier != null ? this.copier.Invoke(item) : item;
+            this.capturedValues.Add(capturedValue);
+
+            if (this.callback != null)
+                this.callback.Invoke(item);
+
+            return true;
```
kzu: Since we're always returning `true

kzu: Since we're always returning `true` from this method, maybe it should be `void`?

Source/MockCapture.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <summary>
+        /// Provide an action to run on captured parameters.
+        /// </summary>
+        /// <paramref name="callback">an action to run on the captured parameters</paramref>
+        public void RunOnCapture(Action<T> callback)
+        {
+            this.callback = callback;
+        }
+
+        private bool CaptureParameter(T item)
```
kzu: If this method was `protected virtual void OnCapture(T item)`, someone could implement the copier and execute on capture via a custom capture class...

Source/MockCapture.cs
ocoanet: I choosed the names `CaptureAny` and `Capture` to match `IsAny` and `Is` methods but I also like your idea to have only one method name with overloads. However, I think `Any` is confusing because the class also exposes captured values through methods like `All` or `Last`. Those methods could be renamed but it seems to me that naming the capture method `Capture` is more straightforward. What do you think ?

Source/MockCapture.cs
ocoanet: Ok, no problem. This is the name I wanted to use in the first place but I changed it to `match` in order to use the name of the `It.Is` method parameter.

UnitTests/MockCaptureFixture.cs
ocoanet: The test only uses the capture for the callback because it is a unit test. This feature is supposed to be used when the developer needs to capture values **and** to run a callback. In this case the `RunOnCapture` is quite usefull and more simple to use for methods with many parameters.

Source/MockCapture.cs
ocoanet: Although`CopyOnCapture` is not supposed to be used very often, it is quite convenient in some situations:
- When the same object is modified and used multiple times as an argument.
- When the object is reset after the method call, for example when the object is returned to an object pool.
I think that the copy use case is standard enough to be included in the default implementation of `MockCapture`. The method name is far from perfect but it seems quite explicit to me.

Of course, the capture class is usefull without this feature, so if you do not like it, it can be removed.

kzu: I just realized something: we're mixing multiple responsibilities in this class that are unrelated and therefore make it harder to understand how to use properly.

For example, the Single, Last, All, essentially are exposing list/collection-like behavior, whereas the Capture() method is the one that provides the one true feature of this class. Then, the run and copy on capture are essentially exposing hooks into that one true feature of this class.

Sooo, how about making the constructor of the class receive a function that could do both the copying and/or running a callback instead?:

```
public MockCapture(Func<T, T> onCapture)
```

And moreover, remove the list-keeping altogether too, so that it's up to the consumer to store them in a List<T>, a HashSet or whatever, also simplifying how you consume the captured values?

```
var captures = new List<Foo>();
var capture = new MockCapture<Foo>(x => 
{
  captures.Add(x);
  // Could copy, run callback, mutate the x, whatever
  return x;
});

mock.Setup(x => x.DoSomething(capture.Any());

Assert.Equal(1, captures.Count);
```

Thoughts?

kzu: Not sure if type inference would just work, but maybe the ctor could even be overloaded with an `Action<T>` if you don't care about mutating/copying the value? (Although it makes it slightly harder to understand why there are two and when I'd use one over the other, but I think it's  a small learning curve?)

kzu: Note also, that by moving out the list-keeping responsibility, `Any` now is very unambiguous on what it applies to, it's a single member on the class: `capture.Any(..)` :)

ocoanet: I like your idea to use a constructor parameter to provide the copy or capture callback.

However in my experience 99% of the time the developer will need to access the list of the captured objects or the last captured object. And because this is the main use case, it should be very straightforward. If the captured object list is removed from the MockCapture class, we must provide an easy way to create a capture from a list.

What I can think up at the moment:
- Add an extension method for `ICollection<T>`
```cs
mock.Setup(x => x.DoSomething(items.AsCapture());
```
- Add an other class, for example `MockCaptureList<T>` that exposes the captured items

But in my opinion it is simpler to expose the captured items on the `MockCapture` class.

ocoanet:
Ok, I decided to refactor everything to take your remarks into account.

I removed `MockCapture`, there are now two new classes:
- `Capture`, a static class that can be used to create simple captures.
- `CaptureMatch<T>`, a match class that can run a callback on captured values.

The main use case is now straightforward:
```cs
// Arrange
var mock = new Mock<IFoo>();
var items = new List<string>();
mock.Setup(x => x.DoSomething(Capture.In(items)));

// Act
// ...

// Assert
var expectedValues = new List<string> { "Hello!" };
Assert.Equal(expectedValues, items);
```

Advanced use cases are still supported using `CaptureMatch<T>`:
```cs
var mock = new Mock<IFoo>();
var captureMatch = new CaptureMatch<string>(s => CustomCode(s));
mock.Setup(x => x.DoSomething(Capture.With(captureMatch)));
```

kzu:
Beautiful! 

Thanks a lot!

kzu:
btw, are you https://twitter.com/ocoanet?

kzu:
Shipped! https://www.nuget.org/packages/Moq/4.5.10

PR #255, opened 2016-04-09
Fix the out parameter issue of delegate mock

urasandesu:
The out parameter of delegate mock was handled as same as ref parameter of method mock, because `ParameterInfo.Attributes` wasn't copied when creating delegate proxy.


kzu:
Awesome PR :). Thanks!

PR #261, opened 2016-05-17
Adding exension methods ReturnsAsync and ThrowsAsync for ISetupSequentialResult

abatishchev:


Source/SequenceExtensions.cs
```diff
@@ -25,5 +22,27 @@ public static class SequenceExtensions
 		{
 			return new SetupSequentialContext<TMock, TResult>(mock, expression);
 		}
+
+		/// <summary>
+		/// Return a sequence of tasks, once per call.
+		/// </summary>
+		public static ISetupSequentialResult<Task<TResult>> ReturnsAsync<TResult>(this ISetupSequentialResult<Task<TResult>> setup, TResult value)
+		{
+			var tcs = new TaskCompletionSource<TResult>();
```
kzu: Why not Task.FromResult directly?

abatishchev: Yes, I was wondering myself too.
Apparently it's available only in .NET 4.5 http://stackoverflow.com/a/21258011/41956
But all projects seems are targeting .NET 4.0 so I decided do not change much  including tests same PR.

UnitTests/SequenceExtensionsFixture.cs
```diff
@@ -21,6 +22,29 @@ public void PerformSequence()
...
+		public void PerformSequenceAsync()
+		{
+			var mock = new Mock<IFoo>();
+
+			mock.SetupSequence(x => x.DoAsync())
+				.ReturnsAsync(2)
+				.ReturnsAsync(3)
+				.ThrowsAsync(new InvalidOperationException());
+
+			Assert.Equal(2, mock.Object.DoAsync().Result);
```
abatishchev: Because of .NET 4.0 I can't properly await DoAsync() here and below.

kzu: Maybe it's time to drop the ball on .NET < 4.5? Since it isn't even a
aupported version anymore, is it?

abatishchev: I'm for it. The minimal officially supported  version now days is 4.5.2. So I'd jump to 4.6.1.

abatishchev:
Cool, thanks for merging!
Do you have any idea when to release the latest changes?

kzu:
Should be there already in the Moq.4.5.6-alpha

Thanks!

abatishchev:
How often do you release stable version usually? Please consider one.

kzu:
Usually after a couple PRs, but right now we're in the middle of adding
support for. Net core, so I'd rather wait a bit so it's a worthy 4.5 upgrade

abatishchev:
Does it mean wait for Core RTM/RTW? Would you consider the last release before Core?
Personally I'd have two: the last before .NET 4.5 and the last before Core.
As a compromising solution: upgrade to 4.5 now and then wait indefinitely till Core is released.

kzu:
Good point. Yes, upgrade to .NET 4.5 can ship now

PR #262, opened 2016-05-21
Fixed setup issues when specifying the same parameters in a different order (issue #252)

### LINKED ISSUE #252: Method setup broken when specifying the same parameters in a different order in setup
Code to reproduce:
```C#
using Moq;
using Xunit;

public class MoqTest
{
    [Fact]
    public void MoqWorks()
    {
        var a = new MyClass();
        var b = new MyClass();
        var mock = new Mock<IDoStuff<MyClass>>();
        mock.Setup(m => m.DoStuff(a, b)).Returns(1);
        mock.Setup(m => m.DoStuff(b, a)).Returns(1);
        Assert.Equal(1, mock.Object.DoStuff(a, b));
        Assert.Equal(1, mock.Object.DoStuff(b, a));
    }

    public interface IDoStuff<in T>
    {
        int DoStuff(T a, T b);
    }

    public class MyClass { }
}

```
Both asserts should pass but one fails as `DoStuff` returns the default value of 0.

This was tested with Moq version 4.2.1510.2205.

LeonidLevin:
This should fix #252.
The problem was in how hash code was calculated for `ExpressionKey` objects: expression keys with the same values but in a different order would have the same hash code, so that `calls.ContainsKey(key)' would think that the key already existed in the dictionary.

I modified the `GetHashCode()` function in the `ExpressionKey` class to fix this. And added a unit test in `IssueReportsFixture.cs`.

### DIFF
--- a/Source/Interceptor.cs
+++ b/Source/Interceptor.cs
@@ -175,21 +175,26 @@ namespace Moq
 				return eq;
 			}
 
-			public override int GetHashCode()
-			{
-				var hash = fixedString.GetHashCode();
+            public override int GetHashCode()
+            {
+                var hash = fixedString.GetHashCode();
 
-				foreach (var value in values)
-				{
-					if (value != null)
-					{
-						hash ^= value.GetHashCode();
-					}
-				}
+                var factor = 1;
+                foreach (var value in values)
+                {
+                    if (value != null)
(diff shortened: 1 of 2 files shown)


kzu:
Creative workaround ;)

Issue #220, opened 2015-12-02
Unexpected behaviour with MockExtensions.Reset


tonyhallett:
The only test for Reset shows that for a loose mock set up with a return value, after Reset has been called, the return value defaults.  Aside from the additional reset calls is this the only intended behaviour ? I ask as this is not what I expected from the documentation ( although I am not familiar with Testeroids if this is the sole purpose )

I expected that the mock would behave as if no setup had been done.  e.g a strict mock will not throw when invoking a method that has been setup but will when reset has subsequently been called.

cscorley:
I'm also experiencing this confusion with what this method is supposed to accomplish.  I think it does exactly what it says it does, but perhaps the documentation around it is a bit misleading.  For example, all of these tests fail on the `VerifyAll`:
```csharp
[TestClass]
public class MoqResetVerifyTests {
    [TestMethod]
    [TestCategory("Unit")]
    public void Loose()
    {
        var myMock = new Mock<IEnumerable<int>>(MockBehavior.Loose);
        myMock
            .Setup(a => a.ToString())
            .Returns("Hello");
        myMock.Reset();
        Assert.AreNotEqual("Hello", myMock.Object.ToString());
        myMock.VerifyAll();
    }

    [TestMethod]
    [TestCategory("Unit")]
    public void Strict()
    {
        var myMock = new Mock<IEnumerable<int>>(MockBehavior.Strict);
        myMock
            .Setup(a => a.ToString())
            .Returns("Hello");
        myMock.Reset();
        Assert.AreNotEqual("Hello", myMock.Object.ToString());
        myMock.VerifyAll();
    }

    [TestMethod]
    [TestCategory("Unit")]
    public void LooseNoCall()
    {
        var myMock = new Mock<IEnumerable<int>>(MockBehavior.Loose);
        myMock
            .Setup(a => a.ToString())
            .Returns("Hello");
        myMock.Reset();
        myMock.VerifyAll();
    }

    [TestMethod]
    [TestCategory("Unit")]
    public void StrictNoCall()
    {
        var myMock = new Mock<IEnumerable<int>>(MockBehavior.Strict);
        myMock
            .Setup(a => a.ToString())
            .Returns("Hello");
        myMock.Reset();
        myMock.VerifyAll();
    }
}
```

kzu:
Could you post the failures too, for the record?

cscorley:
[...]

PR #277, opened 2016-08-03
Moq Reset should clear all existing calls in the interceptor

anilkamath87:
This should fix https://github.com/Moq/moq4/issues/220

The issue was that the calls in the interceptor were not cleared when reset was called.


kzu:
Thanks @anilkamath87!

PR #282, opened 2016-08-20
Fix Castle.Core dependency version

jeremymeng:
The net45 project still depends on 3.3.3

### DIFF
--- a/Moq.nuspec
+++ b/Moq.nuspec
@@ -178,7 +178,7 @@ Version 1.0
 </releaseNotes>
     <dependencies>
       <group targetFramework=".NETFramework4.5">
-        <dependency id="Castle.Core" version="4.0.0-beta001" />
+        <dependency id="Castle.Core" version="3.3.3" />
       </group>
       <group targetFramework=".NETStandard1.3">
         <dependency id="Castle.Core" version="4.0.0-beta001" />


jeremymeng:
@nil4 This surprised me.  Depending on Castle.Core v3.3.3 for .NETFramework4.5 is the same as before I added the .NET Standard support.  I double-checked.  [The v4.5 project builds against Castle.Core v3.3.3](https://github.com/moq/moq4/blob/080c0fc231c8dcbe4e606d5fe10694b95124e876/Source/project.json) and the .nuspec has [the right binary](https://github.com/moq/moq4/blob/080c0fc231c8dcbe4e606d5fe10694b95124e876/Moq.nuspec#L209).  So I wonder where the problem comes from.

Do you happen to have a small project that demonstrates this issue?

nil4:
Afraid I don't have a minimal repro that I can share. I plan to upgrade to .NET Core 1.1 preview1 later this week and will give it another go; if it still happens, I'll try to trim my project down.

nil4:
I upgraded to .NET Core 1.1 and the error is still present. Unfortunately my attempts to reduce the code to a simple repro failed (it works fine when reduced). Since Castle.Core 4.0.0-beta002 became available in the mean time, I added it to my test project.json, thus forcing Moq to use this version. This fixes the issue and the tests run as expected.

PR #288, opened 2016-09-20
Fix for issue #228: Events not raised on mocked type with multiple interfaces

### LINKED ISSUE #228: Events not working with CallBase = true depending on order interfaces are inherited.  C : IA, IB behaves different than C : IB, IA
public interface IFoo
{
        event EventHandler Foo;
}

public interface IBar
{
        event EventHandler Bar;
}

// reverse the order of IFoo and IBar will change the results of the test
public class FooBar : IFoo, IBar
{
        public event EventHandler Foo;
        public event EventHandler Bar;
        public void RaiseEvents()
        {
               if (Foo != null)
                       Foo(this, EventArgs.Empty);
               if (Bar != null)
                       Bar(this, EventArgs.Empty);
        }
}

[TestClass]
public class MoqTest
{
        [TestMethod]
        public void MultipleInterfacesWithEvents()
        {
               var mock = new Mock&lt;FooBar&gt;() { CallBase = true };
               IFoo foo = mock.Object;
               IBar bar = mock.Object;

               bool fooCalled = false;
               bool barCalled = false;
               foo.Foo += (s, e) => fooCalled = true;
               bar.Bar += (s, e) => barCalled = true;

               mock.Object.RaiseEvents();

               Assert.IsTrue(fooCalled);
               Assert.IsTrue(barCalled);
        }
}

bewaretherobots:
The ImplementedInterfaces list on Mock contains the public interfaces of our mocked type, the internally defined IMocked<T> type, and anything else added to the mock via calls to As<T>.  When looking for events, only the last of these (and the concrete mocked type) should be looked at.

Previously the assumption was that ImplementedInterfaces' first entry was the IMocked<T> type, and that the rest were always needed.  This meant the first entry in that list could be skipped, and the rest were always taken.  After 162a54379026a42bbc746335d63316819d078b56 added more interfaces to the list, however, this logic needs to change to count all internally implemented entries in the list and skip them all.

Fixing this will fix #228 which was caused when the EventInfo from an interface was intercepted, but the EventInfo from the underlying concrete type was actually expected.

### DIFF
--- a/Source/InterceptorStrategies.cs
+++ b/Source/InterceptorStrategies.cs
@@ -208,38 +208,25 @@ namespace Moq
 		/// <param name="eventName">Name of the event, with the set_ or get_ prefix already removed</param>
 		private EventInfo GetEventFromName(string eventName)
 		{
-			var depthFirstProgress = new Queue<Type>(ctx.Mock.ImplementedInterfaces.Skip(1));
-			depthFirstProgress.Enqueue(ctx.TargetType);
-			while (depthFirstProgress.Count > 0)
-			{
-				var currentType = depthFirstProgress.Dequeue();
-				var eventInfo = currentType.GetEvent(eventName);
-				if (eventInfo != null)
-				{
-					return eventInfo;
-				}
-
-				foreach (var implementedType in GetAncestorTypes(currentType))
-				{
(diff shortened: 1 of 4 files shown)


kzu:
Awesome instantly merge-able PR. Thanks!!!

PR #289, opened 2016-09-26
Added option to delay async returns and async throws

jochenz:
This helps when unit testing code which runs multiple async tasks in parallel. Simulating delays (especially many different random delays) allows for much better testing of multi-threaded code.


jdnichollsc:
+1

Issue #84, opened 2014-01-16
Provide better error reporting in Verify()


ashmind:
Example of reporting I get in 4.1.1311.615:

```
Moq.MockException : 
Expected invocation on the mock at least once, but was never performed:
r => r.Save(It.Is<Entity>(e => e.Description == .expectedDescription))
No setups configured.

Performed invocations:
IRepository`2.Save(My.Entity)
```

Reporting that would be more useful:

```
Moq.MockException : 
Expected invocation on the mock at least once, but was never performed:
r => r.Save(It.Is<Entity>(e => e.Description == "ABC"))
No setups configured.

Performed invocations:
IRepository`2.Save(My.Entity { Description = "XYZ" })
```

As for the second point, I have done similar thing once -- can be done by finding all distinct property/method paths in expression tree passed to `It.Is` and requesting their values for actual argument.

icalvo:
I think that for objects you have a nice workaround most of the times, which is to override `ToString()`. But if you do not have control over the class, this would be nice to have. You have to take into account the possible length of the resulting string, and also circular references if you recursively print property values. Not trivial at all!

Also interesting for enhancing reporting is when you have array parameters, which is specially useful for `params` methods. Also in-memory collections such as `Dictionary`, `List` and `Collection`. I'd not recommend to implement it for `IEnumerable` or other interfaces because you could have a huge (or even infinite) sequence there. The same caveats about the length of the resulting string and circular references apply.

stakx:
I think this idea has a lot of potential. But I don't think it would be good to decide on one fixed way of reporting additional object data, because of the problems this could cause (see @icalvo's [post above](https://github.com/moq/moq4/issues/84#issuecomment-154120405)).

I would suggest that this new functionality be exposed in a slightly more general way. For example, a new public contract could be introduced, say:

```csharp
namespace Moq.Diagnostics
{
    public interface IObjectFormatter
    {
        string Format(object obj);
    }
}
```

Exactly one instance of this new interface type could then be registered on a `Mock` or `MockRepository`. This would then be used in messages to (you guessed it) create a string representing a specific object. Such a string could consist only of the type name, but it could also include the property names and values of an object, or the first *n* items of a sequence, etc.

By default, Moq would be using an implementation that produces the exact same output as it does now. The functionality suggested in this issue could be added as a separate `IObjectFormatter` in the new `Moq.Diagnostics` namespace.

stakx:
[...]

PR #304, opened 2016-11-29
Updated actual invocation formatting to show the values for reference…

apollo2030:
When using verify to check for a specific invocation,  the failure message was listing invocations of the method with the arguments using the objects ToString() method. The default implementation of the ToString() is to provide the full name which for more complex types is not very usefull - we a re more interested in the values of the properties of the complex type arguments.

### DIFF
--- a/Source/Extensions.cs
+++ b/Source/Extensions.cs
@@ -83,15 +83,15 @@ namespace Moq
 			if (invocation.Method.IsPropertySetter())
 			{
 				return invocation.Method.DeclaringType.Name + "." +
-				       invocation.Method.Name.Substring(4) + " = " + GetValue(invocation.Arguments.First());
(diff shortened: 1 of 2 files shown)


kzu:
What if you have provided ToString override just for this purpose?

apollo2030:
I would have to do that for all the types the complex type consist of. Doing it here will prevent adding code to the types just for testing.

stakx:
See also issue #84.

stakx:
@apollo2030, I am closing this PR. The idea behind your proposal is a good one, but as per the comments above, this solution is too specific; a more generic mechanism / extension point would be preferable.

Thank you nevertheless for taking the time to prepare this! Perhaps we can base further work / discussion done on #84 on it so that your effort isn't completely lost!

PR #307, opened 2016-12-06
Update Times.cs

orenh1122:
a method to check if number of calls is a multiplication of a certain number

### DIFF
--- a/Source/Times.cs
+++ b/Source/Times.cs
@@ -121,6 +121,13 @@ namespace Moq
 
 			return new Times(c => c == callCount, callCount, callCount, Resources.NoMatchingCallsExactly);
 		}
+		
+		public static Times MultiplyOf(int num)
+		{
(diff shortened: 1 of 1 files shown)


kzu:
I'd rather have a `Times.Matching(Func<int, bool> evaluator, string failMessage)` instead of adding more very specific calculations.

stakx:
Unfortunately, an even bigger problem with this PR (apart from it being too specific) is that it breaks equality comparison. For example, the following assertion will fail:

```csharp
Assert.True(Times.Exactly(4) != Times.MultiplyOf(4));
```

`Times`'s equality comparison currently relies on the assumption that any `Times` instance can be logically mapped to a bounded, continuous number interval `[minCallCount..maxCallCount]`, thus to check two instances for equality, it would check their bounds for equality. Until now, this assumption held true for all available variants of `Times`.

`Times.MultiplyOf`, however, introduces for the first time a variant of `Times` that *cannot* be mapped to a *continuous* number interval, which is why it breaks equality comparison.

PR #309, opened 2016-12-10
ReturnsAsync() lazy evaluation issue fix #303

### LINKED ISSUE #303: ReturnsAsync() with factory method does not behave as expected
There is a bug in `ReturnsAsync<TMock, TResult>(this IReturns<TMock, Task<TResult>> mock, Func<TResult> value)`:

`Returns(() => f())` behaves as one would expect; i.e. call `f()` each time the method that was setup is called, to determine the value.

`ReturnsAsync(() => f())`, however, returns the same value every time the setup method is called.  Looking at the code, the reason why is clear: `ReturnsAsync()` evaluates `f()` at _`Setup()`_ time, rather than when the setup method is called, as one would expect:

		public static IReturnsResult<TMock> ReturnsAsync<TMock, TResult>(this IReturns<TMock, Task<TResult>> mock, Func<TResult> value) where TMock : class
		{
			return mock.ReturnsAsync(value());
		}

Workaround: drop back to `.Returns(() => Task.FromResult(f()))`

(That's the code fix as well, BTW).

SaroTasciyan:
###### Changes:
- `ReturnsAsync()` overload implementation
- Summary standardization for all `ReturnsAsync()` and `ThrowsAsync()` overloads
- Unit tests for lazy evaluation behavior
- New unit test fixture for _GeneratedReturnsExtensions_

### DIFF
--- a/Source/ReturnsExtensions.Generated.cs
+++ b/Source/ReturnsExtensions.Generated.cs
@@ -11,138 +11,157 @@ namespace Moq
     /// </summary>
 	public static class GeneratedReturnsExtensions
 	{
-        /// <summary>
-		/// Allows to specify the return value of an asynchronous method.
-		/// </summary>
-		public static IReturnsResult<TMock> ReturnsAsync<T, TMock, TResult>(this IReturns<TMock, Task<TResult>> mock, Func<T, TResult> value) where TMock : class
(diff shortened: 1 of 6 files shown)


kzu:
Thanks!

PR #310, opened 2016-12-14
Added the ability to specify the default value provider.

jwChung:
I have noticed the feedback - [Provide a callback before returning default mocks as values](https://moq.uservoice.com/forums/11304-general/suggestions/527991-provide-a-callback-before-returning-default-mocks) - which was already planned. I also hope that this feature will be added.

There is the PR #109 addresses this issue, but I think the PR causes breaking-change as the DefaultValue enum is changed to static class. This is slightly different from that to avoid the breaking-change.


Source/Mock.cs
```diff
@@ -148,18 +148,56 @@ public static void VerifyAll(params Mock[] mocks)
...
+			}
+			else if (value.GetType() == typeof(MockDefaultValueProvider))
+			{
+				this.defaultValue = DefaultValue.Mock;
+			}
+			else
+			{
+				this.defaultValue = DefaultValue.Custom;
+			}
+
```
stakx: Instead of faking polymorphism via a bunch of `if`-based type checks, perhaps it would be better to have a `DefaultValueKind` property on `IDefaultValueProvider` and then simply copy that property's value into `this.defaultValue[Kind]`.

Source/MockDefaultValueProvider.cs
```diff
@@ -48,34 +48,63 @@ namespace Moq
...
+				throw new ArgumentNullException(nameof(owner));
+			}
+
+			this.Owner = owner;
+		}
+
+		/// <summary>
+		/// Gets the owner of this provider passed through the constructor.
+		/// </summary>
+		public Mock Owner { get; }
```
stakx: Does this need to be made `public`?

Source/Mock.cs
```diff
@@ -148,18 +148,56 @@ public static void VerifyAll(params Mock[] mocks)
...
+				this.defaultValue = DefaultValue.Empty;
+			}
+			else if (value.GetType() == typeof(MockDefaultValueProvider))
+			{
+				this.defaultValue = DefaultValue.Mock;
+			}
+			else
+			{
+				this.defaultValue = DefaultValue.Custom;
+			}
```
stakx: Instead of faking polymorphism via a chain of `if`-based type checks, why not let a `IDefaultValueProvider` declare its kind via a `Kind` property (which is of type `DefaultValue`) so we can simply copy it over here?

Source/Mock.cs
```diff
@@ -148,18 +148,56 @@ public static void VerifyAll(params Mock[] mocks)
...
+			{
+				throw new ArgumentException("It's impossible that the Custom value is set directly. If you want to use custom default value provider, set the DefaultValueProvider property instead.");
+			}
+
 			this.defaultValue = value;
 			this.defaultValueProvider = defaultValue == DefaultValue.Mock
-				                            ? new MockDefaultValueProvider(this)
-				                            : new EmptyDefaultValueProvider();
+				? (IDefaultValueProvider)new MockDefaultValueProvider(this)
+				: new EmptyDefaultValueProvider();
```
stakx: Same here. Having both `DefaultValue` and `IDefaultValueProvider` is obviously somewhat redundant (and it's good that the former has been marked as `[Obsolete]` and would probably be phased out), but I wonder if there isn't a more robust approach for "converting" between the two types as long as we have them?

stakx:
Superseded by #366. Sorry for closing this, I really did give it a good long try (see rebased version at [`stakx/moq4:jwchung-defaultvalueprovider`](https://github.com/stakx/moq4/tree/c8ea891f1052e649767d9ad2d5c51f8d060b0827)), but the problem here is that it doesn't implement something like a `DefaultValueProvider` property on `MockFactory`. I tried implementing this on top of your commits but it turned out to be not easy at all. Understanding the changes was really difficult since your commits were based on a much older version of Moq. Thanks for the great effort, however!

stakx:
@jwChung: It's been a while. Just in case you're still interested, I just merged #533, which exposes a new abstract `DefaultValueProvider` base class and a `Mock[Repository].DefaultValueProvider` property.

This will be in the next release of Moq (4.8).

stakx:
@jwChung - A pre-release version of Moq (`4.8.0-rc1`) is now available on NuGet. It includes support for custom `DefaultValueProvider`s. Feel free to test away.

jwChung:
@stakx Thank you for your notifications and great work. After being interested in functional language, I rarely use [test double](http://xunitpatterns.com/Test%20Double.html). I hope I can try it later.

Issue #295, opened 2016-10-19
Identical setups are now not considered the same (regression)


cyanite:
The (partial) fix for #278 causes a regression in how identical setups are handled. Internally, a dictionary is used to prevent duplicates (and to let identical setups override eachother), but with the new equality code, this will almost never work. The reason is that each argument is compared by reference equality, from a `List<object>`. Value types are boxed in order to be in this list, and will thus never be equal (consider that `(object)1 != (object)1`).

The equality _and_ hash code should be reworked. Ideally only the string in the key should be used, and #278 should be solved by adding the appropriate arguments to the key string.

We're forced to downgrade Moq due to this bug, and due to a bug in Castle.Core 3.3.3 (used by this Moq version) involving default parameters.

MatKubicki:
I finally got back to taking a look at this.

The issue with comparing arguments by reference rather than by value has existed since long before my change, but my fix to make the comparison less buggy makes it more visible.

Since I haven't got the time to come up with a solution to the GetHashCode and Equals implmentations, i've fixed the problem I was trying to fix in a different way (by making the string key include the arguments for properties).  I've also added a comment to the problematic code in the Interceptor covering both the issue I found and the issue @cyanite found in the hope that it will help the next person to look at this code.

To top it off I added a unit test for this issue.

kzu:
Please reopen if the mentioned commit/PR didn't fix the issue.

Issue #278, opened 2016-08-04
Setups for properties with multiple arguments override each other even with differing values


MatKubicki:
This seems to relate to #252.

Consider the following test (VB specific due to the ability to have named properties with arguments):

```
    <Fact()>
    Public Sub SetupsForPropertiesWithMultipleArgsDoNotOverwriteEachOther()
        Dim mock As New Mock(Of ISimpleInterface)()

        mock.Setup(Function(m) m.PropertyWithMultipleArgs(1, 1)).Returns(1)
        mock.Setup(Function(m) m.PropertyWithMultipleArgs(1, 2)).Returns(2)

        Assert.Equal(1, mock.Object.PropertyWithMultipleArgs(1, 1))
        Assert.Equal(2, mock.Object.PropertyWithMultipleArgs(1, 2))

    End Sub

   Public Interface ISimpleInterface

        ReadOnly Property PropertyWithMultipleArgs(setting As Integer, setting2 As Integer) As Integer

    End Interface
```

This should pass, but instead it fails.  The problem is that the second setup is overwriting the first, even though the arguments are different 1, 2 instead of 1, 1.

The reasons are many and wonderful it seems!
The crux is all in Interceptor.AddCall
-The keyText created for this setup does not include the arguments, this is because the code sees it as a property getter, rather than a property with arguments.
-This means that the KeyText for both setups is the same, we have to rely on the arguments being different
-The ExpressionKey GetHashCode function (updated in #252) returns the same hash for both 1, 1 and 1, 2.
-Then The ExpressionKey equals method as a bug where it comparesthe arguments, it combines all the checks with an OR rather than an AND.

As this can only be tested in VB i've created a VB test and added it to the solution.
I've quickly fixed the OR that should be an AND, this fixes the problem.
But the GetHashCode implmentation needs a rethink as well I would suggest.

I'll push my simple changes and new test ASAP.

MatKubicki:
Ok so i've added my commit, not making a PR yet as I'd like to see the GetHashCode fixed as well.  Any thoughts from anyone on how to do this?

kzu:
IIRC, it's doing a logical OR, not a boolean OR to build the hash, so that should be the correct way

MatKubicki:
It does indeed use a logical XOR during the building of the hash, but the change i've made is in the Equals implmentation, that should be using an AND but was using an OR.

There appears to be a problem with the GetHashCode implementation as well, but i've haven't had a good look at that.

[...]

PR #313, opened 2017-01-03
Fixing #295 and changing fix for #278

MatKubicki:
So as my commit mentions this is an alternative fix for 278 that doesn't cause #295.  I've also added a test to cover the problem in 295.


kzu:
Thanks!

PR #322, opened 2017-02-16
Fix typo in summary of Mock<T>.Setup docs

xavierdecoster:
Before: 
`Specifies a setup on the mocked type for a call to to a void method.`

After: 
`Specifies a setup on the mocked type for a call to a void method.`


kzu:
Heya @xavierdecoster , just realized that this PR should have gone to master, not dev :(

xavierdecoster:
Opened a new PR against `master`, feel free to revert in `dev` :) 
https://github.com/moq/moq4/pull/325

PR #323, opened 2017-02-16
Upgrade Castle.Core dependency to 4.0.0

jeremymeng:
@kzu 

Any plan to get a non-pre release out?

After VS 2017 RTM, we can try switching to the new .csproj for both .NET Full and .NET Core.

### DIFF
--- a/Moq.nuspec
+++ b/Moq.nuspec
@@ -178,10 +178,10 @@ Version 1.0
 </releaseNotes>
     <dependencies>
       <group targetFramework=".NETFramework4.5">
-        <dependency id="Castle.Core" version="3.3.3" />
+        <dependency id="Castle.Core" version="4.0.0" />
       </group>
       <group targetFramework=".NETStandard1.3">
-        <dependency id="Castle.Core" version="4.0.0-beta001" />
+        <dependency id="Castle.Core" version="4.0.0" />
         <dependency id="System.Collections" version="4.0.11" />
         <dependency id="System.Collections.Concurrent" version="4.0.12" />
         <dependency id="System.Diagnostics.Debug" version="4.0.11" />
--- a/Source.NetCore/project.json
+++ b/Source.NetCore/project.json
@@ -1,6 +1,6 @@
 ﻿{
   "dependencies": {
-    "Castle.Core": "4.0.0-beta001",
+    "Castle.Core": "4.0.0",
     "IFluentInterface": "2.0.0",
     "GitInfo": "1.1.14",
     "Microsoft.CSharp": "4.0.1",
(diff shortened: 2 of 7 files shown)


kzu:
Absolutely! Since Castle.Core is no longer -pre, we could go ahead and merge this to master and ship

jeremymeng:
Thanks!

JohanLarsson:
Out of curiosity: do you bump major version on Moq when updating major version for a dependency?

kzu:
Not really. I'd like to bump the major version when there are major new features in Moq itself, so it's more meaningful for users.

kzu:
heya @jeremymeng, I pushed a merge from master>netcore to your branch. I dealt with all conflicts I found so far with master, but I'm having a few remaining errors:

```
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(21,87): error CS1660: Cannot convert lambda expression to type 'int' because it is not a delegate type [C:\Code\Personal\moq
4\build.proj]
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(33,87): error CS1660: Cannot convert lambda expression to type 'int' because it is not a delegate type [C:\Code\Personal\moq
4\build.proj]
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(47,109): error CS1660: Cannot convert lambda expression to type 'string' because it is not a delegate type [C:\Code\Personal
\moq4\build.proj]
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(58,109): error CS1660: Cannot convert lambda expression to type 'string' because it is not a delegate type [C:\Code\Personal
\moq4\build.proj]
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(71,31): error CS1660: Cannot convert lambda expression to type 'DateTime' because it is not a delegate type [C:\Code\Persona
l\moq4\build.proj]
C:\Code\Personal\moq4\UnitTests\GeneratedReturnsExtensionsFixture.cs(84,31): error CS1660: Cannot convert lambda expression to type 'DateTime' because it is not a delegate type [C:\Code\Persona
l\moq4\build.proj]
```

Also, the xproj projects don't open in VS2017, which wants to make a one-way upgrade. If you feel that's fine, plz go ahead and accept the migration so we can have something people can open right after cloning in the latest & greatest VS :D. 

Finally, once these issues are resolved, we should be able to merge it to master and bump the version to make it stable. It should be just a matter of updating https://github.com/moq/moq4/blob/netcore/GitInfo.txt

jeremymeng:
@kzu thanks for merging master to netcore and resolving the conflicts!  I've fixed the build errors by including the the one new file added in master.

I will spend some time to migrate the .xproj's.  Just to be clear that after migration the projects will not work in VS 2015. I hope this it not a concern at all.

kzu:
Maybe we can do the migration in a separate branch?

Issue #343, opened 2017-05-13
Open Generic Method Callback


michal-ciechan:
I can't find any way to mock an open Generic Method without specifying to type beforehand.

I have a couple questions, that someone might be able to help with:

1. Is this already implemented but I'm not finding it?
2. Is this possible (is has someone tried this but come up against a blocker)
3. Is this something that would be merged?
4. Some pointers of what classes I should look init. (To help speed up understanding the code, but will be going through it tomorrow anyhow)

Thinking something like

```csharp
Mock.Setup(x => x.GenericMethod<It.AnyType>())
    .Returns(context => *return type here*)
```

Where `context` will contain:

* Method Generic Type Args
* Passed in Parameters

The reason I would like to do this, is so so I can register hooks for my library 'Moqqer'.

kzu:
> GenericMethod<It.AnyType>

If you find a way to make that compile, I'll merge that ;)

michal-ciechan:
@kzu 
[Pull Request - Issue #343 - Allow Mocking of Open Generic Method](https://github.com/moq/moq4/pull/346) ;-)

stakx:
Closing this dormant issue, but marking it as "unresolved" so it can be easily found again. Please see #642 for details. If you'd like to pick this up and work on it, please post here briefly and we'll see what we can do!

stakx:
Reopening this in the hope that we can include this feature in [version 4.11.0](https://github.com/moq/moq4/milestone/8).

oddbear:
Nice.
This should work in general, but I don't see that it can work in case of ['Constraints on type parameters'](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters).
I can think of a couple of workarounds, but those would be more like hacks.

Any ideas?

stakx:
@oddbear - I'm not sure I understand. Could you please elaborate a little? What about generic type parameter constraints? (Yes, an `It.IsAnyType` wouldn't take those into account. What is stopping us from something like `It.IsAnySubtypeOf<T, U, V, ...>`?)

oddbear:
It's kind of an edge case, but I thinking about something like mocking this method:
```csharp
void MockThisMethod<T, U>()
    where T : ITest<U> //This is one of the rules allowed in C#
{
    //
}

public interface ITest<U>
{
    //
}
```

Then IsAnySubtype must be implementing that interface, and since it is not allowed to do:
```csharp
[...]

PR #346, opened 2017-05-20
Issue #343 - Allow Mocking of Open Generic Method

michal-ciechan:
Creating initial pull request to kick off discussions.

As per issue implemented `It.AnyType` interface, which allows for mocking of OpenGeneric methods.

Have exposed `IPublicCallContext` (was thinking of renaming existing `ICallContext` to `IInternalCallContext`, and public facing as `ICallContext`

Can be used as follows:

```
public interface IHasOpenGeneric
{
	TReturnArg Get<TInputArg1, TArgInput2, TReturnArg>(TInputArg1 arg1);
}

mock.Setup(x => x.Get<It.AnyType, It.AnyType, It.AnyType>(It.IsAny<It.AnyType>()))
	.Returns(context =>
	{
		callContext = context;
		callCount++;
		return "Test";
	});
```

Have created 2 tests, that test OpenGeneric input arg, return arg, and unused arg, including `where` clause.


kzu:
Up-front, this looks *very* promising. /cc @stakx

stakx:
I started looking into this on today's commute, but haven't quite finished yet. Once I understand all the changes, l'll try to post a proper review. Hopefully in the next couple of days.

Source/Proxy/IPublicCallContext.cs
```diff
@@ -0,0 +1,21 @@
+using System.Reflection;
+
```
stakx: This new file is missing the copyright header comment. Please add.

Source/Proxy/IPublicCallContext.cs
```diff
@@ -0,0 +1,21 @@
...
+		/// Arguments passed into the method
+		/// </summary>
+		object[] Arguments { get; }
+
+		/// <summary>
+		/// Closed generic method currently being executed.
+		/// </summary>
+		MethodInfo Method { get; }
+	}
+}
```
stakx: I do not yet quite understand why you chose to split `IPublicCallContext` off `ICallContext` and make it public? Why is this required at all? As far as I can see, it's only used for the new overload of `.Returns()`. To me, this seems like an orthogonal (i.e. completely separate) feature. Could this new public contract possibly be moved off into a separate PR? Or will the new `It.AnyType` feature not work without this?

Additionally, if this has to remain public in order for `It.AnyType` to work, I think it *might* be better to move it into the main `Moq` namespace. Not entirely certain on that point, though.

Source/It.cs
```diff
@@ -141,5 +141,35 @@ public static string IsRegex(string regex, RegexOptions options)
...
+		/// <summary>
+		/// Allows for setting up open generic methods on Mock Fixtures
+		/// </summary>
+		public interface AnyType
+		{
+			/// <summary>
+			/// Will contain the specific object being compared inside an open generic method.
+			/// Used for checking inside It.Is&lt;It.AnyType&gt;(x => ((int)x.Object) == 2)
+			/// </summary>
+			object Object { get; set; }
```
stakx: Is this actually required? It appears you're only ever set this property, but never read it...?

Source/It.cs
```diff
@@ -141,5 +141,35 @@ public static string IsRegex(string regex, RegexOptions options)
...
+		}
+
+		internal class AnyTypeImplementation : AnyType
+		{
+			public object Object { get; set; }
+		}
```
stakx: If I understand it correctly (guessing from your unit tests), the distinction between `interface AnyType` and `class AnyTypeImplementation` is needed so that one is able to bound / constrain type parameters. That would be a great feature to include. However I am not certain whether it will be obvious to most people how to use this feature. Do you think there would be a "more obvious" way of doing this, perhaps something along the lines of `It.AnySubTypeOf<TBase>` alongside `It.AnyType`?

Source/Language/IReturns.cs
```diff
@@ -67,6 +68,26 @@ public partial interface IReturns<TMock, TResult> : IFluentInterface
...
+		/// <code>
+		/// mock.Setup(x => x.Execute&lt;It.IsAnyType>())
+		///     .Returns(context => returnValues[0]);
+		/// </code>
+		/// The lambda expression to retrieve the return value is lazy-executed, 
+		/// meaning that its value may change depending on the moment the method 
+		/// is executed and the value the <c>returnValues</c> array has at 
+		/// that moment.
+		/// </example>
+		IReturnsResult<TMock> Returns(Func<IPublicCallContext, object> valueFunction);
```
stakx: The `Func`'s return type should be strongly typed, not `object`. Otherwise, I can write this:

```csharp
public interface IFoo
{
    int GetInt<T>(T arg);
}

[Fact]
public void Test()
{
    var mock = new Mock<IFoo>();
    mock.Setup(_ => _.GetInt(It.IsAny<string>())).Returns((callContext) =>
    {
        return "output";  // type mismatch not caught at compile time!
    });
    var foo = mock.Object;
    Assert.IsType<int>(foo.GetInt("input"));
}
```

Contrast this with Moq's `README`, which lists the following as a feature:

> Strong-typed: no strings for expectations, no object-typed return values or constraints

Source/MethodCall.cs
```diff
@@ -85,7 +85,7 @@ internal class TypeEqualityComparer : IEqualityComparer<Type>
 	{
 		public bool Equals(Type x, Type y)
 		{
-			return y.IsAssignableFrom(x);
+			return y.IsAssignableFrom(x) || y.IsAnyType();
 		}
 
 		public int GetHashCode(Type obj)
```
stakx: Please be aware that the proposed change to `Equals` breaks `GetHashCode`: If `a.Equals(b)`, then `a.GetHashCode() == b.GetHashCode()` should also be true. Therefore, if every type is considered equal to `It.AnyType`, then it follows that every type should have the same hash code as `It.AnyType`. That is, `GetHashCode` should return a constant. Which renders it pretty much useless for hashed containers.

Fortunately, at the moment, this argument is rather academic. This implementation of `IEqualityComparer<Type>` is currently only used together with `Enumerable.SequenceEqual`, which doesn't use `GetHashCode` at all. In order to stop future contributions from using this equality comparer for scenarios in which it would not perform well, please change `GetHashCode` to the following:

```csharp
int IEqualityComparer<Type>.GetHashCode(Type obj) => throw new NotSupportedException();
```

... and add a brief comment why it throws.

Source/MethodCallReturn.cs
```diff
@@ -90,6 +91,12 @@ public IVerifies Raises(Action<TMock> eventExpression, params object[] args)
 			return this.RaisesImpl(eventExpression, args);
 		}
 
+		public IReturnsResult<TMock> Returns(Func<IPublicCallContext, object> valueFunction)
```
stakx: As mentioned in a different context, the `Func`'s return type should not be `object`.

Source/Proxy/IPublicCallContext.cs
michal-ciechan: Thanks for pointing out. Added it now

Source/It.cs
michal-ciechan: Have added 2 tests to demonstrate usage, I added it for completeness in case anyone wants to add a conditional filter to the setup an open generic argument.

```
[Fact]
SetupReturns_ConditionalAnyType_MetCondition_CallsMockedReturn()
[Fact]
SetupReturns_ConditionalAnyType_UnmetCondition_DoesNotCallMockedReturn()

mock.Setup(x => x.Get<It.AnyType, It.AnyType, It.AnyType>(
	// Conditional input argument similar to any other It.Is<Type>(x => x == y)
	It.Is<It.AnyType>(type => type.Object as string == "TestInput")) 
)
	.Returns(context =>
	{
		callCount++;
		return "TestOutput";
	});
```

Source/It.cs
michal-ciechan: Will definitely look into trying to get It.AnySubTypeOf<TBase> style condition working. Thanks for the tip 👍

Source/Proxy/IPublicCallContext.cs
michal-ciechan: The reason I split it out, is to keep the `IPublicCallContext` as minimal as I can, without exposing some of the internal helper properties/methods which seemed to be `Castle.DynamicProxy` wrappers such as the following:

```
// Obsolete as the retunrr of the Func<T> is used, inline with rest of Moq
object ReturnValue { get; set; } 

// Could be potentially exposed? But rest of Moq uses a `.CallBase()` method?
void InvokeBase(); 

// Helper method Used internally by Moq?
void SetArgumentValue(int index, object value); 
```

`IPublicCallContext` is exposed in order to be able to construct meaningfull `.Returns()` calls as without it, it would be impossible to know what parameters the method was called with, nor the generic type parameters for the specific call.

It could be moved into the main `Moq` namespace as I guess `Moq.Proxy` seems to be `Castle.DynamicProxy` related items?

Source/Language/IReturns.cs
michal-ciechan: I did contemplate about adding this when I initially created it. Unfortunately for Open Generic methods where the return type is open generic, will require a `Func` that returns `object` no matter what, as at setup time the return type is not known.

But I definitely agree, that potentially allowing the existing workflow to not catch type mismatches at compile time is definitely a no-no. I will take a look at trying to extract the open generic interface so that the `Func<object>` return type is only available when in "return type open-generic" mode.

Worst case scenario may need to introduce (which I was contemplating initially) one of the following:

```
moq.SetupGeneric(...) 
// Then can provide different branch of IReturns, which includes a Return<Func<...,object>>()
```

or

```
moq.Setup(...).ReturnsGeneric()
// Overload that can accept a Func<...,object>
```

But I will first take another look at trying to infer `Func<...,It.AnyType>`, at which point I might be able to provide suitable overloads without breaking existing functionality.

michal-ciechan:
> Also, could you please retarget this PR at the develop branch (this can be done via the Edit button at the top, to the right of the PR title) and also rebase onto the develop branch? (I've done this for testing purposes, you can grab the rebased commits off stakx/moq4:itisanytype if you want.) That would be awesome.

Hey saw your 2nd part of your comment after doing a number of commits already to this branch. I have rebased to develop branch, hopefully, should be ok?.

Thanks for your feedback by the way!

Will look into `Func<object>` + `It.AnySunTypeOf<TBase>` over the course of this week.

stakx:
Thanks for having followed up on some suggestions already. I'm finishing for today, but will take a look at your changes tomorrow. 👍

Source/Language/IReturns.cs
michal-ciechan: The following now does not compile, as a `.Setup()' no that is not 'It.AnyType return type in the expression' does not contain the 'Returns(Func<T,object>)' overload anymore:

`new Mock<IBar>().Setup(x => x.Foo).Returns(context => 2);`

I did have a question regarding the documentation for the following:

```
// TODO: Find out about this (.xdoc?)
/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.Setup{}"]/*'/>
public ISetupAnyType<T, It.AnyType> Setup(Expression<Func<T, It.AnyType>> expression)
{
	return Mock.Setup(this, expression, null);
}
```

Wasn't sure what to put as the comment/documentation, and not sure how the current documentation works, any pointers would be great!

Currently to setup a method with an Open Generic return type that has a where clause requires a direct case (so that the generic `.Setup<TResult>(Expression<Func<TMock,TResult>>)` overload does not kick in. See updated test:

```
// Have to cast to (Expression<Func<IHasOpenGeneric, It.AnyType>>)
mock.Setup((Expression<Func<IHasOpenGeneric, It.AnyType>>)(x => x.GetWhere<ItAnySomeBaseClass, ItAnySomeBaseClass, ItAnySomeBaseClass>(It.IsAny<ItAnySomeBaseClass>())))
	.Returns(context =>
	{
		callContext = context;
		callCount++;
		return new SomeImplementationClass();
	});
```

I will be taking a look at changing SubClassing via It.AnySubTypeOf<T>, see if I get anywhere with that (as part of your `Source/It.cs` comment)

Source/Proxy/IPublicCallContext.cs
stakx: * I understand now why you chose to make part of the original interface public. (Even though it won't allow you to reliably discover the generic parameter types: You cannot call `.GetType()` on an argument that is `null`!) So far so good. What I am not entirely sure about is how your change potentially affects the preexisting "separation of concerns" between `.Callback()` and `.Returns()`. Your new `.Returns()` overload can be a little bit of both. This makes me wonder whether the new `.Returns()` is more similar to, and therefore should be, a `.Callback()` instead. I really don't mind the new syntax at all, I just can't foresee what consequences this might have further down the road.

* Splitting the interface to keep internals hidden is fine with me. I also think that you've made a good choice for what should go into the public interface. As long as noone demands `InvokeBase` or `SetArgumentValue`, let's keep them internal. They can always be added later, but once they're part of the public API, it'll be be more difficult to take them away again.

* Yes, please move the public interface to the `Moq` namespace. `Moq.Proxy` appears to bundle a few internal things surrounding DynamicProxy, and the end user probably doesn't care about that.

* Regarding the interfaces' names, to me it seems somewhat counterintuitive to let the internal `ICallContext` inherit from the public `IPublicCallContext`. The base interface's name would appear to imply that `ICallContext` has to be public, too. I personally find it more logical to have a `public interface ICallContext` (equivalent to your `IPublicCallContext`) and to have an internal subtype, `internal interface IExtendedCallContext` or something (which is today's `ICallContext`, but hopefully you can find a better new name for it ;-).

Source/It.cs
stakx: I'm sorry but I still don't understand what this is for. Can you perhaps show a brief example how this would be useful in practice?

Source/Language/IReturns.cs
stakx: I believe that at some point in the past, a decision was made to keep code and XML documentation comments separate. The XML documentation comments were then put into the `.xdoc` files. However, this isn't currently followed consistently and needs to be cleaned up at some point, anyway, so I tink it's fine if you just add normal XML documentation comments directly about your new code elements. ;-)

stakx: If you can make sure that the `object`-returning `.Returns()` overload isn't generally available, but only in the new "open generic mode", that'd be fine with me!

stakx:
Hi @michal-ciechan, sorry it's taken me a little longer to reply. (See new comments above.) I have a busy weekend ahead of me, so I probably won't be able to respond until sometime next week.

Source/Proxy/IPublicCallContext.cs
michal-ciechan: 1. Regarding `null` arguments, yes, the user can take a look at the `MethodInfo Method { get; }` property, which will have all the generic arguments populated.

2. I agree, better to start off slim with the interface.

3.  Moved to `Moq`

4. Yes agree as per my initial PR comment, I prefer the public one to be `ICallContext` and the internal one something else, some candidates:
 - `IExtendedCallContext`
 - `IInternalCallContext`
 - `IProxyCallContext`

For now, have used `IExtendedCallContext` as per suggestion :-)

Source/It.cs
michal-ciechan: To be honest, everything that could be in the `It.Is<It.AnyType>` `Predicate`, could be done in the `Returns` method.

The only reason I added it was because it was in line with the other `It.Is<T>` predicates.

I guess it would only be useful if `Verify` for `It.AnyType` was implemented.

I have no issue dropping this overload (less things to maintain)

michal-ciechan:
@stakx thanks for your comments, I am also away starting tomorrow until Wednesday, so won't be able to look into `It.AnySubTypeOf<TBase>` this week/weekend, so will have to wait till the latter part of next week. Have a good weekend and again, thanks for the great feedback!

Source/Proxy/IPublicCallContext.cs
stakx: Regarding the first bullet point, what on Earth was I thinking? I stand corrected!

Source/It.cs
michal-ciechan: It is not possible to get `It.AnySubTypeOf<TBase>` working as a type parameter directly due to the following

- `It.AnySubTypeOf<SomeBaseClass>' cannot be used as type parameter 'TReturnArg' in the generic type or method '...some method...'. There is no implicit reference conversion from 'It.AnySubTypeOf<SomeBaseClass>' to 'SomeBaseClass'.

I am still getting this compiler error even if I provide an implicit converter:

```
public class AnySubTypeOf<T> : It.AnyType where T : class
{
	public static implicit operator T(AnySubTypeOf<T> t)
	{
		return null;
	}
}
```
- Also cannot derive from a type parameters e.g.:

```
public class AnySubTypeOf<T> : T
{
}
```

The best I can come up with is the following:

```
mock.SetupGeneric(x => x
		.GetWhere<SomeBaseClass, SomeBaseClass, SomeBaseClass>(It.IsAnySubTypeOf<SomeBaseClass>())
	)
    .Return(...)
```

Where the mock is mocking the following method:

```
TReturnArg GetWhere<TInputArg1, TArgInput2, TReturnArg>(TInputArg1 arg1)
	where TInputArg1 : SomeBaseClass
	where TArgInput2 : SomeBaseClass
	where TReturnArg : SomeBaseClass;
```

Would that be sufficient?

It means adding a new method(s) `SetupGeneric()` and then using `It.IsAnySubTypeOf<SomeBaseClass>` which has the above implicit conversion to subtype, in order to satisfy input args.

stakx:
@michal-ciechan - This PR got closed accidentally because I terminated the `develop` branch. (That is, I merged it back into `master` and deleted it. Maintaining a separate `develop` branch ended up being more trouble than it solved.)

This is not to say that your PR wasn't welcome (it was, and still is, welcome). Please feel free to reopen your PR against the `master` branch if you're still interested in getting this in.

Secondly, I apologize for letting you wait so long. I should've let you know much sooner why I stayed away from your PR all this time. Until very recently, I couldn't quite put the finger on it myself, but now I know: It was the fact that we were about to expose `ICallContext` in the public API, and I wasn't myself ready (familiar enough with Moq) to judge how to best do this. For one thing, I wasn't quite happy with letting the public see that name; it's not exactly obvious what a "call context" encompasses, and that it simply represents a single call. I've spent some time making terms and class names more consistent internally, `ICallContext` has turned into `Invocation`. We'd now be free to let it implement an interface `IInvocation` that exposes those parts of an `Invocation` that we want public consumers to see (`Method` and `Arguments`, both in a read-only form).

hahn-kev:
I'm still interested in this feature, do we know if it'll get merged in?

michal-ciechan:
It'll need some refactoring to use the new `Invocator` via an `IInterface`. I'm currently snowed under with other stuff so unlikely I'll get to it before summer.
________________________________

stakx:
@hahn-kev: We're currently blocking all further changes to Moq 4's API, while Moq 5 is being finalized. Because of this, it would probably make more sense to let Moq 5 lead the way from now on and have this new feature implemented there.

/cc @kzu

stakx:
A similar feature based on the ideas first posted here has found its way into Moq 4.13.0, so I am removing the _unresolved_ label.

Issue #341, opened 2017-05-05
Protected<T>.Setup fails on specific overloads


80er:
Hi,

was it by intention that the  commit c8c5db1 broke identification of some overloads? 

Up to that one it was possible to have overloads with a base and a derived class. Now the IsAssignableFrom leads to InvalidOperationException in SingleOrDefault as now more then one method is found. 

The previous implementation with Type.GetMethod also went  for strict parameter type matching.

stakx:
Closing this issue, since the PR has been merged. The additional overload is available starting with Moq version 4.7.25.

PR #347, opened 2017-05-22
Added overload to enforce old behavior of exact parameter matching

80er:


80er:
Hi,

this is adding an overload in Protected Mock to enable the old behavior of exact parameter type matching again (see also issue #341).

In the tests you can see the scenario in which the new behavior broke existing code. Hope that makes the problem more clear.

UnitTests/ProtectedMockFixture.cs
```diff
@@ -751,6 +771,16 @@ protected virtual string DoReturn(string a, string b)
 			protected virtual void SameFirstParameter(object a) { }
 
 			protected virtual void SameFirstParameter(object a, object b) { }
+
+			protected virtual FooBase OverloadWithDerived(MyBase myBase)
```
kzu: It's kinda confusing that the method is called `WithDerived` yet it receives a base type. How about naming it simply `Overloaded` or the like?

80er: Your totally right. Updated the method name accordingly. But a better name than "overloaded" also did not come into my mind. Finding good names is harder than coding ;)

PR #350, opened 2017-05-31
Add new method `Times.Matching` to `Times`

stakx:
This is a response to #307. A new method `Times.Matching(TimesEvaluator)` is added that represents an extension point for custom invocation call count checks. To make this possible, a new abstract base class `TimesEvaluator` is introduced for which `Times` becomes a thin wrapper. (This is necessary to preserve the publicly visible surface of `Times`.)

### Why a new type `TimesEvaluator`? Why not `Func<int, bool>`? ### 

Implementing a delegate-based `Times.Matching(Func<int, bool>)` would be difficult for at least two reasons:

1. The private fields `from` and `to` cannot be assigned any sensible values when all we have is a delegate. There is no guarantee that some arbitrary `Func<int, bool>` represents / works with a continuous interval of the form [`callCountFrom`..`callCountTo`], so these two fields don't even make sense in the general case of `Times.Matching`. They are however (currently) needed for equality comparison.

2. Speaking of which, how can equality comparison be done with only a delegate at hand? Object reference equality is too imprecise, and invoking the delegate for all possible `int` values and comparing return values is way too inefficient.

One possible solution for the second problem would be to use an `Expression<Func<int, bool>>`, but this introduces some complexity and overhead that probably isn't warranted; and equality comparisons like `Exactly(1) == Once()` are hard to do; and expression trees & expression tree compilation might not be available on all of Moq's target platforms.

Thus I chose to define a new abstract base class `TimesEvaluator` with the following responsibilities: 

* verifying a call count against some arbitrary criterion,
* providing an exception message when the criterion isn't met,
* comparing the evaluator with another one by value (where possible).

### Example for a custom evaluator: ###

What @orenh1122 originally asked for in #307 could be implemented as follows:

```csharp
class MultipleOfTimesEvaluator : TimesEvaluator
{
    private int n;

    public MultipleOfTimesEvaluator(int n)
    {
        this.n = n;
    }

    public override bool Verify(int callCount)
        => callCount % this.n == 0;

    public override string GetExceptionMessage(int callCount)
        => $"Expected invocation on the mock a multiple of {this.n} times, " + 
           $"but was {callCount} times";

    public override bool Equals(TimesEvaluator other)
        => other is MultipleOfTimesEvaluator mo && this.n == mo.n:
}

…
var multipleOf5 = new MultipleOfTimesEvaluator(5);
var times = Times.Matching(multipleOf5);
```

### DIFF
--- a/Source/Times.cs
+++ b/Source/Times.cs
@@ -47,17 +47,11 @@ namespace Moq
 	/// <include file='Times.xdoc' path='docs/doc[@for="Times"]/*'/>
 	public struct Times
 	{
-		private Func<int, bool> evaluator;
-		private string messageFormat;
-		private int from;
-		private int to;
+		private TimesEvaluator evaluator;
 
-		private Times(Func<int, bool> evaluator, int from, int to, string messageFormat)
+		private Times(TimesEvaluator evaluator)
 		{
 			this.evaluator = evaluator;
-			this.from = from;
-			this.to = to;
-			this.messageFormat = messageFormat;
 		}
 
 		/// <include file='Times.xdoc' path='docs/doc[@for="Times.AtLeast"]/*'/>
@@ -65,13 +59,15 @@ namespace Moq
 		{
 			Guard.NotOutOfRangeInclusive(() => callCount, callCount, 1, int.MaxValue);
 
-			return new Times(c => c >= callCount, callCount, int.MaxValue, Resources.NoMatchingCallsAtLeast);
+			var evaluator = new RangeInclusiveBasedTimesEvaluator(callCount, int.MaxValue, Resources.NoMatchingCallsAtLeast);
+			return new Times(evaluator);
 		}
 
 		/// <include file='Times.xdoc' path='docs/doc[@for="Times.AtLeastOnce"]/*'/>
 		public static Times AtLeastOnce()
 		{
-			return new Times(c => c >= 1, 1, int.MaxValue, Resources.NoMatchingCallsAtLeastOnce);
+			var evaluator = new RangeInclusiveBasedTimesEvaluator(1, int.MaxValue, Resources.NoMatchingCallsAtLeastOnce);
+			return new Times(evaluator);
 		}
 
 		/// <include file='Times.xdoc' path='docs/doc[@for="Times.AtMost"]/*'/>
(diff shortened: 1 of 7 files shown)


Source/Times.cs
```diff
@@ -168,17 +174,83 @@ internal string GetExceptionMessage(string failMessage, string expression, int c
...
+				expression);
 		}
 
 		internal bool Verify(int callCount)
 		{
-			return this.evaluator(callCount);
+			return this.evaluator.Verify(callCount);
+		}
+
+		private sealed class RangeExclusiveBasedTimesEvaluator : TimesEvaluator
```
stakx: Are you OK with nested private classes? Should this and the other one (starting at line 222 below) be placed in a separate code file?

Source/Times.cs
```diff
@@ -168,17 +174,83 @@ internal string GetExceptionMessage(string failMessage, string expression, int c
...
+				this.evaluator.GetExceptionMessage(callCount),
+				expression);
 		}
 
 		internal bool Verify(int callCount)
 		{
-			return this.evaluator(callCount);
+			return this.evaluator.Verify(callCount);
+		}
+
```
stakx: Are you OK with nested private classes in the same code file?

Source/TimesEvaluator.cs
```diff
@@ -0,0 +1,71 @@
...
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+using System;
+
+namespace Moq
+{
+	/// <include file='TimesEvaluator.xdoc' path='docs/doc[@for="TimesEvaluator"]/*'/>
+	public abstract class TimesEvaluator : IEquatable<TimesEvaluator>
+	{
```
stakx: This would be nicer as an interface (`ITimesEvaluator`), but defining it as an abstract base class saves subclassers most of the hassle of implementing the inherited interface `IEquatable<T>`.

Source/TimesEvaluator.cs
```diff
@@ -0,0 +1,71 @@
...
+
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+using System;
+
+namespace Moq
+{
+	/// <include file='TimesEvaluator.xdoc' path='docs/doc[@for="TimesEvaluator"]/*'/>
+	public abstract class TimesEvaluator : IEquatable<TimesEvaluator>
```
stakx: It would be nicer if this were an interface (`ITimesEvaluator`), but this way saves subclassers most of the hassle of fully implementing the `IEquatable<T>`.

Source/Times.cs
stakx: Are you OK with nested private classes placed in the same code file?

stakx:
I need to think about this some more before it has the chance to get merged. Opening up `Times` that way is possibly dangerous as long as `Times` is equatable.

PR #351, opened 2017-05-31
Simplify logic in `ExpressionMatcher.Matches`

### LINKED ISSUE #237: Compared with null after type conversion using 'as' keyword
[V3019](http://www.viva64.com/en/d/0388/) Possibly an incorrect variable is compared to null after type conversion using 'as' keyword. Check variables 'value', 'valueExpression'. ExpressionMatcher.cs 62

``` C#
public bool Matches(object value)
{
  var valueExpression = value as Expression;
  if (value == null) //<==
  {
    return false;
  }

  return ExpressionComparer.Default.Equals(this.expression, valueExpression);
}
```

It is most likely the object of the derived class that the programmer intended to check for null before using it. This is the fixed version of the code:

``` C#
public bool Matches(object value)
{
  var valueExpression = value as Expression;
  if (valueExpression == null)
  {
    return false;
  }

  return ExpressionComparer.Default.Equals(this.expression, valueExpression);
}
```

This place is found by a static analyzer [PVS-Studio](http://www.viva64.com/en/pvs-studio/).

stakx:
This fixes #237.

The method's previous semantics are preserved, but note that it could be simplified to just:

```csharp
return ExpressionComparer.Default.Equals(this.expression, value as Expression);
//                                                        ^^^^^^^^^^^^^^^^^^^
```

Nothing in `ExpressionMatcher` ensures that `this.expression` cannot be `null`. Therefore, the `null` special case could (and for correctness, probably should) be left entirely up to `ExpressionComparer.Equals`.

### DIFF
--- a/Source/Matchers/ExpressionMatcher.cs
+++ b/Source/Matchers/ExpressionMatcher.cs
@@ -58,13 +58,8 @@ namespace Moq.Matchers
 
 		public bool Matches(object value)
 		{
-			var valueExpression = value as Expression;
-			if (value == null)
-			{
-				return false;
-			}
-
-			return ExpressionComparer.Default.Equals(this.expression, valueExpression);
+			return value is Expression valueExpression
+				&& ExpressionComparer.Default.Equals(this.expression, valueExpression);
 		}
 	}
 }


stakx:
Closing this PR as I will re-submit its commit as part of a larger PR.

PR #353, opened 2017-05-31
Ensure default mock names are unique (fixes #348)

### LINKED ISSUE #348: new Mock<T> can generate not unique names
When submitting a pull request this test failed in the first try:

https://github.com/moq/moq4/blob/master/UnitTests/MockFixture.cs#L38

After checking the code two times the same name for a mock was generated but that should be unique. Problem is this line:

https://github.com/moq/moq4/blob/master/Source/Mock.Generic.cs#L126

Although it is (more or less) guaranteed that the whole guid is unique this does not hold true for the first 4 characters. This can lead to Heisenbugs while executing tests i think. However. Not sure how much functionality is depending on having unique names for Mocks currently.

stakx:
Default mock object names are built from the type name and 4 hex digits taken from a GUID. While whole GUIDs are unique in practice, the same cannot be said of only 4 of its digits.

Instead of using 4 hex digits from a GUID, create 8 unique but non-random hex digits in the simplest possible way: using a thread-safe ever-increasing counter.

This fixes #348.

### DIFF
--- a/Source/Mock.Generic.cs
+++ b/Source/Mock.Generic.cs
@@ -46,6 +46,7 @@ using Moq.Language.Flow;
 using Moq.Proxy;
 using Moq.Language;
 using System.Reflection;
+using System.Threading;
 
 #if !NETCORE
 using System.CodeDom;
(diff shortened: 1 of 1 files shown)


stakx:
Closing this PR as I will re-submit its commit as part of a larger PR.

PR #361, opened 2017-06-05
Make `It.IsAny`, `It.IsNotNull` work for COM types

### LINKED ISSUE #269: IsAssignableFrom breaks __ComObject support
trying to setup setter for a property of ISymbol type, which is COM interface:
ISymbol setSymbol = null;
elMock.SetupGet(e => e.Symbol).Returns(setSymbol);
elMock.SetupSet(e => e.Symbol = It.IsAny<ISymbol>()).Callback<ISymbol>(s => setSymbol = s);

set it during a test:
el.Symbol = (ISymbol)Com.CreateInstance<SimpleFillSymbolClass>();

this never works, setSymbol stays null, Symbol property also always returns null

stakx:
This fixes #269:

Testing a COM object's type using `IsAssignableFrom` can report false negatives because .NET might only see the generic RCW type `__ComObject` instead of the actual COM interface type. The `is` operator on the other hand appears to be performing `QueryInterface` for COM objects to get the right result.

This commit introduces an additional branch in `It.IsAny<>` that runs an `is`-based check for COM objects, and the as-is reflection-based check for everything else. The added logic takes effect only in builds targeting the full .NET Framework, since COM is not available on all .NET Core platforms.

I have successfully verified the new behaviour against a COM object, but am reluctant to add any permanent test code involving COM to Moq. That would tie the unit test project to a Windows platform-specific facility. (It's possible to work around this by running a test conditionally, depending on the value of `Environment.OSVersion.Platform`, but that seems a little hacky.)

### DIFF
--- a/Source/It.cs
+++ b/Source/It.cs
@@ -45,6 +45,9 @@ using System.Linq.Expressions;
 using System.Text.RegularExpressions;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+#if !NETCORE
+using static System.Runtime.InteropServices.Marshal;
+#endif
 
 namespace Moq
 {
@@ -55,7 +58,12 @@ namespace Moq
 		public static TValue IsAny<TValue>()
 		{
 			return Match<TValue>.Create(
+#if !NETCORE
+				value => value == null || (IsComObject(value) ? value is TValue
+				                                              : typeof(TValue).IsAssignableFrom(value.GetType())),
+#else
 				value => value == null || typeof(TValue).IsAssignableFrom(value.GetType()),
+#endif
 				() => It.IsAny<TValue>());
 		}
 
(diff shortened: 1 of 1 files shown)


Source/It.cs
```diff
@@ -55,15 +55,15 @@ public static class It
...
+				value => value == null || value is TValue,
 				() => It.IsAny<TValue>());
 		}
 
 		/// <include file='It.xdoc' path='docs/doc[@for="It.IsNotNull"]/*'/>
 		public static TValue IsNotNull<TValue>()
 		{
 			return Match<TValue>.Create(
-				value => value != null && typeof(TValue).IsAssignableFrom(value.GetType()),
+				value => value is TValue,
```
stakx: The null check can be incorporated into the type check, since `null is …` is always false.

stakx:
This probably needs some more work to make it safe (i.e. not break existing code).

PR #363, opened 2017-06-05
Fix equality check bug in ExpressionKey

### LINKED ISSUE #135: Wrong equality check in Interceptor.ExpressionKey
The equality check in Interceptor.ExpressionKey (line 171) is wrong. The check must be conjunction (&), but not disjunction(|), because then the values don't matter and only the fixedString is considered, but not the values. 
Also, the GetHashCode() must be calculated in a way the value position matters. Currently, it is val1.GetHashCode() ^ val2.GetHashCode()  ^ ... ^ valN.GetHashCode(), but it should be:
( ... ((val1.GetHashCode() \* <prime_num>) ^ val2.GetHashCode() \* <prime_num>)  ^ ... ^ valN.GetHashCode())

The fixes of these issues are in pull proposal - https://github.com/Moq/moq4/pull/134

So, invocations:
aMock.Setup(m => m.Foo(i1, i2));
aMock.Setup(m => m.Foo(i2, i1));

have the same expression key. As result the second overwrites the first and this is wrong.

Here is a test for it:

[Test]
public void MultipleSetupTest()
{
    Mock<IA> aMock = new Mock<IA>(MockBehavior.Strict);

```
Item i1 = new Item { Id = 1 };
Item i2 = new Item { Id = 2 };

aMock.Setup(m => m.Foo(i1, i2));
aMock.Setup(m => m.Foo(i2, i1));

aMock.Object.Foo(i1, i2);
aMock.Object.Foo(i2, i1);
```

}

public interface IA
{
    void Foo(Item x, Item y);
}

public class Item
{
    public int Id { get; set; }
}

The issue can be worked around in one of the following ways:
1) Override ToString() of "Item", so it returns value dependent on the class fields ("Id").
2) Change the setup be:
aMock.Setup(m => m.Foo(It.Is<Item>(x => x.Id == i1.Id), It.Is<Item>(x => x.Id == i2.Id)));
aMock.Setup(m => m.Foo(It.Is<Item>(x => x.Id == i2.Id), It.Is<Item>(x => x.Id == i1.Id)));
Here the idea is again that the value is int and its ToString() is overriden.

In both cases the value is written in the "fixedString" of ExpressionKey and that's why it works.

### LINKED ISSUE #328: Issue with setting up mock Objects, i.e. calling Mock<T>.Setup, when arguments have the same hashcodes
Hi,

I believe there is inconsistency in which Moq handles setting up matching call signatures.
Let consider following snippet of code:

```csharp
Argument arg1 = new Argument("aaa");
Argument arg2 = new Argument("bbb");

var mocked = new Mock<IToBeMocked>(MockBehavior.Strict);
mocked.Setup(x => x.BlaBla(arg1));
mocked.Setup(x => x.BlaBla(arg2));
// This invocation of Setup override's the first invocation

mocked.Object.BlaBla(arg1);
// This calls fails with statement: invocation failed with mock behavior Strict.
// All invocations on the mock must have a corresponding setup
```

Expected behavior: mocked.Object.BlaBla(arg1) call does not throw an exception.

My clue: call to `mocked.Setup(x => x.BlaBla(arg2))`

- is invoking: Interceptor.AddCall method
- which in turn invokes Interceptor.calls.ContainsKey(key)
- this invokes ExpressionKey.Equal and ExpressionKey.getHashCode
Then, ExpressionKey.getHashCode seems to be working fine, whether ExpressionKey.Equals seems to be extremely tolerant and generally always returns equality.

Finally, this introduce very awkward behavior where ability to match objects depends mainly at hashcodes and their potential collisions.

I would like to suggest fixing it, by changing ExpressionKey.Equal implementation to be as restrictive as gethashcode, i.e. for each different hashcode equals should return false. If you agree with such approach I could introduce such change by myself.

stakx:
This PR is a cleaned-up version of #134. It fixes the buggy `Equals` and `GetHashCode` implementations in `Interceptor.ExpressionKey`. 

This closes #135, and closes #328.

### DIFF
--- a/Source/Interceptor.cs
+++ b/Source/Interceptor.cs
@@ -173,7 +173,9 @@ namespace Moq
 				var index = 0;
 				while (eq && index < this.values.Count)
 				{
-					eq &= this.values[index] == key.values[index];
+					// using `object.Equals` instead of == ensures that we get the correct
+					// comparison result for boxed value types:
+					eq &= object.Equals(this.values[index], key.values[index]);
 					index++;
 				}
 
@@ -182,20 +184,17 @@ namespace Moq
 
 			public override int GetHashCode()
 			{
-				unchecked
+				var hash = fixedString.GetHashCode();
+
+				foreach (var value in values)
 				{
-					var hash = fixedString.GetHashCode();
-
-					foreach (var value in values)
+					if (value != null)
 					{
-						if (value != null)
-						{
-							hash = (hash * 397) ^ value.GetHashCode();
-						}
+						hash = unchecked((hash * 397) ^ value.GetHashCode());
 					}
-
-					return hash;
 				}
+
+				return hash;
 			}
 		}
 
--- a/Source/Interceptor.cs
+++ b/Source/Interceptor.cs
@@ -170,15 +170,10 @@ namespace Moq
 
 				var eq = key.fixedString == this.fixedString && key.values.Count == this.values.Count;
 
-				//the code below is broken as it uses an OR when checking the arguments, this means that if any pair of arguments match 
-				//the result is a match.
-				//This is only going to hit some edge cases as for the most part the fixed string above spots arguments correctly.
-				//Fixing this really needs a reworking of the GetHashCode, and also some sorting out of how to compare value types that have been
-				//boxed correctly (another problem this code has)
 				var index = 0;
 				while (eq && index < this.values.Count)
 				{
-					eq |= this.values[index] == key.values[index];
+					eq &= this.values[index] == key.values[index];
 					index++;
 				}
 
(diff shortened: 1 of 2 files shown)


Source/Interceptor.cs
```diff
@@ -170,41 +170,33 @@ public override bool Equals(object obj)
...
-        }
+			public override int GetHashCode()
+			{
+				var hash = fixedString.GetHashCode();
+
+				foreach (var value in values)
+				{
+					if (value != null)
+					{
+						hash = unchecked((hash * 397) ^ value.GetHashCode());
```
stakx: We don't care whether the hash value overflows or not, but we do not want overflows to trigger an exception. That is why `unchecked` is necessary here.

PR #366, opened 2017-06-06
Make default value providers part of Moq's public API

stakx:
This commit is a replacement for PRs #109 and #310. (See note further down.) It adds `IDefaultValueProvider` et al. to Moq's public API. It enables the usage pattern demonstrated below.

### Demo: ###

```csharp
// Register a default value provider with a Mock:
var mock = new Mock<IAnything>();
mock.DefaultValueProvider = new QuuxDefaultValueProvider();

// Or, register a default value provider factory with a MockRepository:
var mockRepository = new MockRepository();
mockRepository.DefaultValueProviderFactory = QuuxDefaultValueProviderFactory.Instance;
var mock = mockRepository.Create<IAnything>(…);

// This is how you implement a default value provider:
sealed class QuuxDefaultValueProvider : IDefaultValueProvider
{
   public IDefaultValueProviderFactory Factory => QuuxDefaultValueProviderFactory.Instance;
   public void DefineDefault<T>(T value) => …
   public object ProvideDefault(MethodInfo member) => …
}

// And this is how you implement a default value provider factory:
sealed class QuuxDefaultValueProviderFactory : IDefaultValueProviderFactory
{
   public static IDefaultValueProviderFactory Instance => new QuuxDefaultValueProviderFactory();
   public IDefaultValueProvider CreateProviderFor(Mock owner) => new QuuxDefaultValueProvider(…);
}
```

### Notes: ###

* Factories are introduced as an additional concept to allow parameterized provider instantiation. This is necessary for both `MockDefaultValueProvider` and the internal `SerializableTypesDefaultValueProvider`.

* Unlike the other two PRs, this one does not mark `DefaultValue` as `[Obsolete]`. After all, which of the following would you rather write as a Moq user:

    ```csharp
    // simple, and people are probably used to this:
    var mock = new Mock<IFoo> { DefaultValue = DefaultValue.Mock };`

    // not so simple, but what you'd have to do without the handy DefaultValue shortcut:
    var mock = new Mock<IFoo>();
    mock.DefaultValueProvider = new MockDefaultValueProvider(mock);
    ```

### Why not merge one of the previous pull requests #109 or #310? ###

There are currently two other PRs with the same goal, and they are great efforts. Unfortunately both of them suffer from small but important issues:

* #109 introduces a breaking API change by turning `DefaultValue` from an `enum` into a `public static` class. We really shouldn't do that without a previous warning / phase-out period. And it could be easily avoided.

* #310 on the other hand exposes `DefaultValueProvider` on `Mock`, but does not include similar functionality in `MockFactory` (aka `MockRepository`).

I spent a day trying to fix each PRs problems (see [`stakx/moq:jamesfoster-defaultvalueprovider`](https://github.com/stakx/moq4/tree/512d3ca9cdc24116e3452a7625dbb9894478a106) and [`stakx/moq:jwchung-defaultvalueprovider`](https://github.com/stakx/moq4/tree/c8ea891f1052e649767d9ad2d5c51f8d060b0827)), only to realise it would be easier to start anew. (This is partly due to changes in the repository since the PRs were created; both have to be carefully rebased, which I've done but still ended up with failing unit tests in both.)


stakx:
I'm closing this briefly. I suspect the API could be simplified a little more.

PR #368, opened 2017-06-07
Renormalize indentation whitespace to tabs

stakx:


JohanLarsson:
FWIW I prefer four spaces, renders nicer when reading code in the browser if nothing else.

stakx:
Yes, I personally prefer four spaces, too. However, in this commit, I tried not to let my personal preferences intrude: I decided to adhere to `.editorconfig` and to what the majority of Moq's lines of code already use.

PR #376, opened 2017-06-15
Make event accessor recognition logic more correct

### LINKED ISSUE #82: System.InvalidCastException when invoking method that starts with "add"
Hello, I faced with the problem when i used Moq.
I wrote the simple test

```
public interface ILogger
{
    void add_info(string info);
    void Add_info(string info);
}

[TestFixture]
public class Class1
{
    [Test]
    public void Test_add_info()
    {
        var mock = new Mock<ILogger>();
        mock.Setup(x => x.add_info(It.IsAny<string>()));
        mock.Setup(x => x.Add_info(It.IsAny<string>()));

        mock.Object.add_info("asd");
        //mock.Object.Add_info("asd");
    }
}
```

When i call

mock.Object.add_info("asd"); 

it throws 
System.InvalidCastException : Не удалось привести тип объекта "System.String" к типу "System.Delegate".
   в Moq.AddActualInvocation.HandleIntercept(ICallContext invocation, InterceptStrategyContext ctx)
   в Moq.Interceptor.Intercept(ICallContext invocation)
   в Castle.DynamicProxy.AbstractInvocation.Proceed()
   в Test.Class1.Test_add_info() в Class1.cs: line 22

Sorry, i have Russian VS, it says: "Can't cast System.String to System.Delegate". But, when i call 

mock.Object.Object.Add_info("asd"); 

string, it works.
I use VS 2012, .NET4.5. Nuget packages - NUnit, Moq(v.4.2.1312.1622). The Code was runned from Resharper test runner.

stakx:
Up until now, if Moq encounters a method named `add_X` or `remove_X` it will assume that it's found an specialname event accessor method without performing any further checks.

This commits adds some trivial type checks and null reference checks to improve that logic somewhat. I hesitate to make this too perfect since that just might break things for F# or COM (see commit 44070a9).

This partially fixes #82.

### DIFF
--- a/Source/MemberInfoExtensions.cs
+++ b/Source/MemberInfoExtensions.cs
@@ -61,12 +61,12 @@ namespace Moq
 		}
 
 
-		public static bool IsEventAttach(this MethodBase method)
+		public static bool LooksLikeEventAttach(this MethodBase method)
 		{
 			return method.Name.StartsWith("add_", StringComparison.Ordinal);
 		}
 
-		public static bool IsEventDetach(this MethodBase method)
+		public static bool LooksLikeEventDetach(this MethodBase method)
 		{
 			return method.Name.StartsWith("remove_", StringComparison.Ordinal);
 		}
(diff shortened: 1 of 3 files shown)


Source/MemberInfoExtensions.cs
```diff
@@ -61,12 +61,12 @@ public static bool IsDestructor(this MethodInfo method)
 		}
 
 
-		public static bool IsEventAttach(this MethodBase method)
+		public static bool LooksLikeEventAttach(this MethodBase method)
```
stakx: Let's be honest and call it what it *really* does.

PR #390, opened 2017-06-21
Mock.Of<T>(MockBehavior)

JohanLarsson:
Enable specifying mock behavior when using Mock.Of<T>(MockBehavior behavior)

### DIFF
--- a/Source/Linq/Mock.cs
+++ b/Source/Linq/Mock.cs
@@ -63,6 +63,17 @@ namespace Moq
 		/// <summary>
 		/// Creates an mock object of the indicated type.
 		/// </summary>
+		/// <typeparam name="T">The type of the mocked object.</typeparam>
+		/// <param name="behavior">Behavior of the mock.</param>
+		/// <returns>The mocked object created.</returns>
+		public static T Of<T>(MockBehavior behavior) where T : class
+		{
+			return new Mock<T>(behavior).Object;
+		}
+
+		/// <summary>
+		/// Creates an mock object of the indicated type.
+		/// </summary>
 		/// <param name="predicate">The predicate with the specification of how the mocked object should behave.</param>
 		/// <typeparam name="T">The type of the mocked object.</typeparam>
 		/// <returns>The mocked object created.</returns>
--- a/UnitTests/Linq/BehaviorFixture.cs
+++ b/UnitTests/Linq/BehaviorFixture.cs
@@ -1,0 +1,71 @@
+﻿namespace Moq.Tests.Linq
+{
+	using Xunit;
+
+	public class BehaviorFixture
+	{
+		[Fact]
+		public void NoQuery_Default()
+		{
+			var target = Mock.Of<IFoo>();
+			Assert.Equal(MockBehavior.Default, Mock.Get(target).Behavior);
+			Assert.False(target.BoolProperty1);
+		}
+
+		[Fact]
+		public void WithQuery_Default()
+		{
+			var target = Mock.Of<IFoo>(x => x.BoolProperty1 == true);
+			Assert.Equal(MockBehavior.Default, Mock.Get(target).Behavior);
+			Assert.True(target.BoolProperty1);
+		}
+
+		[Fact]
+		public void NoQuery_Strict()
+		{
+			var target = Mock.Of<IFoo>(MockBehavior.Strict);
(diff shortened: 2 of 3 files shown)


JohanLarsson:
Yes, that was in the original PR. I chopped out this part for a first small PR.
While not extremely useful it is still cleaner than `new Mock<T>().Object`. In fact this is the method I miss most often. But you do whatever you want of course.

stakx:
Exactly. I fail to see why anyone would ever need `Mock.Of<T>()` since, in order to set up such a mock, you'll have to go through `Mock.Get`, then use `.Setup`, so why not just start with `new Mock<T>()` in the first place.

`Mock.Of` actually has somewhat different semantics, so perhaps I am missing something here.

JohanLarsson:
Here is the use case:
Say we want to test this method:
```c#
public static int Foo(ILogger logger)
{
    if (logger == null)
    {
        throw new ArgumentNullException(nameof(logger));
    }

    try
    {
        ...
    }
    catch (Exception e)
    {
        logger.Error(e);
        throw;
    }
}
```

In the test for a happy path we expect the logger passed in to not be used. If there was no null check we would pass in `null` to make it clear that it is not used in the test. Now as there is a null check we do:

```c#
[Fact]
public void Test()
{
	Assert.Equal(1, Foo(Mock.Of<ILogger>(MockBehavior.Strict)));
}
```

Just a tad bit cleaner than `Assert.Equal(1, Foo(new Mock<ILogger>(MockBehavior.Strict).Object));` but a very common scenario.

stakx:
I'm afraid I still don't quite understand.

If you want to test the happy path, you need to pass in a logger alright. But it doesn't have to be a strict one.

If you want to explicitly test for failure, you are going to want to have a setup for the logger's `Error` method, so you'll need to retrieve the mock via `Mock.Get`, then `Setup`. See my reasoning above concerning this.

Of course you can just let the SUT method try to invoke a strict mock which isn't set up at all and will thus throw (causing the test to fail). It probably comes down to preferences; I would personally prefer a test to be more explicit about what exactly should fail.

I still think `Mock.Of<T>(MockBehavior)` is not a terribly great thing to have, but I would probably be willing to merge this anyway if the other overload (with an expression) is added to the PR. What I am trying to avoid is that this first one gets in, but not the other.

JohanLarsson:
> If you want to test the happy path, you need to pass in a logger alright. But it doesn't have to be a strict one.

Here we disagree, if there were no null check we would pass in null making it as clear as possible that it will not be used. When there is a null check passing in a strict mock with no setup is the clearest way I can think of to communicate that it is not supposed to be used when reading the test. Passing in a loose mock is not a good idea as the test will not fail if there is a bug and the logger is called. 

Maybe I made my sample too short for it to be clear, perhaps better if the logger is injected in a constructor with a null check.

I'll try to find time to add the other overload.

`MockBehavior.Strict` should have been the default but that is another discussion and that ship has sailed.

kzu:
@JohanLarsson if the test assertion asserts on the return value, the code will throw regardless of the usage of the mock logger. It adds nothing to the test. Depending on a side effect of the mocking library is not the best choice, I'd say.

If you expect a specific exception to *not* be thrown, you can use `Assert.DoesNotThrow()`, say, but assuming the *logger* will throw and cover that base, is wrong. Especially since by definition, loggers typically never throw at all (bad to fail an app just because logging is broken ;)).

JohanLarsson:
@kzu  I see now that I dumbed the sample a bit, made it too short. Does not change the fact that passing in strict mocks with no setup is common and useful.

informatorius:
If we can provide the **MockBehavior** in Mock.Of<>(), then can we also add **DefaultValue** and **SetupAllProperties**? 
So I can configure the Mock if it is deep and has properties setup all in one line and just get the mocked interface back?

stakx:
@informatorius: `Mock.Of<T>` always does a `SetupAllProperties` already, no need to expose that again. Regarding `DefaultValue`, I'm not sure it makes sense to expose this on the constructor. As far as I can tell, LINQ to Mocks was more or less programmed with `DefaultValue.Mock` in mind, at least with resepct to the expression originally passed in to the ctor. I can't even begin to guess what might break if we allowed an expression with deep setups *and*  `DefaultValue.Empty`. Therefore I don't think exposing `DefaultValue` is a good idea.

stakx:
I'm going to close this PR for the time being (for the reason given in my first review comment above).

PR #394, opened 2017-06-26
Add NeutralResourcesLanguage to Assembly info for portable library use

benbillbob:
This is to remove the following build warning when used in a UWP app.

Moq.dll : warning MSB3817: The assembly "...nuget\packages\Moq\4.7.63\lib\netstandard1.3\Moq.dll" does not have a NeutralResourcesLanguageAttribute on it. To be used in an app package, portable libraries must define a NeutralResourcesLanguageAttribute on their main assembly (ie, the one containing code, not a satellite assembly).


stakx:
Thank you!

PR #397, opened 2017-06-28
Allow `Mock<T>.Raise` to raise events on child mocks

### LINKED ISSUE #166: Mock<T>.Raise only raises events on root object
https://github.com/Moq/moq4/wiki/Quickstart#events
Example 2, "descendant events", doesn't function as expected

Events on the root object are raised when trying to raise events on the child.

This is noticed in 1/17/2013
http://stackoverflow.com/questions/14386540/mocked-inner-objects-events-will-not-fire

As tested in LINQPad:

``` c#
void Main()
{
    var mock = new Mock<IFoo> { DefaultValue = DefaultValue.Mock };
    mock.Object.Event += (s,e) => "FIRED FOO".Dump();
    mock.Object.Bar.Event += (s,e) => "FIRED BAR".Dump();

    mock.Raise(x => x.Event += null, EventArgs.Empty);
    mock.Raise(x => x.Bar.Event += null, EventArgs.Empty);
}

public interface IFoo
{
    IBar Bar { get; }
    event EventHandler Event;
}

public interface IBar
{
    event EventHandler Event;
}
```

Expected output:

```
FIRED FOO
FIRED BAR
```

Actual output:

```
FIRED FOO
FIRED FOO
```

All events named "Event" raise IFoo.Event.

``` c#
mock.Raise(x => x.Bar.Child.Another.Event += null, EventArgs.Empty);
```

Returns:

```
FIRED FOO
```

stakx:
This fixes #166.

### DIFF
--- a/Source/Mock.Generic.cs
+++ b/Source/Mock.Generic.cs
@@ -503,7 +503,7 @@ namespace Moq
 
 			try
 			{
-				this.DoRaise(ev, args);
+				ev.Target.DoRaise(ev.MemberInfo, args);
 			}
 			catch (Exception e)
 			{
@@ -521,7 +521,7 @@ namespace Moq
 
 			try
 			{
-				this.DoRaise(ev, args);
+				ev.Target.DoRaise(ev.MemberInfo, args);
 			}
 			catch (Exception e)
 			{
--- a/UnitTests/MockedEventsFixture.cs
+++ b/UnitTests/MockedEventsFixture.cs
@@ -341,7 +341,7 @@ namespace Moq.Tests
 		}
 
 		[Fact]
-		public void DoesNotRaiseEventOnSubObject()
+		public void CanRaiseEventOnSubObject()
 		{
 			var mock = new Mock<IParent> { DefaultValue = DefaultValue.Mock };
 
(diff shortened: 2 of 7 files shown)


Source/Mock.Generic.cs
```diff
@@ -503,7 +503,7 @@ public void Raise(Action<T> eventExpression, EventArgs args)
 
 			try
 			{
-				this.DoRaise(ev, args);
+				ev.Target.DoRaise(ev.MemberInfo, args);
```
stakx: Replacing `this` with `ev.Target` is what really fixes this issue.

UnitTests/MockedEventsFixture.cs
```diff
@@ -352,7 +352,7 @@ public void DoesNotRaiseEventOnSubObject()
 
 			mock.Raise(p => p.Adder.Added += null, EventArgs.Empty);
 
-			Assert.False(raised);
+			Assert.True(raised);
```
stakx: Note that this is in effect a functional change. The changes of this breaking client code are probably quite small (see https://github.com/Moq/moq4/issues/166#issuecomment-311574982).

At the same time, replacing this unit test with its exact opposite also has the advantage of fixing an inconsistency, since the quickstart guide already documents the ability to raise events on sub-objects.

Issue #373, opened 2017-06-13
Set Times expectation on Setup


skarllot:
Would be nice if I can set times expectation on Setup instead of separated Verify. e.g.:

```csharp
mockInstance
    .Setup(a => a.Start(It.IsInRange(0, 1000)))
    .Count(Times.AtLeast(5))
    .Returns(Task.FromResult(true));
```

tparikka:
+1

stakx:
Being able to specify `Times` on a setup sounds like a handy feature. Contrary to that, two extant setup methods `.AtMostOnce()` and `.AtMost(n)` that are related in purpose were both deprecated and hidden from IntelliSense years ago. This old [Google Code issue 216](https://code.google.com/archive/p/moq/issues/216) states why:

> These were hidden as we're encouraging people to manage the former occurrence constraints via explicit Verify assertions, which aligns better with AAA style testing.

I'm not against this feature, but I'm digging up this bit of history so we won't end up swinging back and forth between two opposing design decisions.

@kzu: What is your position on the ability to specify `Times` during setup? Do you still consider this a feature that should belong in `Verify` exclusively?

IanYates83:
While I completely agree with the AAA comment, the reason I submitted #401 is to cover scenarios like my `CannotPostVerifyCallsByExpressionForModifiedReferenceTypeArgument` test, where the post-Act `Verify` cannot be used to check the invocation argument values.

I did play around with having a `Verify` directly on `IFluentInterface`, which gives a better AAA approach, but it felt like a larger change and less elegant when you have many setups.

Example:
```
var setup1 = mock.Setup(m => m.Call1(It.Is<MyClass>(c => c.Name == "test"))).Returns("result");

Act();

setup1.Verify(Times.Exactly(2));
```

I realise that this is slightly an edge-case issue.
We currently solve it through argument capture and comparison, which is very inelegant.

stakx:
@IanYates83 - that makes sense, I've come across one other issue where someone had the same problem with reference type state changing and not being verifiable after the fact. I'm generally in favour of your proposed changes. Like I already [commented on your PR](https://github.com/moq/moq4/pull/401#pullrequestreview-47930518)

* Can you perhaps add your thoughts on why you chose to add a `Times` on the `.Verifiable()` method? Did you consider any alternatives (extending the `IOccurrence` interface; extending `.Setup()`; adding a new method such as `.Count()` as proposed by @skarllot; etc.)?

[...]

PR #401, opened 2017-06-30
Can specify Times a verifiable expectation must be met

IanYates83:
Defaults to `AtLeastOnce` to match current functionality.

We had an issue testing a method modified properties within a reference type shared to a dependency. The current "Verify" checks can only compare against the current reference type state (see new `CannotPostVerifyCallsByExpressionForModifiedReferenceTypeArgument` fact).
The approach that we came up with was to support the "Times" concept directly in the "Verifiable" setup.

Given that the change is non-breaking by default and pretty clean, I thought it worth pushing back.

Full example in `CanPreVerifyExactCallsByExpressionForModifiedReferenceTypeArgument` fact, but in short:
```
mock.Setup(m => m.Method()).Verifiable(Times.Once()); // Require method to be called only once
// act
mock.Verify(); // Verify invocations match expected
```

### DIFF
--- a/Source/IProxyCall.cs
+++ b/Source/IProxyCall.cs
@@ -50,7 +50,7 @@ namespace Moq
 		bool IsConditional { get; }
 		string FailMessage { get; set; }
 		bool Invoked { get; set; }
-		bool IsVerifiable { get; set; }
+		Times? VerifiableTimes { get; set; }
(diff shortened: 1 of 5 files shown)


stakx:
This will probably have to wait a while until we have an answer in #373. Thanks for being patient! :)

Issue #409, opened 2017-07-10
Internal interception on a mock changes the DefaultValue property


vladonemo:
Let's consider this use case:

```C#		
public void DefaultValueIsNotChangedWhenPerformingInternalInvocation()
{
	var mockBar = new Mock<IBar> { DefaultValue = DefaultValue.Empty };
	var mockFoo = new Mock<IFoo>();
	mockFoo.SetupSet(m => m.Bar = mockBar.Object);
	Assert.Equal(DefaultValue.Empty, mockBar.DefaultValue);
}
```

The SetupSet changes the DefaultValue of the mockBar to DefaultValue.Mock.

I'm submitting a PR shortly.

stakx:
Your fix has been merged. Thank you for correcting this! :)

PR #410, opened 2017-07-10
Fix for the #409.

vladonemo:


Source/FluentMockContext.cs
```diff
@@ -90,6 +90,7 @@ public MockInvocation LastInvocation
 
 		public void Dispose()
 		{
+			invocations.Reverse();
```
stakx: Out of curiosity (I haven't yet looked at this in detail), can you briefly explain why this works?

UnitTests/MockDefaultValueProviderFixture.cs
```diff
@@ -93,6 +93,15 @@ public void CreatedMockIsVerifiedWithOwner()
 			Assert.Throws<MockVerificationException>(() => mock.Verify());
 		}
 
+		[Fact]
+		public void DefaultValueIsNotChangedWhenPerformingInternalInvocation()
+		{
+			var mockBar = new Mock<IBar> { DefaultValue = DefaultValue.Empty };
```
stakx: Given that `DefaultValue` is set to `DefaultValue.Empty`, would this test probably be more at home in `EmptyDefaultValueProviderFixture.cs`?

stakx:
I'm sorry but this cannot be merged as is. Let's try this with a fresh start. Please check out a new feature branch splitting of from the current tip of this repo's `develop` branch. Then add the changes from commit 4b02b918ebc54be69b80a5613739892080d6fb7e to that feature branch (e.g. by `cherry-pick`ing). Then submit a new PR from that feature branch to Moq's `develop` branch. Thank you!

UnitTests/MockDefaultValueProviderFixture.cs
vladonemo: I think that the MockDefaultValueProvuderFixture is better place for this test, as the EmptyDefalutValueProviderFixture tests the EmptyDefaultValueProvider class - unrelated to the scenario.

Source/FluentMockContext.cs
vladonemo: I explained this in the new PR.

vladonemo:
I have submitted a new PR - #411. I hope it is correct now :) Sorry for the mess ...

PR #411, opened 2017-07-11
Fix for the #409

vladonemo:
Ok, trying again. I hope that this PR is now OK - there are only 2 files changed.

The fix for #409 is to revert the order of the items of the invocations list in he FluentMockContext before disposing the items. The new test runs into a situation, that there are at least 2 MockInvocations created on the mock which is constructed with DefaultValue.Empty. The first one remembers the original DefalutValue (Empty) and sets it to Mock. The second one remembers the current value of DefaultValue (Mock) and sets it to the Mock, too. Now, if the MockInvocations are disposed in the same order as they were created, the first MockInvocation will set the DefaultValue of the mock back to the value it remembered - Empty. Then the second MockInvocation is disposed and sets the DefaultValue to the Mock, as this is the one that it remembered. By reverting the disposal order we make sure that the DefaultValue is reverted back to the original value before the first MockInvocation was constructed.

### DIFF
--- a/Source/FluentMockContext.cs
+++ b/Source/FluentMockContext.cs
@@ -90,6 +90,7 @@ namespace Moq
 
 		public void Dispose()
 		{
+			invocations.Reverse();
 			foreach (var invocation in invocations)
 			{
 				invocation.Dispose();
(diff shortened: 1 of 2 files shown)


stakx:
Thanks for cleaning this up, this looks *much* better! :+1: I'll look at this in detail later today.

PR #415, opened 2017-07-12
Improve method match accuracy in `ExtractProxyCall`

### LINKED ISSUE #131: The result of setting up methods in an hierarchy of interfaces depends on the order by which they're set up
**Test Types**:

``` csharp
public interface BaseA
{
    string Method();
}

public interface BaseB
{
    string Method();
}

public interface Derived : BaseA, BaseB
{
    new string Method();
}
```

**Reproduce bug:**

``` csharp
var mock = new Mock<Derived>();

mock.Setup(x => x.Method())
    .Returns("a");
mock.As<BaseA>().Setup(x => x.Method())
    .Returns("b");
mock.As<BaseB>().Setup(x => x.Method())
    .Returns("c");

var derived = mock.Object;
Assert.Equal("a", derived.Method());            //equals "c" <---- Test fails
Assert.Equal("b", (derived as BaseA).Method()); //equals "b"
Assert.Equal("c", (derived as BaseB).Method()); //equals "c"
```

If I change the order of the setups, so that `Derived.Method` is configured last, the program now prints "a", "b", "c" as expected and the tests pass.

``` csharp
var mock = new Mock<Derived>();

mock.As<BaseA>().Setup(x => x.Method())
    .Returns("b");
mock.As<BaseB>().Setup(x => x.Method())
    .Returns("c");
mock.Setup(x => x.Method())
    .Returns("a");

var derived = mock.Object;
Assert.Equal("a", derived.Method());            //equals "a"
Assert.Equal("b", (derived as BaseA).Method()); //equals "b"
Assert.Equal("c", (derived as BaseB).Method()); //equals "c"
```

stakx:
If you're setting up methods in a type hierarchy (e.g. `Base.Method` and `Derived.Method`), the setup order currently matters, even though two different methods are set up.

This commits is a first step in making Moq associate the "right" setup with a proxy method invocation. This will make sure that if there are several setups that potentially match an invocation, it will attempt to pick the one for the exact same method.

This is only a start, but probably still not good enough in all situations (for example, for overridden methods, the distance of the inheritance relationship is not taken into account).

This should fix #131.

/cc @dcastro, @pjquirk

### DIFF
--- a/Source/InterceptorStrategies.cs
+++ b/Source/InterceptorStrategies.cs
@@ -102,7 +102,28 @@ namespace Moq
 
 		public InterceptionAction HandleIntercept(ICallContext invocation, InterceptorContext ctx, CurrentInterceptContext localctx)
 		{
-			localctx.Call = FluentMockContext.IsActive ? (IProxyCall)null : ctx.OrderedCalls.LastOrDefault(c => c.Matches(invocation));
+			if (FluentMockContext.IsActive)
+			{
+				localctx.Call = null;
+			}
+			else
+			{
+				IProxyCall lastMatchingSetup = null;
+				IProxyCall lastMatchingSetupForSameMethod = null;
+				foreach (var setup in ctx.OrderedCalls)
+				{
+					if (setup.Matches(invocation))
+					{
+						lastMatchingSetup = setup;
+						if (setup.Method == invocation.Method)
+						{
+							lastMatchingSetupForSameMethod = setup;
+						}
+					}
+				}
+				localctx.Call = lastMatchingSetupForSameMethod ?? lastMatchingSetup;
+			}
+
 			if (localctx.Call != null)
 			{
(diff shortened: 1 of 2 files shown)


Source/InterceptorStrategies.cs
```diff
@@ -102,7 +102,28 @@ internal class ExtractProxyCall : IInterceptStrategy
...
+					{
+						lastMatchingSetup = setup;
+						if (setup.Method == invocation.Method)
+						{
+							lastMatchingSetupForSameMethod = setup;
+						}
+					}
+				}
+				localctx.Call = lastMatchingSetupForSameMethod ?? lastMatchingSetup;
+			}
```
stakx: At first glance, this looks much more involved than a simple `LastOrDefault`, but it has the advantage of only performing a single enumeration over `ctx.OrderedCalls` while looking for two different items.

Issue #421, opened 2017-07-14
Passing mocked IEnumerable in Setup of method call causes NullReferenceException


kgybels:
After upgrading from Moq 4.7.1 to 4.7.63, passing a mocked IEnumerable in the setup of a method call causes a NullReferenceException.

```C#
[Fact]
public void SetupShouldAcceptMockedIEnumerableInMethodCall()
{
	var mock = new Mock<IFoo>();
	var values = new Mock<IEnumerable>().Object;
	mock.Setup(m => m.Bar(values));
}
public interface IFoo
{
	void Bar(IEnumerable values);
}
```

kgybels:
This issue is already fixed on the develop branch.

stakx:
Duplicate of #169.

Issue #169, opened 2015-05-06
NullReferenceException when argument implements IEnumerable


FrolYo:
if I make IInput not implementing IEnumerable, then NullRef does not occur.

```
using System.Collections.Generic;

using Microsoft.VisualStudio.TestTools.UnitTesting;

using Moq;

[TestClass]
public class FooTest
{
    [TestMethod]
    public void ShouldCalculate()
    {
        var mocks = new MockRepository(MockBehavior.Loose);
        var service = mocks.Create<IDependency>();

        var input1 = mocks.OneOf<IInput>();
        service.Setup(x => x.Run(input1)).Returns(1);

        var input2 = mocks.OneOf<IInput>();
        service.Setup(x => x.Run(input2)).Returns(2);

        var foo = new Foo(service.Object);
        foo.Calculate(input1);
        foo.Calculate(input2);
    }
}

public interface IDependency
{
    decimal Run(IInput input);
}

public interface IInput : IEnumerable<int>
{
}

public class Foo
{
    private readonly IDependency dependency;

    public Foo(IDependency dependency)
    {
        this.dependency = dependency;
    }

    public void Calculate(IInput input)
    {
        this.dependency.Run(input);
    }
}
```

stakx:
This appears to be remotely related to #22. Except here, the mock apparently has a setup for `IEnumerable.GetEnumerator`, but not for `IEnumerable<T>.GetEnumerator` (and not the other way around).

stakx:
Note to self: This error is due to `Interceptor.AddCall` creating an `ExpressionKey`, which involves converting the method setup to a string representation using `ExpressionStringBuilder`, which in turn tries to enumerate the mock object (and since `IEnumerable` isn't correctly set up, this will fail due to a small bug). Getting rid of the string representation would help (as suggested by https://github.com/moq/moq4/issues/328#issuecomment-285655952) but I'm not sure how that can be done. The simplest solution is to fix the enumeration in `ExpressionStringBuilder`.

stakx:
This will be fixed in the next release of Moq (version >4.7.63).

PR #422, opened 2017-07-14
Add regression test for issue #421

kgybels:
Using a mocked IEnumerable in the setup of a method call shouldn't cause
any exceptions.


stakx:
As you've notied, this has already been fixed, and there are already regression tests that check for this (see #169). However, I think your proposed test is still a good addition because it is much terser than the preexisting one. Merging.

Thank you!

PR #454, opened 2017-09-23
Add support for void sequences

alexbestul:


alexbestul:
Thanks for the thorough feedback. The changes all sound perfectly reasonable and easily doable (with the possible exception of Async--more on that shortly). I'll make them and update this PR.

I did make a very brief attempt at Async support before submitting my initial PR. I ran into problems because Async methods must return a `Task`, and my code started to conflict with the existing `SetupSequence<TMock, TResult>` implementation. I'm worried that a complete/correct implementation might require creating separate `SetupSequence` and `SetupSequenceAsync` APIs. As I said, though, my first attempt was very brief--I'll give it a more sincere effort, and see if I can come up with something.

stakx:
Don't worry too much about the async bit, it can be dealt with later. Async support in Moq is somewhat spotty already, and there's still an open issue about cleaning it all up... I'll need to get back on that soon.

alexbestul:
The feedback in this PR has been addressed in #463. Any further comments can be posted there.

PR #468, opened 2017-10-03
Add new `Callback` and `Returns` overloads for setting up methods with `ref` parameters

### LINKED ISSUE #105: How can I mock a function with ref parameter?
I have this simple code, but I can't do it work

```
[TestMethod]
        public void TestMoq()
        {
            var mock = new Moq.Mock<Hola>();            

            string o = "1";
            mock.Setup(x => x.DoSomething(ref o)).Callback((string a) => o = "2");

            mock.Object.DoSomething(ref o);

        }
```

```
public class Hola
    {
        public Hola()
        {

        }

        public virtual void DoSomething(ref string a)
        {

        }
    }
```

The error is this:
capturemoqbug

I prefer to use the Moq library. But only I get it on the Rhino Mocks library, with the function **OutRef**:

```
[TestMethod]
        public void TestRhinoMocks()
        {
            MockRepository mocker = new MockRepository();

            Hola mock = mocker.StrictMock<Hola>();

            string o = "1";

            Expect.Call(() => mock.DoSomething(ref o)).OutRef("2");

            mocker.ReplayAll();

            //i.Setup(x => x.DoSomething(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>())).Callback((string a, string b, string c) => o = "2");

            mock.DoSomething(ref o);
        }
```

Thanks in advance!

stakx:
This PR proposes new method overloads for the `Callback` and `Returns` setup method that would allow setting up callbacks for methods having `ref` parameters. They do this by accepting delegates of any delegate type (including such ones having `ref` parameters); the present overloads only accept `Action`s or `Func`s, respectively.

This would resolve #105.


ajbeaven:
I think it was due to this change that a previously passing test started failing after upgrading from 4.7.145 to 4.8.0:
```
public interface IUserRepository
{
  List<TUser> GetUsers<TUser>(UserSearch search) where TUser : User;
  List<TUser> GetUsers<TUser>(UserSearch search, out int totalUsers) where TUser : User;
}

int totalUsers;
unitOfWorkMock.Setup(x => x.Users.GetUsers(It.IsAny<UserSearch>())).Returns(new List<User>());
var result = unitOfWorkMock.Object.Users.GetUsers(new UserSearch(), out totalUsers); // result is null
```

The fix was to specify the out parameter in the call to `Setup`.

I'm not sure if this warrants a **Breaking change** note on the changelog, but it would have saved me a bit of confusion when trying to figure out why my tests started failing.

stakx:
Not sure either, but regardless of whether it's a breaking change or not, it seems to me that your test code used to be written incorrectly (setting up the wrong overload), and adding the second parameter to the setup was the correct fix. Do you agree, or have I misunderstood something?

ajbeaven:
Yes you're exactly right and I completely agree. I suppose it probably shouldn't be considered a breaking change if the test was only passing due to a bug, so feel free to ignore this. Thanks for all your contributions to moq, its awesome!

stakx:
@ajbeaven - I'll look at your scenario more closely as soon as I get a spare moment, but unless it turns out to be an unexpected corner case that I overlooked back then, I'll probably leave things as they are now for the reason you mentioned.

Glad to hear that Moq continues to be useful to you!

PR #476, opened 2017-10-06
Reimplement `SetupSequence` to make it thread-safe

### LINKED ISSUE #467: SetupSequence() is not thread-safe
`SetupSequence()` is not thread-safe. If two threads call the mock, it can return the same value more than once. The following failing test demonstrate it:

```csharp
using System.Collections.Concurrent;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using Moq;

namespace MoqTests
{
    [TestClass]
    public class SetupSequenceTests
    {
        private interface ITest
        {
            int Method();
        }

        [TestMethod]
        public async Task Test()
        {
            const int length = 20;

            var m = new Mock<ITest>();

            // Setups the mock to return a sequence of results.
            var seqSetup = m.SetupSequence(x => x.Method());
            for (int i = 1; i <= length; i++)
            {
                seqSetup = seqSetup
                    .Returns(i);
            }
            
            // Queue to store the invocations of the mock.
            var invocationsQueue = new ConcurrentQueue<int>();

            // We will invoke the mock from two different tasks.
            var t1 = Task.Run(() => RegisterInvocations(m, invocationsQueue));
            var t2 = Task.Run(() => RegisterInvocations(m, invocationsQueue));
            await Task.WhenAll(t1, t2);

            var orderedInvocations = invocationsQueue.OrderBy(x => x).ToArray();

            // The assertion prints the real invocations, you will see duplicates there!
            CollectionAssert.AreEqual(
                Enumerable.Range(1, length).ToArray(),
                orderedInvocations,
                string.Join(", ", orderedInvocations.Select(x => x.ToString())));
        }

        // Calls the mock until it returns default(int)
        private static void RegisterInvocations(IMock<ITest> m, ConcurrentQueue<int> invocationsQueue)
        {
            int result;
            do
            {
                result = m.Object.Method();
                if (result != default(int))
                {
                    invocationsQueue.Enqueue(result);
                }
            } while (result != default(int));
        }
    }
}
```

stakx:
~~This is an early draft and shouldn't be merged just yet.~~

~~It does appear to resolve #467, however since it's a complete reimplementation of `SetupSequence` which works in a completely different way than the previous implementation, it needs much more thorough testing of edge cases, as well as a code cleanup.~~

This resolves #467.


Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+
+namespace Moq.Language.Flow
+{
+	// keeping the fluent API separate from `SequenceMethodCall` saves us from having to
+	// define a generic variant `SequenceMethodCallReturn<TResult>`, which would be much more
+	// work that having a generic fluent API counterpart `SequenceMethodCall<TResult>`.
```
stakx: This demonstrates a pattern that should probably be applied more in Moq. Moq's "plumbing" classes (e.g. `MethodCall`, `MethodCallReturn<TResult>`, `Mock`) should be made largely non-generic and kept strictly separate from the generic classes that are part of the fluent API. These latter types only flow the context of a fluent API "phrase" (i.e. a chain of method calls starting with `mock.Setup…` and ending with the end-of-line semicolon). That way, we could probably get rid of some code duplication.

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		public ISetupSequentialAction Pass()
+		{
+			this.setup.AddPass();
+			return this;
+		}
+
+		public ISetupSequentialAction Throws<TException>()
+			where TException : Exception, new()
+			=> this.Throws(new TException());
```
stakx: Perhaps delayed instantiation should be preferred over newing up exception objects way in advance?

Source/Mock.cs
```diff
@@ -768,7 +768,7 @@ private static Expression GetPropertyExpression(Type mockType, PropertyInfo prop
 		/// Gets the interceptor target for the given expression and root mock, 
 		/// building the intermediate hierarchy of mock objects if necessary.
 		/// </summary>
-		private static Interceptor GetInterceptor(Expression fluentExpression, Mock mock)
+		internal static Interceptor GetInterceptor(Expression fluentExpression, Mock mock)
```
stakx: Making this and the following helper methods `internal` was a quick hack to see if the reimplementation was feasible at all. This needs to be cleaned up. Do these methods actually have to be located in the `Mock` class?

Source/SequenceExtensions.cs
```diff
@@ -2,7 +2,10 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq.Expressions;
 using System.Threading.Tasks;
+using System.Reflection;
 using Moq.Language;
+using Moq.Language.Flow;
+using System.Linq;
```
stakx: (Detail:) Sort `using`s properly.

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
```
stakx: (Detail:) Add missing license text to all new files!

Source/SequenceExtensions.cs
```diff
@@ -20,7 +23,28 @@ public static class SequenceExtensions
 			Expression<Func<TMock, TResult>> expression)
 			where TMock : class
```
stakx: This new implementation of `SetupSequence` is really ugly at the moment and needs more work. Can this be cleaned up somehow? And does this (method calls and property getters) cover all cases? What about e.g. indexers? Do matchers work correctly if we install a setup like this? I don't actually understand just what is happening here. :-/

Source/SequenceMethodCall.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace Moq
+{
+	// this corresponds to a setup created by `mock.SetupSequence`
+	internal class SequenceMethodCall : MethodCall
+	{
+		// contains the responses set up with `CallBase`, `Pass`, `Returns`, and `Throws`
+		// TODO: the use of a queue without keeping track of the number of received calls, and tying each
+		// response to an expected invocation count, means that things won't work quite like they used to!
+		// this is possibly a breaking change right now.
+		private ConcurrentQueue<Tuple<ResponseKind, object>> responses;
```
stakx: The mentioned todo should be taken seriously. With the new implementation, we can exhaust the set up sequence, then receive default values for a while, then configure some more responses, and get those back immediately until the sequence is exhausted again. This might have worked differently with the old implementation, because it kept track of the exact call count. At least add a unit test verifying this scenario before and after the change.

Source/SequenceMethodCall.cs
```diff
@@ -0,0 +1,97 @@
...
+
+		public SequenceMethodCall(Mock mock, Condition condition, Expression originalExpression, MethodInfo method, params Expression[] arguments)
```
stakx: Not sure if we need all these parameters, e.g. `condition`. Could probably be shortened.

Source/SequenceMethodCall.cs
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		public void AddThrows(Exception exception)
+		{
+			this.responses.Enqueue(Tuple.Create(ResponseKind.Throws, (object)exception));
+		}
+
+		public override void Execute(ICallContext call)
+		{
+			base.Execute(call);
```
stakx: I've quickly gone over the base implementation of `Execute`. Since the fluent API started by `SetupSequence` doesn't allow setting up any explicit `Callback`s, `Return`s, `Raises`, or `Throws`, the base implementation should be largely inert, except for increasing its own `base.CallCount` and setting `base.Invoked`. But it makes me curious how the new `SetupSequence` implementation interacts with `mock.Verify`? Perhaps add some unit tests about that before and after the change.

Source/SequenceMethodCall.cs
```diff
@@ -0,0 +1,97 @@
...
+
+					case ResponseKind.Throws:
+						throw (Exception)response.Item2;
+				}
+			}
+			else
+			{
+				// we get here if there are more invocations than configured responses.
+				// if the setup method does not have a return value, we don't need to do anything;
+				// if it does have a return value, we produce the default value.
```
stakx: Verify (via additional unit tests) that the new implementation acts the same way as the old implementation when the configured responses have been exhausted by too many invocations.

Source/SequenceMethodCall.cs
```diff
@@ -0,0 +1,97 @@
...
+
+				if (call.Method.ReturnType == typeof(void))
+				{
+
+				}
+				else if (call.Method.ReturnType.GetTypeInfo().IsValueType)
+				{
+					call.ReturnValue = Activator.CreateInstance(call.Method.ReturnType);
```
stakx: AFAIK other parts of Moq do this in a more complicated fashion, because of nullable value types, COM types, etc. Check whether this is sufficient.

UnitTests/Regressions/IssueReportsFixture.cs
```diff
@@ -1631,6 +1631,63 @@ public class GenericClass2<T> : BaseType<T>
...
+
+		public class Issue467
+		{
+			public interface ITest
+			{
+				int Method();
+			}
+
+			[Fact]
+			public async Task Test()
```
stakx: Give this test a more sensible name.

UnitTests/Regressions/IssueReportsFixture.cs
```diff
@@ -1631,6 +1631,63 @@ public class GenericClass2<T> : BaseType<T>
...
+				const int length = 20;
+
+				var m = new Mock<ITest>();
+
+				// Setups the mock to return a sequence of results.
+				var seqSetup = m.SetupSequence(x => x.Method());
+				for (int i = 1; i <= length; i++)
+				{
+					seqSetup = seqSetup
+						.Returns(i);
```
stakx: Unnecessary line break, reformat.

UnitTests/Regressions/IssueReportsFixture.cs
```diff
@@ -1631,6 +1631,63 @@ public class GenericClass2<T> : BaseType<T>
...
+
+				// We will invoke the mock from two different tasks.
+				var t1 = Task.Run(() => RegisterInvocations(m, invocationsQueue));
+				var t2 = Task.Run(() => RegisterInvocations(m, invocationsQueue));
+				await Task.WhenAll(t1, t2);
+
+				var orderedInvocations = invocationsQueue.OrderBy(x => x).ToArray();
+
+				// The assertion prints the real invocations, you will see duplicates there!
+				Assert.Equal(
```
stakx: @icalvo's original repro code used MSTest's `CollectionAssert.AreEqual`. Make sure that this is indeed xUnit.NET's way of comparing two sequences.

Source/Mock.Generic.cs
```diff
@@ -315,6 +315,25 @@ public Mock<T> SetupAllProperties()
 			return this;
 		}
 
+		/// <summary>
+		/// Return a sequence of values, once per call.
+		/// </summary>
+		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
+		public ISetupSequentialResult<TResult> SetupSequence<TResult>(
+			Expression<Func<T, TResult>> expression)
```
stakx: Unnecessary line break, remove.

Source/Mock.cs
```diff
@@ -38,10 +38,8 @@
 //[This is the BSD license, see
 // http://www.opensource.org/licenses/bsd-license.php]
 
-using System.Collections.Concurrent;
-using Moq.Properties;
-using Moq.Proxy;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
```
stakx: Sort `using`s properly.

Source/Mock.cs
```diff
@@ -667,6 +669,44 @@ private static Expression GetValueExpression(object value, Type type)
 			return Expression.Convert(Expression.Constant(value), type);
 		}
 
+		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
+		internal static SetupSequencePhrase<TResult> SetupSequence<T, TResult>(Mock<T> mock, Expression<Func<T, TResult>> expression)
+			where T : class
```
stakx: Does this method really have to be generic? Why not make the parameters be of types `Mock` and `LambdaExpression`? The public method in `Mock<T>` can make sure that everything is appropriately typed.

Source/Mock.cs
```diff
@@ -667,6 +669,44 @@ private static Expression GetValueExpression(object value, Type type)
...
+				var methodCall = expression.GetCallInfo(mock);
+				var targetInterceptor = GetInterceptor(methodCall.Object, mock);
+				var setup = new SequenceMethodCall(mock, expression, methodCall.Method, methodCall.Arguments.ToArray());
+				targetInterceptor.AddCall(setup, SetupKind.Other);
+				return new SetupSequencePhrase<TResult>(setup);
+			}
+		}
+
+		internal static SetupSequencePhrase SetupSequence<T>(Mock<T> mock, Expression<Action<T>> expression)
+			where T : class
```
stakx: Same here: Does this method really have to be generic? Why not make the parameters be of types `Mock` and `LambdaExpression`? The public method in `Mock<T>` can make sure that everything is appropriately typed.

Source/Obsolete/SequenceExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+		/// <summary>
+		/// Return a sequence of values, once per call.
+		/// </summary>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Please use instance method Mock<T>.SetupSequence instead.")]
+		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
+		public static ISetupSequentialResult<TResult> SetupSequence<TMock, TResult>(
+			this Mock<TMock> mock,
+			Expression<Func<TMock, TResult>> expression)
+			where TMock : class
```
stakx: `TMock` should be just `T`, but changing a generic type parameter's name is possibly a (minor) breaking change. Leave as is for now.

Source/Obsolete/SequenceExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		/// <summary>
+		/// Performs a sequence of actions, one per call.
+		/// </summary>
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Please use instance method Mock<T>.SetupSequence instead.")]
+		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By Design")]
+		public static ISetupSequentialAction SetupSequence<TMock>(this Mock<TMock> mock, Expression<Action<TMock>> expression)
+			where TMock : class
```
stakx: Same here: `TMock` should be just `T`, but changing a generic type parameter's name is possibly a (minor) breaking change. Leave as is for now.

Source/SequenceMethodCall.cs
stakx: Done. Added a unit test that ensures `Verify` can correctly check a sequence's invocation count.

Source/SequenceMethodCall.cs
stakx: The new implementation acts differently; see explanation in https://github.com/moq/moq4/issues/467#issuecomment-337821836. I personally favour the new behaviour as it brings `SetupSequence` more in line with how other `Setup` methods work: unless a setup is a conditional setup (`.When(...)`), then it overrides any previous setup for the same invocation.

PR #494, opened 2017-10-21
Various refactorings and optimizations

stakx:


Source/MethodCall.cs
```diff
@@ -129,6 +123,11 @@ public MethodCall(Mock mock, Condition condition, Expression originalExpression,
 						throw new NotSupportedException(Resources.OutExpressionMustBeConstantValue);
 					}
 
+					if (outValues == null)
+					{
+						outValues = new List<KeyValuePair<int, object>>();
+					}
```
stakx: `outValues` is no longer guaranteed to be pre-allocated, so we need to check here. This won't hurt a lot, since it's only checked once per `out` parameter (of which there usually aren't many). Also, we're in a ctor, so no need to worry about thread safety.

Source/MethodCall.cs
```diff
@@ -151,39 +150,29 @@ public MethodCall(Mock mock, Condition condition, Expression originalExpression,
...
 		{
-			get { return condition != null; }
+			get => this.failMessage;
+			set => this.failMessage = value;
 		}
 
-		public bool IsVerifiable { get; set; }
+		bool IProxyCall.IsConditional => this.condition != null;
 
-		public bool Invoked { get; set; }
```
stakx: This doesn't need to have a backing field. `Invoked` can be derived from the call count (see redefinition below).

Source/MethodCall.cs
```diff
@@ -151,39 +150,29 @@ public MethodCall(Mock mock, Condition condition, Expression originalExpression,
...
-		public MethodBase TestMethod { get; private set; }
+		bool IProxyCall.Invoked => this.callCount > 0;
 
-		public Expression SetupExpression
-		{
-			get { return this.originalExpression; }
-		}
-
-		public int CallCount { get; private set; }
+		MethodInfo IProxyCall.Method => this.method;
```
stakx: Note that most properties required by `IProxyCall` are now explicitly implemented to emphasize that they're not actually needed internally.

Source/MethodCall.cs
```diff
@@ -253,67 +245,44 @@ public void EvaluatedSuccessfully()
 
 		public virtual void Execute(ICallContext call)
 		{
-			this.Invoked = true;
```
stakx: The `Invoked` property's value now gets computed from `callCount`, no need to set it anymore.

Source/MethodCall.cs
```diff
@@ -253,67 +245,44 @@ public void EvaluatedSuccessfully()
...
+			this.throwExceptionResponse = exception;
 			return this;
 		}
 
 		public IThrowsResult Throws<TException>()
 			where TException : Exception, new()
 		{
-			this.thrownException = new TException();
-			return this;
+			return this.Throws(new TException());
```
stakx: DRY. This should have been an extension method in the first place. 😞

Source/MethodCall.cs
```diff
@@ -81,42 +81,36 @@ public IVerifies Raises(Action<TMock> eventExpression, params object[] args)
...
 	internal partial class MethodCall : IProxyCall, ICallbackResult, IVerifies, IThrowsResult
 	{
-		// Internal for AsMockExtensions
-		private Expression originalExpression;
-		private Exception thrownException;
-		private Action<object[]> setupCallback;
-		private List<IMatcher> argumentMatchers = new List<IMatcher>();
-		private EventInfo mockEvent;
-		private Delegate mockEventArgsFunc;
-		private object[] mockEventArgsParams;
```
stakx: These 3 event fields get replaced by a single one, `raiseEventResponse`. This means that for all setups who do *not* raise some event (which is probably most of them), we save two fields. (And the few setups that *do* raise some event will incur one additional allocation.)

Source/MethodCall.cs
```diff
@@ -81,42 +81,36 @@ public IVerifies Raises(Action<TMock> eventExpression, params object[] args)
...
+		private List<IMatcher> argumentMatchers;
+		private Action<object[]> callbackResponse;
+		private int callCount;
+		private Condition condition;
 		private int? expectedMaxCallCount;
```
stakx: This could probably be turned into an `int` and initialized to `int.MaxValue`, let's not for the moment.

Source/MethodCall.cs
```diff
@@ -151,39 +150,29 @@ public MethodCall(Mock mock, Condition condition, Expression originalExpression,
...
+		MethodInfo IProxyCall.Method => this.method;
 
-		protected internal Mock Mock { get; private set; }
+		Expression IProxyCall.SetupExpression => this.originalExpression;
 
 		[Conditional("DESKTOP")]
 		[SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
 		private void SetFileInfo()
 		{
+#if !NETCORE
```
stakx: moving the `#if` guard around the `try` block instead of inside it means that the method no longer contains any exception handling clauses and becomes eligible for inlining. (But then it's omitted anyway.) Still makes more sense when reading source code this way.

Source/MethodCall.cs
```diff
@@ -195,24 +184,27 @@ private void SetFileInfo()
...
 		{
+			if (this.outValues == null)
+			{
+				return;
+			}
+
 			foreach (var item in this.outValues)
 			{
-				// it's already evaluated here
-				// TODO: refactor so that we
```
stakx: That comment makes no sense at all; removed.

Source/MethodCall.cs
```diff
@@ -253,67 +245,44 @@ public void EvaluatedSuccessfully()
...
-					var argsFuncType = mockEventArgsFunc.GetType();
-					if (argsFuncType.GetTypeInfo().IsGenericType && argsFuncType.GetGenericArguments().Length == 1)
-					{
-						this.Mock.DoRaise(this.mockEvent, (EventArgs)mockEventArgsFunc.InvokePreserveStack());
-					}
-					else
-					{
-						this.Mock.DoRaise(this.mockEvent, (EventArgs)mockEventArgsFunc.InvokePreserveStack(call.Arguments));
-					}
-				}
```
stakx: This whole event triggering logic is moved into the `RaiseEventResponse` nested class. at the end of the file.

Source/MethodCall.cs
```diff
@@ -340,20 +309,20 @@ public ICallbackResult Callback(Delegate callback)
 
 		protected virtual void SetCallbackWithoutArguments(Action callback)
 		{
-			this.setupCallback = delegate { callback(); };
+			this.callbackResponse = (object[] args) => callback();
```
stakx: When I first read that code, I was wondering why old-style `delegate` was used here. Make the lambda parameter explicit so it's clear exactly what is going on here, and also to make this method look more similar to `SetCallbackWithArguments` (below).

Source/MethodCall.cs
```diff
@@ -416,51 +385,67 @@ public IVerifies AtMost(int callCount)
...
-			this.mockEvent = ev.MemberInfo;
-			this.mockEventArgsFunc = func;
+			var ev = eventExpression.GetEvent((TMock)mock.Object).MemberInfo;
+			if (ev != null)
+			{
+				this.raiseEventResponse = new RaiseEventResponse(this.mock, ev, func, null);
+			}
+			else
+			{
+				this.raiseEventResponse = null;
```
stakx: This should possibly `throw` instead, like `Callback` does when given a `null` argument.

Source/MethodCall.cs
```diff
@@ -416,51 +385,67 @@ public IVerifies AtMost(int callCount)
...
-			this.mockEvent = ev.MemberInfo;
-			this.mockEventArgsParams = args;
+			var ev = eventExpression.GetEvent((TMock)mock.Object).MemberInfo;
+			if (ev != null)
+			{
+				this.raiseEventResponse = new RaiseEventResponse(this.mock, ev, null, args);
+			}
+			else
+			{
+				this.raiseEventResponse = null;
```
stakx: Same here, this should possibly `throw` instead, like `Callback` does when given a `null` argument.

Source/MethodCall.cs
```diff
@@ -416,51 +385,67 @@ public IVerifies AtMost(int callCount)
...
 			}
 
-			var lambda = SetupExpression.PartialMatcherAwareEval().ToLambda();
+			var lambda = this.originalExpression.PartialMatcherAwareEval().ToLambda();
 			var targetTypeName = lambda.Parameters[0].Type.Name;
 
 			message.Append(targetTypeName).Append(" ").Append(lambda.ToStringFixed());
 
-			if (TestMethod != null && FileName != null && FileLine != 0)
+#if !NETCORE
```
stakx: Now that the fields are removed when targeting .NET Core, we need this directive to completely omit this piece of code.

Source/MethodCall.cs
```diff
@@ -478,34 +463,57 @@ public string Format()
 
 			return builder.ToString();
 		}
-	}
-
-	internal class Condition
```
stakx: Moved into its own file.

Issue #249, opened 2016-02-16
Unable to mock protected generic method


mattzink:
It is not possible to currently mock a protected method that has generic type parameters.

`protected virtual TEntity GetEntityByName<TEntity>(string name)`

[Related question on StackOverflow](http://stackoverflow.com/questions/9416291/mocking-a-protected-generic-method-with-moq)

Prashantpola:
Is there any way to mock protected generic method of base class from which is calling from child.
I am calling base method from child class as below
this.ExecuteHttpPutCall<ClassT, ServiceResult<IEnumerable<string>>>(url, request,handler)

PR #495, opened 2017-10-22
Introduce `.Protected().As<TDuck>()` to perform duck-typed setup and verification of protected members

### LINKED ISSUE #223: Stub protected method with out/ref params
Howdy, I can't seem to work out a way to stub a protected method with an out param. Is it possible? Is/could there be something like `mock.Protected().Setup<Thing>("ImProtectedWithOut", ItExpr.IsAnyOut<int>())`?

stakx:
This proposes a new feature that aims to offer more complete support for setting up and verifying protected members (including the ability to set up protected generic methods, and protected methods with by-ref parameters).

It closes #223 and #249 if merged.

### Problem:

The current `Setup` and `Verify` methods accessible via `mock.Protected()` admittedly offer a handy way to get at protected members, but with the following downsides:

* They are by necessity completely "untyped". As such, they do not fit well into Moq's generally strongly-typed API.
* They are not easy to extend with support for generic methods, or methods with by-ref parameters.

### Proposed solution:

The new feature proposed here adds a new method, `mock.Protected().As<TDuck>()`, which offers `Setup` and `Verify` methods by which protected members can be accessed in a strongly-typed manner through a totally unrelated type `TDuck`. This type's members are expected to correspond to the mocked type's members (by having identical signatures). 

```csharp
abstract class Foo
{
    protected abstract void Protected(); 
}

interface Fooish
{
    void Protected();
}

var mock = new Mock<Foo>();
mock.Protected().As<Fooish>()
    .Setup(m => m.Protected())....; // `Foo.Protected` gets set up instead of `Fooish.Protected`
```

### When does this feature offer true added value?

This feature might seem pointless since it forces the user to declare a type just for testing; one could argue that it would be just as easy, and more meaningful, to simply let `Foo` implement `Fooish` and then mock `Fooish` directly. This is a fair assessment, but there might be cases where the type to be mocked is not owned by the user and cannot be modified. In these cases, the new feature offers a way to get at all protected members of that type.

### DIFF
--- a/Source/Protected/IProtectedAsMock.cs
+++ b/Source/Protected/IProtectedAsMock.cs
@@ -1,0 +1,76 @@
+﻿//Copyright (c) 2007. Clarius Consulting, Manas Technology Solutions, InSTEDD
+//https://github.com/moq/moq4
+//All rights reserved.
+
+//Redistribution and use in source and binary forms, 
+//with or without modification, are permitted provided 
+//that the following conditions are met:
+
+//    * Redistributions of source code must retain the 
+//    above copyright notice, this list of conditions and 
+//    the following disclaimer.
+
+//    * Redistributions in binary form must reproduce 
+//    the above copyright notice, this list of conditions 
+//    and the following disclaimer in the documentation 
+//    and/or other materials provided with the distribution.
+
+//    * Neither the name of Clarius Consulting, Manas Technology Solutions or InSTEDD nor the 
+//    names of its contributors may be used to endorse 
+//    or promote products derived from this software 
+//    without specific prior written permission.
+
+//THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+//CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+//INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+//MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+//DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR 
+//CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+//SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+//BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+//SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+//INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+//WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+//NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+//OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+//SUCH DAMAGE.
+
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+using System;
+using System.Linq.Expressions;
+
+using Moq.Language.Flow;
+
+namespace Moq.Protected
+{
+	/// <summary>
+	/// Allows setups to be specified for protected members (methods and properties)
+	/// seen through another type with corresponding members (that is, members
+	/// having identical signatures as the ones to be set up).
+	/// </summary>
+	/// <typeparam name="T">Type of the mocked object.</typeparam>
+	/// <typeparam name="TDuck">
+	/// Any type with members whose signatures are identical to the mock's protected members (except for their accessibility level).
+	/// </typeparam>
+	public interface IProtectedAsMock<T, TDuck> : IFluentInterface
+		where T : class
+		where TDuck : class
+    {
+		/// <summary>
+		/// Specifies a setup on the mocked type for a call to a <see langword="void"/> method.
+		/// </summary>
+		/// <param name="expression">Lambda expression that specifies the expected method invocation.</param>
+		/// <seealso cref="Mock{T}.Setup(Expression{Action{T}})"/>
+		ISetup<T> Setup(Expression<Action<TDuck>> expression);
+
+		/// <summary>
+		/// Specifies a setup on the mocked type for a call to a value-returning method.
+		/// </summary>
+		/// <typeparam name="TResult">Type of the return value. Typically omitted as it can be inferred from the expression.</typeparam>
+		/// <param name="expression">Lambda expression that specifies the expected method invocation.</param>
+		/// <seealso cref="Mock{T}.Setup{TResult}(Expression{Func{T, TResult}})"/>
+		ISetup<T, TResult> Setup<TResult>(Expression<Func<TDuck, TResult>> expression);
+	}
+}
--- a/Source/Protected/IProtectedAsMock.cs
+++ b/Source/Protected/IProtectedAsMock.cs
@@ -39,6 +39,7 @@
 // http://www.opensource.org/licenses/bsd-license.php]
 
 using System;
+using System.ComponentModel;
 using System.Linq.Expressions;
 
 using Moq.Language;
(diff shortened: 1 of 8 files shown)


Source/Protected/IProtectedMock.cs
```diff
@@ -53,6 +53,15 @@ namespace Moq.Protected
...
 		where TMock : class
 	{
+		/// <summary>
+		/// Set up protected members (methods and properties) seen through another type with identical member signatures.
+		/// </summary>
+		/// <typeparam name="TDuck">
+		/// Any type with members whose signatures are identical to the mock's protected members (except for their accessibility level).
+		/// </typeparam>
+		IProtectedAsMock<TMock, TDuck> As<TDuck>()
+			where TDuck : class;
```
stakx: Is it wise to overload the `As` verb with yet another meaning, or should this be named differently (perhaps `Like<TDuck>`)?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+
+		private static DuckReplacer DuckReplacerInstance = new DuckReplacer(typeof(TDuck), typeof(T));
+
+		public ProtectedAsMock(Mock<T> mock)
+		{
+			this.mock = mock;
+		}
+
+		public ISetup<T> Setup(Expression<Action<TDuck>> expression)
+		{
```
stakx: What kind of argument validation needs to be performed here?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			return Mock.Setup(this.mock, (Expression<Action<T>>)ReplaceDuck(expression), null);
+		}
+
+		public ISetup<T, TResult> Setup<TResult>(Expression<Func<TDuck, TResult>> expression)
+		{
```
stakx: What kind of argument validation needs to be performed here?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			return Mock.Setup(this.mock, (Expression<Func<T, TResult>>)ReplaceDuck(expression), null);
+		}
+
+		private static LambdaExpression ReplaceDuck(LambdaExpression expression)
+		{
+			var targetParameter = Expression.Parameter(typeof(T), expression.Parameters[0].Name);
```
stakx: Add a `Debug.Assert` verifying that the passed-in `LambdaExpression` has one parameter.

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			return Expression.Lambda(DuckReplacerInstance.Visit(expression.Body), targetParameter);
+		}
+
+		/// <summary>
+		/// <see cref="ExpressionVisitor"/> used to replace occurrences of `TDuck.Member` sub-expressions with `T.Member`.
+		/// </summary>
+		private sealed class DuckReplacer : ExpressionVisitor
```
stakx: Does nesting this non-generic type inside a generic type mean that the JIT / runtime will potentially generate it more than once? It would perhaps be safer to un-nest it.

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			{
+				this.duckType = duckType;
+				this.targetType = targetType;
+			}
+
+			protected override Expression VisitMethodCall(MethodCallExpression node)
+			{
+				if (node.Object is ParameterExpression left && left.Type == this.duckType)
+				{
+					var targetParameter = Expression.Parameter(typeof(T), left.Name);
```
stakx: Replace `typeof(T)` with `this.targetType`.

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+				{
+					return node;
+				}
+			}
+
+			protected override Expression VisitMember(MemberExpression node)
+			{
+				if (node.Expression is ParameterExpression left && left.Type == this.duckType)
+				{
+					var targetParameter = Expression.Parameter(typeof(T), left.Name);
```
stakx: Replace `typeof(T)` with `this.targetType`.

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+				{
+					return FindCorrespondingMethod(duckMethod);
+				}
+				else if (duckMember is PropertyInfo duckProperty)
+				{
+					return FindCorrespondingProperty(duckProperty);
+				}
+				else
+				{
+					throw new NotSupportedException();
```
stakx: Should this include a meaningful exception `Message` (defined in `Resoures.resx`)?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			private MethodInfo FindCorrespondingMethod(MethodInfo duckMethod)
+			{
+				var candidateTargetMethods =
+					this.targetType
+				    .GetMethods(BindingFlags.NonPublic | BindingFlags.Instance)
+				    .Where(ctm => IsCorrespondingMethod(duckMethod, ctm));
+
+				var targetMethod = candidateTargetMethods.Single();
```
stakx: Should there be a check whether exactly one target method was identified (and if not, throw an exception with a meaningful `Message`)?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			private PropertyInfo FindCorrespondingProperty(PropertyInfo duckProperty)
+			{
+				var candidateTargetProperties =
+					this.targetType
+					.GetProperties(BindingFlags.NonPublic | BindingFlags.Instance)
+					.Where(ctp => IsCorrespondingProperty(duckProperty, ctp));
+
+				return candidateTargetProperties.Single();
```
stakx: Should there be a check whether exactly one target property was identified (and if not, throw an exception with a meaningful `Message`)?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			private static bool IsCorrespondingMethod(MethodInfo duckMethod, MethodInfo candidateTargetMethod)
+			{
+				if (candidateTargetMethod.Name != duckMethod.Name)
+				{
+					return false;
+				}
+
+				if (candidateTargetMethod.IsGenericMethod != duckMethod.IsGenericMethod)
```
stakx: If `candidateTargetMethod` is generic, its generic parameters will be open. Which property is the right one to check: `.IsGenericMethod` or `.IsGenericMethodDefinition`?

Source/Protected/ProtectedAsMock.cs
```diff
@@ -0,0 +1,231 @@
...
+			{
+				var candidateTargetProperties =
+					this.targetType
+					.GetProperties(BindingFlags.NonPublic | BindingFlags.Instance)
+					.Where(ctp => IsCorrespondingProperty(duckProperty, ctp));
+
+				return candidateTargetProperties.Single();
+			}
+
+			private static bool IsCorrespondingMethod(MethodInfo duckMethod, MethodInfo candidateTargetMethod)
```
stakx: Check if we are comparing method signatures for a precise match anywhere else in Moq's code base, and either reuse existing code, or make this implementation here reusable. Also check if the framework already offers something to compare method signatures&mdash;no need to reinvent the wheel.

Source/Protected/IProtectedAsMock.cs
```diff
@@ -0,0 +1,76 @@
...
+		ISetup<T> Setup(Expression<Action<TDuck>> expression);
+
+		/// <summary>
+		/// Specifies a setup on the mocked type for a call to a value-returning method.
+		/// </summary>
+		/// <typeparam name="TResult">Type of the return value. Typically omitted as it can be inferred from the expression.</typeparam>
+		/// <param name="expression">Lambda expression that specifies the expected method invocation.</param>
+		/// <seealso cref="Mock{T}.Setup{TResult}(Expression{Func{T, TResult}})"/>
+		ISetup<T, TResult> Setup<TResult>(Expression<Func<TDuck, TResult>> expression);
+	}
```
stakx: Missing: `SetupGet`, `SetupSet`, `SetupSequence`, and all `Verify` methods.

Source/Protected/ProtectedAsMock.cs
stakx: Based on the answers to [the Stack Overflow question, "Should I avoid nested types in generic types?"](https://stackoverflow.com/questions/35956028), I'd say its fairly safe to keep this class nested, since the outer class' type parameters are all constrained to be reference types.

Source/Protected/ProtectedAsMock.cs
stakx: Done: `expression` must not be null, and it must not be about a member that doesn't exist in `T`.

PR #499, opened 2017-10-22
Let `SetupAllProperties` skip inaccessible methods

### LINKED ISSUE #498: `SetupAllProperties` tries to set up inaccessible properties
#455 made Moq more strict with regard to inaccessible methods: if an attempt is made to set up such a one, setup methods now throw an exception. What was forgotten back then is that `mock.SetupAllProperties()` will try to set even inaccessible methods (it doesn't perform the appropriate check), leading to exceptions in use cases such as #491 or the following:

```csharp
internal interface IFooInternals
{
	object C { get; }
	object D { get; }
}

public abstract class Foo : IFooInternals
{
	public virtual object C => "C";
	public object D => "D";
}

[Fact] // does not pass
public void SetupAllProperties_skips_inaccessible_properties_instead_of_throwing()
{
    var mock = new Mock<Foo>();

    var exception = Record.Exception(() =>
    {
        mock.SetupAllProperties();
    });

    Assert.Null(exception);
}
```

stakx:
This fixes #498.

### DIFF
--- a/Source/Mock.cs
+++ b/Source/Mock.cs
@@ -727,11 +727,12 @@ namespace Moq
 				//   interface as a getter-and-setter property.
 				.Where(p =>
 					   p.CanRead && p.CanOverrideGet() &&
-					   p.GetIndexParameters().Length == 0 &&
-					   p.CanWrite == p.CanOverrideSet())
-					   // ^ The last condition will be true for two kinds of properties:
+					   p.CanWrite == p.CanOverrideSet() &&
+					   // ^ This condition will be true for two kinds of properties:
 					   //    (a) those that are read-only; and
 					   //    (b) those that are writable and whose setter can be overridden.
+					   p.GetIndexParameters().Length == 0 &&
+					   ProxyFactory.IsMethodVisible(p.GetGetMethod(), out _))
 				.Distinct();
 
 			var setupPropertyMethod = mock.GetType().GetMethods()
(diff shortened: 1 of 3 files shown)


Source/Mock.cs
```diff
@@ -727,11 +727,12 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
 				//   interface as a getter-and-setter property.
 				.Where(p =>
 					   p.CanRead && p.CanOverrideGet() &&
-					   p.GetIndexParameters().Length == 0 &&
```
stakx: This condition gets moved from 2nd to 3rd position because it's more expensive.

Source/Mock.cs
```diff
@@ -727,11 +727,12 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
...
-					   p.CanWrite == p.CanOverrideSet())
-					   // ^ The last condition will be true for two kinds of properties:
+					   p.CanWrite == p.CanOverrideSet() &&
+					   // ^ This condition will be true for two kinds of properties:
 					   //    (a) those that are read-only; and
 					   //    (b) those that are writable and whose setter can be overridden.
+					   p.GetIndexParameters().Length == 0 &&
+					   ProxyFactory.IsMethodVisible(p.GetGetMethod(), out _))
```
stakx: This is the additional missing condition that fixes the bug.

PR #502, opened 2017-10-24
Allow `DefaultValue.Mock` to mock `Task<TMockable>` and `ValueTask<TMockable>`

### LINKED ISSUE #171: Support mocking the return value when the type is Task/Task<T>
When `Mock<T>.DefaultValue == DefaultValue.Empty`, the default return value of async method, which returns `Task` or `Task<T>`, will directly return null, then `await` on them will cause NullRefereceException.
In this case, the "Empty" value should be `Task.FromResult(default(T))`.

When `Mock<T>.DefaultValue == DefaultValue.Mock` the default return value of async method will change to the mocking `Task`/`Task<T>`, however, the `Task`/`Task<T>` are not able to be mocked, then the **System.ArgumentExceptionType to mock must be an interface or an abstract or non-sealed class.** will be thrown.
In this case, the "Mock" value should be `Task.FromResult(Mock.Of<T>())`.

stakx:
This PR enables the following scenario (shown for `Task<T>`, but same goes for `ValueTask<T>`):

```csharp
var mock = new Mock<IFoo>() { DefaultValue = DefaultValue.Mock };
var property = (await mock.Object.GetBarAsync()).Property;
             // ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             // With DefaultValue.Mock, Task<T> no longer returns tasks
             // that resolve to default(T), but to a mocked T, so these are
             // now awaitable straight away.

public interface IFoo
{
    Task<IBar> GetBarAsync();
}

public interface  IBar
{
    object Property { get; }
}
```

This closes #171.

### DIFF
--- a/Source/Mock.cs
+++ b/Source/Mock.cs
@@ -48,6 +48,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 
 using Moq.Diagnostics.Errors;
 using Moq.Language.Flow;
@@ -72,7 +73,7 @@ namespace Moq
 		{
 			this.eventHandlers = new EventHandlerCollection();
 			this.ImplementedInterfaces = new List<Type>();
-			this.InnerMocks = new ConcurrentDictionary<MethodInfo, Mock>();
+			this.InnerMocks = new ConcurrentDictionary<MethodInfo, MockWithWrappedMockObject>();
 			this.invocations = new InvocationCollection();
 			this.setups = new SetupCollection();
 			this.switches = Switches.Default;
@@ -200,7 +201,7 @@ namespace Moq
 			return value;
 		}
 
-		internal virtual ConcurrentDictionary<MethodInfo, Mock> InnerMocks { get; private set; }
+		internal virtual ConcurrentDictionary<MethodInfo, MockWithWrappedMockObject> InnerMocks { get; private set; }
 
 		internal virtual InvocationCollection Invocations => this.invocations;
 
@@ -281,7 +282,7 @@ namespace Moq
 
 			foreach (var inner in this.InnerMocks.Values)
 			{
-				if (!inner.TryVerify(out error))
+				if (!inner.Mock.TryVerify(out error))
 				{
 					return false;
 				}
@@ -311,7 +312,7 @@ namespace Moq
 
 			foreach (var inner in this.InnerMocks.Values)
 			{
-				if (!inner.TryVerifyAll(out error))
+				if (!inner.Mock.TryVerifyAll(out error))
 				{
 					return false;
 				}
@@ -1206,8 +1207,9 @@ namespace Moq
 			}
 
 			var result = (useAlternateProvider ?? this.DefaultValueProvider).GetDefaultReturnValue(method, this);
+			var unwrappedResult = TryUnwrapResultFromCompletedTaskRecursively(result);
 
-			if (result is IMocked mockedResult)
+			if (unwrappedResult is IMocked unwrappedMockedResult)
 			{
 				// TODO: Perhaps the following `InnerMocks` update isn't in quite the right place yet.
 				// There are two main places in Moq where `InnerMocks` are used: `Mock<T>.FluentMock` and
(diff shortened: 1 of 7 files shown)


Source/EmptyDefaultValueProvider.cs
```diff
@@ -92,27 +92,32 @@ private static object GetReferenceTypeDefault(Type valueType)
 			else if (valueType == typeof(Task))
 			{
 				// Task<T> inherits from Task, so just return Task<bool>
-				return GetCompletedTaskForType(typeof(bool));
+				return Task.FromResult(false);
```
stakx: (This optimization is not directly related to this PR.)

Source/EmptyDefaultValueProvider.cs
```diff
@@ -92,27 +92,32 @@ private static object GetReferenceTypeDefault(Type valueType)
...
 			}
-			else if (valueType.GetTypeInfo().IsGenericType && valueType.GetGenericTypeDefinition() == typeof(IEnumerable<>))
+			else if (valueType.GetTypeInfo().IsGenericType)
```
stakx: (This optimization is not directly related to this PR, but since one commit targets code de-duplication in `MockDefaultValueProvider`, it seems only logical to make the same effort in `EmptyDefaultValueProvider`, too.)

Source/MockDefaultValueProvider.cs
```diff
@@ -59,23 +60,40 @@ public MockDefaultValueProvider(Mock owner)
...
+			{
+				var returnTypeGenericDef = returnType.GetGenericTypeDefinition();
+				if (returnTypeGenericDef == typeof(Task<>))
+				{
+					taskResultType = returnType.GetGenericArguments()[0];
+				}
+			}
+
+			object value = null;
+			if (taskResultType == null)
```
stakx: This is still pretty ugly. What these lines of code basically amount to is, "If I have some special handling capabilities for some special type (`Task<T>`), don't let the fallback value provider generate a value for it; I'll deal with it myself right away." Find a way to express this more clearly in code.

Source/MockDefaultValueProvider.cs
```diff
@@ -59,23 +60,40 @@ public MockDefaultValueProvider(Mock owner)
...
 					newMock.DefaultValue = owner.DefaultValue;
 					newMock.CallBase = owner.CallBase;
-					return newMock;
+					var wrapper = taskResultType != null ? CreateCompletedTask(taskResultType, newMock.Object) : null;
+					return new MockWithOptionalWrapper(newMock, wrapper);
 				});
 			}
 
-			return mock != null ? mock.Object : value;
+			return mock.Mock != null ? mock.Wrapper ?? mock.Mock.Object : value;
```
stakx: Doesn't read very clearly. Probably the same as `mock.Wrapper ?? mock.Mock?.Object ?? value`, but it would be better to rewrite the whole method so that no convoluted expression is required at its end.

Source/MockWithOptionalWrapper.cs
```diff
@@ -0,0 +1,58 @@
...
+//NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+//OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+//SUCH DAMAGE.
+
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+namespace Moq
+{
+	internal struct MockWithOptionalWrapper
```
stakx: This type is useful because it helps us avoid wrapping and unwrapping values in a completed `Task<T>`, but its semantics are still somewhat confusing, esp. the fact that the `Wrapper` does not wrap `Mock`, but `Mock.Object`. Perhaps rename that property.

Source/Mock.cs
```diff
@@ -1206,8 +1207,9 @@ internal object GetDefaultValue(MethodInfo method, DefaultValueProvider useAlter
 			}
 
 			var result = (useAlternateProvider ?? this.DefaultValueProvider).GetDefaultReturnValue(method, this);
+			var unwrappedResult = TryUnwrapResultFromCompletedTaskRecursively(result);
```
stakx: For the standard default value providers (`Empty` and `Mock`), it would be great if the wrapping (inside the provider) and unwrapping (here) could be avoided. It should be possible to add a fast path as follows:

1. Check whether `this.DefaultValue != DefaultValue.Custom`. If false, stop.
2. Check whether `method` returns a task. If so, ask the provider not for the actual task return type via `GetDefaultReturnValue`, but for the type of the wrapped value via `GetDefaultValue`.
3. Perform the wrapping into completed tasks here, given the value returned from the provider.

This would save us one unnecessary recursive task unwrapping operation.

Issue #437, opened 2017-08-31
Can ValueTask IReturnsResult<TMock> extensions be added?


AdamDotNet:
The `ReturnsAsync()` and `ThrowsAsync()` are great, but sadly don't work with the newer type, `ValueTask<T>`. [I've cooked up a gist](https://gist.github.com/AdamDotNet/9394db2f89cada7c83bd6fc55a6c7695) that shows it pretty simple to add (most of the changes were `Task` to `ValueTask`), though I didn't peruse through your unit tests, the cost could be a little higher to implement. Also, Moq would have to depend on the [System.Threading.Tasks.Extensions NuGet](https://www.nuget.org/packages/System.Threading.Tasks.Extensions/). That seems like a small barrier, given it supports .NetStandard1.0.

The idea is to merge in those extensions with your existing [ReturnsExtension class](https://github.com/moq/moq4/blob/e2662721e13d52b081a2706608cb03e458019279/Source/ReturnsExtensions.cs).

I can live with adding these extensions to my own project or private NuGet, but it seems like a value add for everyone who uses the excellent Moq. Thank you!

stakx:
@AdamDotNet - sorry for the delay. This sounds like a reasonable feature addition. A couple of thoughts:

* The additional dependency on `System.Threading.Tasks.Extensions` is not a problem, if we can still target .NET 4.5 and .NET Standard 1.3.

* The new extension methods for `ValueTask<T>` should not cause method resolution conflicts in existing code.

* There should be a few tests (in `ReturnsExtensionsFixture.cs`) targeting the new extension methods.

Would you like to submit a PR? (If so, please target the `develop` branch.)

AdamDotNet:
@stakx - I'd like to be able to help. However, I can't quite get in a compiling state before I get to work on it.

`GenericTypeParameters.tt` Initially couldn't run, with the message `'Enumerable' does not exist in the current context.`

I tested adding
```
<#@ Assembly Name="System.Core.dll" #>
<#@ import namespace="System.Linq" #>
```
The template now runs, but the resulting `GenericTypeParameters.cs` file is empty. If this file is not critical for what I'll be working on, I guess I'll proceed, and be sure to commit only the relevant file changes. Do you have any additional advice for working in the `Moq.sln` solution? Thanks!

stakx:
[...]

PR #506, opened 2017-10-24
Add ValueTask extensions

AdamDotNet:
Addresses #437  
- Reference NuGet System.Threading.Tasks.Extensions
 - Add extension methods to ReturnsExtensions.cs/.tt
 - Add unit tests to ReturnsExtensionsFixture.cs

Tests pass on NetCoreApp2.0 and Net46 TFMs.

### DIFF
--- a/Source/Moq.csproj
+++ b/Source/Moq.csproj
@@ -28,7 +28,7 @@
 
 	<ItemGroup>
 		<PackageReference Include="Castle.Core" Version="4.2.1" />
-    <PackageReference Include="GitInfo" Version="1.1.14" />
+		<PackageReference Include="GitInfo" Version="1.1.14" />
 		<PackageReference Include="IFluentInterface" Version="2.0.0" />
 		<PackageReference Include="SourceLink.Create.CommandLine" Version="2.2.1" PrivateAssets="All" />
 		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.4.0" />
--- a/Source/Moq.csproj
+++ b/Source/Moq.csproj
@@ -28,9 +28,10 @@
 
 	<ItemGroup>
 		<PackageReference Include="Castle.Core" Version="4.2.1" />
-		<PackageReference Include="GitInfo" Version="1.1.14" />
+    <PackageReference Include="GitInfo" Version="1.1.14" />
 		<PackageReference Include="IFluentInterface" Version="2.0.0" />
 		<PackageReference Include="SourceLink.Create.CommandLine" Version="2.2.1" PrivateAssets="All" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.4.0" />
 		<DotNetCliToolReference Include="dotnet-sourcelink" Version="2.2.1" />
 	</ItemGroup>
 	<ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
--- a/Source/ReturnsExtensions.cs
+++ b/Source/ReturnsExtensions.cs
@@ -24,6 +24,18 @@ namespace Moq
 		}
 
 		/// <summary>
+		/// Specifies the value to return from an asynchronous method.
+		/// </summary>
+		/// <typeparam name="TMock">Mocked type.</typeparam>
+		/// <typeparam name="TResult">Type of the return value.</typeparam>
+		/// <param name="mock">Returns verb which represents the mocked type and the task of return type</param>
+		/// <param name="value">The value to return, or <see longword="null"/>.</param>
+		public static IReturnsResult<TMock> ReturnsAsync<TMock, TResult>(this IReturns<TMock, ValueTask<TResult>> mock, TResult value) where TMock : class
+		{
+			return mock.ReturnsAsync(() => value);
+		}
+
+		/// <summary>
 		/// Specifies a function that will calculate the value to return from the asynchronous method.
 		/// </summary>
 		/// <typeparam name="TMock">Mocked type.</typeparam>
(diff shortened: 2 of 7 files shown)


Source/Moq.csproj
```diff
@@ -28,9 +28,10 @@
 
 	<ItemGroup>
 		<PackageReference Include="Castle.Core" Version="4.2.1" />
-		<PackageReference Include="GitInfo" Version="1.1.14" />
+    <PackageReference Include="GitInfo" Version="1.1.14" />
```
stakx: Please fix indentation.

Source/Moq.csproj
```diff
@@ -28,9 +28,10 @@
...
 		<PackageReference Include="IFluentInterface" Version="2.0.0" />
 		<PackageReference Include="SourceLink.Create.CommandLine" Version="2.2.1" PrivateAssets="All" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.4.0" />
```
stakx: Shouldn't the same package reference be added for the `net46` target in both test projects? I suppose the test projects right now compile only because they reference the Moq project, which in turns references `System.Threading.Tasks.Extensions.dll`. I think it would be more correct if the test projects referenced their required NuGet package directly.

(However, the same reference isn't needed for `netcoreapp2.0`, as the SDK already includes it.)

Source/ReturnsExtensions.tt
```diff
@@ -58,5 +58,50 @@ namespace Moq
 <#
 		}
 		#>
+
```
stakx: Could I ask you for a favour? While we're editing this file, could you please change the class definition in this file (line 15) from:

> ```csharp
> public static class GeneratedReturnsExtensions
> ```

to:

```csharp
static partial class ReturnsExtensions
```

and then, in `ReturnsExtensions.cs`:
* also declare `ReturnsExtensions` `partial`; and
* put an `[EditorBrowsable(EditorBrowsableState.Never)]` attribute on the class?

Source/ReturnsExtensions.cs
```diff
@@ -24,6 +24,18 @@ public static class ReturnsExtensions
 		}
 
 		/// <summary>
+		/// Specifies the value to return from an asynchronous method.
+		/// </summary>
+		/// <typeparam name="TMock">Mocked type.</typeparam>
+		/// <typeparam name="TResult">Type of the return value.</typeparam>
+		/// <param name="mock">Returns verb which represents the mocked type and the task of return type</param>
```
stakx: I realise that you copy-pasted these documentation comments from the existing `Task<T>` methods, so this is more of a comment than asking for a change:

1. This parameter's name should have been `setup`, not `mock`: The `.Returns` "verb" acts on a setup of some mock, not on the mock itself. Unfortunately, renaming this parameter everywhere is theoretically a breaking change (albeit one unlikely to affect anyone, since we're talking of the `this` parameter). Let's leave it as is, better to be consistent everywhere.

2. I wonder if anyone readily understands that complicated summary description. :-) I would personally prefer something simpler, e.g. "The setup whose return value is configured." Feel free to change the description to something sensible *everywhere*, if you're in the mood & can spare the time.

Source/Moq.csproj
AdamDotNet: The reason why it "just works" without the explicit reference is because of the new style project file format automatically resolves transitive dependencies. In VS, expand the dependencies node -> .NetFramework 4.6 -> Projects -> Moq and you'll see System.Threading.Tasks.Extensions is indeed referenced by the unit test projects.

I can go ahead and add those references to the csproj and vbproj unit test files since you asked though.

stakx: Hmmmm, OK, if it's officially documented to work that way perhaps it just takes some getting used to on my part. :-) Thanks for teaching me something new. Both is fine with me then, it's your call!

AdamDotNet: No problem! I think I'll prefer the transitive dependency resolution then. Seems to be the more modern approach. See also [Dotnet Project System](/dotnet/project-system/issues/199); it's a feature pulled from the project.json/xproj days of the early .Net Core SDK and CLI tools.

Source/ReturnsExtensions.cs
AdamDotNet: On item 2., you're still referring to `<param name="mock">` right? I read the param description several times and I think I finally get it haha. But since it is an extension method and most of the time extension methods are used with extension method syntax instead of as static method calls where you pass in the `ISetup` so most people using these methods won't actually see the comment in the IDE. All the same though, your suggestion slightly tweaked "The setup whose return value is _to be_ configured." seems to make sense to me. I am just a little anxious about not finding _every_ instance as you asked (assuming these kind of comments exist in places other than just the Task/ValueTask extensions).

stakx:
All good, thanks!

Perhaps you noticed that I deleted my review comment about merging `GeneratedReturnsExtensions` into `ReturnsExtensions`. I really prefer it the way you've changed it, yet I realised too late this rename will cause a binary-level breaking change. If someone updates the Moq DLL without recompiling their code, they'll see problems.

Not sure whether this is acceptable to end users, so I might end up reverting this particular change before releasing v4.8. :-/ Any thoughts on that?

Source/ReturnsExtensions.cs
stakx: With *everywhere* I actually meant only "everywhere in that particular class". Sorry if that wasn't clear enough. Since it's really a detail that doesn't make a huge difference, like you correctly say, I went ahead &  merged your PR without this. :)

AdamDotNet:
I'm still a little novice at GitHub, so I guess I stuck to the comment in the email instead of noticing it was gone here on the website. Sorry about that!

I think you're right. I just refreshed my knowledge on SemVer and if that's what you follow for moq, then you can only make a breaking change like that with a major revision. I think there are a lot of cases where the code is recompiled when pulling new nuggets, but MissingMethodException is a thing if they just slip in the dll, or even when assembly binding redirects are in place.

If you can squirrel away that portion of the commit for a possible future v5, and maintain binary compatibility on v4, that's probably the safest bet.

Again, sorry I didn't put a little more thought into it in the first place!

stakx:
> If you can squirrel away that portion of the commit for a possible future v5, and maintain binary compatibility on v4, that's probably the safest bet.

Agreed. Even though this repo doesn't follow semver (the major version is basically fixed at 4 due to the repo's name, and the patch version gets computed automatically by [GitInfo](https://github.com/kzu/GitInfo), so the only thing we have any real control over is the minor version)&mdash;nevertheless I'm trying to stay as close to semver as possible. I'll probably revert that class name change.

> Again, sorry I didn't put a little more thought into it in the first place!

My bad: I shouldn't have suggested that particular change to begin with.

Thank you for contributing!

PR #509, opened 2017-10-29
Remove `Interceptor.calls`, `ExpressionKey`, and `SetupKind`

stakx:
All of these were used so far to detect when one setup overrides an earlier, equivalent setup. This PR proposes their removal.

1. Instead of performing the checks against `Interceptor.calls`, they can run directly against `InterceptorContext.OrderedCalls` (of which the former is a partial view).

2. Instead of comparing setups via their string representation (built for `ExpressionKey`), they can be compared by directly comparing their expression trees using the already existing `ExpressionComparer`.

3. `SetupKind` becomes redundant after (2).


Source/ExpressionComparer.cs
```diff
@@ -40,14 +40,12 @@
...
 using System;
 using System.Collections.Generic;
-using System.Linq.Expressions;
 using System.Collections.ObjectModel;
-using System.Reflection;
-using System.Linq;
+using System.Linq.Expressions;
 
-namespace Moq.Matchers
+namespace Moq
```
stakx: We no longer use `ExpressionComparer` just together with matchers, so it's moved into the main namespace.

Source/IInterceptStrategy.cs
```diff
@@ -142,7 +138,7 @@ internal IEnumerable<IProxyCall> OrderedCalls
 			{
 				lock (orderedCalls)
 				{
-					return orderedCalls.ToList();
+					return orderedCalls.ToArray();
```
stakx: Micro-optimization. Since callers are only going to read the returned collection, choose a type with as little overhead as possible.

Source/Interceptor.cs
```diff
@@ -75,56 +74,50 @@ internal void VerifyAll()
...
+			var verifiedSetupsPerMethod = new Dictionary<MethodInfo, List<Expression>>();
+
+			foreach (var setup in this.InterceptionContext.OrderedCalls)
 			{
-				var failures = calls.Values.Where(match);
-				if (failures.Any())
+				if (setup.IsConditional)
 				{
-					throw new MockVerificationException(failures.ToArray());
+					continue;
```
stakx: Conditional setups were already excluded from `Mock.Verify[All]` prior to this PR, but that wasn't as easy to see as it is now (because previously, conditional setups only got added to `OrderedCalls`, but not to `calls`, and mass verification happened against the latter).

Source/Interceptor.cs
```diff
@@ -75,56 +74,50 @@ internal void VerifyAll()
...
-			if (kind == SetupKind.PropertySet)
-			{
-				keyText = "set::" + keyText;
-			}
+				List<Expression> verifiedSetupsForMethod;
+				if (!verifiedSetupsPerMethod.TryGetValue(setup.Method, out verifiedSetupsForMethod))
+				{
+					verifiedSetupsForMethod = new List<Expression>();
+					verifiedSetupsPerMethod.Add(setup.Method, verifiedSetupsForMethod);
+				}
```
stakx: There's another occurrence of this code pattern somewhere else in Moq's codebase, perhaps consider creating some kind of `MultiMap` / `MultiDictionary` ADT so this can be abstracted away.

Source/InterceptorStrategies.cs
```diff
@@ -69,22 +69,26 @@ public InterceptionAction HandleIntercept(ICallContext invocation, InterceptorCo
 				return InterceptionAction.Continue;
 			}
 
-			IProxyCall lastMatchingSetup = null;
-			IProxyCall lastMatchingSetupForSameMethod = null;
+			IProxyCall matchedSetup = null;
 			foreach (var setup in ctx.OrderedCalls)
 			{
```
stakx: Since `InterceptorContext.OrderedCalls`, now being a `Stack<IProxyCall>` instead of a `List<IProxyCall>`, is now in reverse order (last setup added gets enumerated first), we no longer search for the last matching setup, but for the first. This makes it possible to break out of the loop earlier.

Source/Mock.cs
```diff
@@ -406,6 +406,7 @@ private static bool AreSameMethod(Expression left, Expression right)
 		private static string FormatSetupsInfo(IEnumerable<IProxyCall> setups)
 		{
 			var expressionSetups = setups
+				.Reverse()
```
stakx: This is needed for continuity (so setups are shown in exception messages in the order in which they were registered). Minor performance penalty here, but since it only happens in the error case, that's OK.

Source/MockException.cs
```diff
@@ -176,23 +178,23 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
 #endif
 	internal class MockVerificationException : MockException
 	{
-		IProxyCall[] failedSetups;
+		IEnumerable<IProxyCall> failedSetups;
 
-		public MockVerificationException(IProxyCall[] failedSetups)
+		public MockVerificationException(IEnumerable<IProxyCall> failedSetups)
```
stakx: This change is to avoid an unnecessary call to `.ToArray()` in the calling code.

PR #511, opened 2017-11-02
Remove leaked `MockVerificationException` and catch-throw-recatch cycle during verification

### LINKED ISSUE #507: `mock.Verify` and `mock.VerifyAll` leak internal `MockVerificationException` type
That exception type is delared `internal` and therefore cannot be specifically caught with a dedicated `catch` clause. But it can be seen in the exception message when it occurs.

This type was apparently introduced as an internal mechanism to aggregate mock exceptions during a broad `MockRepository.Verify` or `VerifyAll` call. Only `MockException` should be thrown directly.

stakx:
Closes #507.

### DIFF
--- a/Source/Obsolete/MockFactory.cs
+++ b/Source/Obsolete/MockFactory.cs
@@ -321,7 +321,7 @@ namespace Moq
 		/// <summary>
 		/// Invokes <paramref name="verifyAction"/> for each mock 
 		/// in <see cref="Mocks"/>, and accumulates the resulting 
-		/// <see cref="MockVerificationException"/> that might be 
+		/// verification exceptions that might be
 		/// thrown from the action.
 		/// </summary>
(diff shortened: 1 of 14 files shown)


Source/Obsolete/MockFactory.cs
```diff
@@ -330,25 +329,23 @@ protected virtual void VerifyMocks(Action<Mock> verifyAction)
...
+			var failures = new List<UnmatchedSetups>();
 
 			foreach (var mock in mocks)
 			{
 				try
 				{
 					verifyAction(mock);
 				}
-				catch (MockVerificationException mve)
+				catch (MockException ex) when (ex.Error is UnmatchedSetups failure)
```
stakx: Unfortunately the throw-catch pattern cannot be replaced with a full try pattern due to this method's signature and the fact that it's `protected`. It would have to be marked as `[Obsolete]`.

PR #514, opened 2017-11-05
Remove redundant code from matcher infrastructure

stakx:
The matcher infrastructure is currently somewhat difficult to understand:

### Current situation:

* There is `Match` and `Match<T>`, which is basically a wrapper around a `Predicate<T>`. (Unfortunately, this delegate type was chosen over `Func<T, bool>`, but alas.)
* Then there is `IMatcher`, having several implementations in the `Moq.Matchers` namespace.
* One of those is `Matcher`, which wraps a `Match<T>` (which wraps a `Predicate<T>`).

Regarding extensibility, the situation is similarly complex:

* There is a `[Matcher]` custom attribute, which appears to be the extension point for the infrastructure that was built first.
* Then there is the static `Match.Create` method, which supposedly deprecates `[Matcher]`.
* Then there is a `[AdvancedMatcher]` attribute, which is perhaps the newest addition and related to `IMatcher`, but hardly used at all.

### Changes:

This PR removes some of these things that aren't actually used, or don't make any difference whatsoever.

* `Matcher` is removed, as `Match` can already do the same things.
* `[Matcher]` is marked as obsolete.
* `[AdvancedMatcher]` is removed, as it has no observable effect.


Source/ExpressionExtensions.cs
```diff
@@ -202,16 +202,11 @@ public static Expression PartialMatcherAwareEval(this Expression expression)
 
 		private static bool ReturnsMatch(MethodCallExpression expression)
 		{
-			if (!expression.Method.IsDefined(typeof(AdvancedMatcherAttribute), true))
```
stakx: One would expect there to be a check for the obsolete `[Matcher]` attribute, but such a check wasn't there previously. It might be good to investigate more thoroughly exactly when `ReturnsMatch` gets used.

Source/MatcherFactory.cs
```diff
@@ -61,11 +66,6 @@ public static IMatcher CreateMatcher(Expression expression, bool isParams)
 			// the values are ints, but if the method to call 
 			// expects, say, a double, a Convert node will be on 
 			// the expression.
-			if (isParams && (expression.NodeType == ExpressionType.NewArrayInit || !expression.Type.IsArray))
```
stakx: This gets moved up a few lines because the preceding comment belongs with code further below; but shouldn't the type conversion code be the first thing in the method instead?

Source/MatcherFactory.cs
```diff
@@ -94,28 +94,15 @@ public static IMatcher CreateMatcher(Expression expression, bool isParams)
...
-					var matcher = attr.CreateMatcher();
-					matcher.Initialize(originalExpression);
-					return matcher;
-				}
-				else if (staticMatcherMethodAttr != null)
-				{
-					var matcher = new MatcherAttributeMatcher();
-					matcher.Initialize(originalExpression);
-					return matcher;
+					return new MatcherAttributeMatcher(call);
```
stakx: Passing `call` instead of `originalExpression` is actually a small change, as the latter has had any type conversion node removed and the former hasn't. But it would seem more correct since code inside `MatcherAttributeMatcher` silently assumes that it's given a `MethodCallExpression` (and not a conversion node).

Source/Matchers/PredicateMatcher.cs
```diff
@@ -1,75 +0,0 @@
...
-//[This is the BSD license, see
-// http://www.opensource.org/licenses/bsd-license.php]
-
-using System;
-using System.Linq.Expressions;
-using System.Reflection;
-
-namespace Moq.Matchers
-{
-	internal class PredicateMatcher : IMatcher
```
stakx: This was used together with `[AdvancedMatcher]` on `ItExpr.IsAny`. Surprisingly, it doesn't appear to make any difference whatsoever whether that attribute is there or not. Perhaps check again whether diagnostic output (rendered expression in error messages) is affected, as that might be something not covered by any test.

Source/Obsolete/MatcherAttribute.cs
```diff
@@ -157,6 +157,8 @@ namespace Moq
 	/// </code>
 	/// </example>
 	[AttributeUsage(AttributeTargets.Method, Inherited = true)]
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	[Obsolete("This feature has been deprecated in favor of `Match.Create`.")]
```
stakx: An existing remark in the documentation comment already states that this type has been deprecated, this just formalises it.

Moq.Tests/ExtensibilityFixture.cs
```diff
@@ -107,26 +107,9 @@ public interface IOrderRepository
...
-
-			public bool Matches(object value)
-			{
-				if (value is Order &&
-					((Order)value).Amount >= 1000)
-					return true;
-
-				return false;
-			}
+			return Match.Create<Order>(o => o.Amount >= 1000);
```
stakx: This is what was originally used before the introduction of `[AdvancedMatcher]`.

Source/MatcherFactory.cs
```diff
@@ -95,22 +95,21 @@ public static IMatcher CreateMatcher(Expression expression, bool isParams)
...
 
 				if (attr != null)
 				{
 					var matcher = attr.CreateMatcher();
 					matcher.Initialize(originalExpression);
 					return matcher;
 				}
-				else if (staticMatcherMethodAttr != null)
+#pragma warning disable 618
+				else if (call.Method.IsDefined(typeof(MatcherAttribute), true))
```
stakx: Minor optimization here: We don't need the actual attribute (`GetCustomAttribute`; above), we just need to know whether it is there or not (`IsDefined`). This call is about twice as fast.

PR #515, opened 2017-11-06
Turn collection of source file information per setup into an opt-in feature

stakx:
Up until now, Moq has been collecting source file information for each setup created. This came at a slight performance penalty. There is currently a regression in the .NET Framework which means this penalty becomes gigantic (more expensive than proxy creation!).

This commit turns source file information collection into an opt-in feature. Towards that end, a `Switches` property is added to `Mock` and `MockRepository` which allows users to opt in or out of certain features. Source file info collection is controlled via the switch `Switches.CollectDiagnosticFileInfoForSetups`.


stakx:
Merging this directly, not via PR.

PR #520, opened 2017-11-10
Validate delegates passed to `Returns` more strictly

### LINKED ISSUE #445: Detect method signature mismatch in `Setup` / `Returns`

stakx:
This PR adds minimal validation logic for delegates passed to `Returns`:

* Return type must be assignment compatible
* Parameter count must match

While it would be desirable from a technical standpoint, parameter types do not get checked for assignment compatibility due to two possibly widespread practices:

* Callback uses a by-value callback parameter while method being set up has a by-ref-parameter:

   ```csharp
   TArg expectedX = ...;
   mock.Setup(m => m.Method(ref expectedX).Returns<TArg>(actualX => ...);
   ```

* Callback uses a subtype and setup uses a matcher to ensure runtime value conversion will always succeed:

   ```csharp
   mock.Setup(m => m.MethodWithObjectParam(It.IsAny<int>()).Returns<int>(i => ...);
   mock.Setup(m => m.MethodWithObjectParam(It.IsAny<string>()).Returns<string>(str => ...);
   ```

While those are possibly not the cleanest programming practices (from a purely technical point of view), many people might find them useful in practice. Allowing them means that parameter type validation becomes pretty much pointless.

This closes #445.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 #### Changed
 
 * **Breaking change:** `SetupSequence` now overrides pre-existing setups like all other `Setup` methods do. This means that exhausted sequences no longer fall back to previous setups to produce a "default" action or return value. (@stakx, #476)
+* Delegates passed to `Returns` are validated a little more strictly than before (return type and parameter count must match with method being set up) (@stakx, #520)
 
 #### Fixed
 
--- a/Moq.Tests/ReturnsValidationFixture.cs
+++ b/Moq.Tests/ReturnsValidationFixture.cs
@@ -1,0 +1,123 @@
+﻿using System;
+using System.Reflection;
+
+using Moq.Language.Flow;
+
+using Xunit;
+
+namespace Moq.Tests
+{
+	public class ReturnsValidationFixture
+	{
+		private Mock<IType> mock;
+		private ISetup<IType, IType> setup;
+
+		public ReturnsValidationFixture()
+		{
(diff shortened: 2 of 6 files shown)


ohadschn:
@stakx Not sure what you mean "Allowing them means that parameter type validation becomes pretty much pointless". If I understand correctly, you would be preventing stuff like `mock.Setup(m => m.MethodWithStringParam(It.IsAny<int>()).Returns<int>(i => ...);` which is a win in my book...

stakx:
Having been more or less the only active contributor to Moq v4 for a while now, I've become aware that this also means I'm increasingly exposed as the one to be blamed when things break. For this reason I'm simply not as keen now as I might've previously when it comes to introducing breaking changes; especially when they're avoidable.

That being said, I still do agree it would be nice to be more consistent and thorough with validation.

PR #523, opened 2017-11-19
Simplify and optimize interception

stakx:
This PR achieves two things:

1. Makes the whole interception mechanism a little simpler and easier to understand. `InterceptorContext` is replaced by dedicated collection properties directly on `Mock`: `Mock.Setups`, `Mock.Invocations`, and `Mock.EventHandlers`. These use dedicated collection types who perform the necessary locking (for thread-safety) internally.

2. Applies several optimizations (such as fewer allocations and collection copying) that speed up mock method interception by approx. 30% (though interception is already pretty fast, so this might not be very noticeable).

(There's possibly room for further speed improvements, esp. regarding the `lock`s inside the collection types `SetupCollection`, `InvocationCollection`, `EventHandlerCollection`.)


Source/Interceptor.cs
```diff
@@ -131,10 +131,8 @@ public void AddCall(IProxyCall call)
 			new Lazy<IInterceptStrategy[]>(
 				() => new IInterceptStrategy[]
 				{
-					HandleFinalizer.Instance,
 					HandleTracking.Instance,
-					InterceptMockPropertyMixin.Instance,
-					InterceptObjectMethodsMixin.Instance,
+					HandleWellKnownMethods.Instance,
```
stakx: Handling finalizers after `HandleTracking` *might* in theory be problematic since the objects related to the one being finalized might already be gone. But then, assuming that the underlying proxy library might intercept a finalizer is just as problematic. (DynamicProxy doesn't intercept finalizers, so all logic in Moq that deals with a finalizer call is really just a backup.)

Source/Mock.cs
```diff
@@ -261,8 +265,10 @@ public void Verify()
 
 		private bool TryVerify(out UnmatchedSetups error)
 		{
-			if (!this.Interceptor.TryVerify(out error))
+			var uninvokedVerifiableSetups = this.Setups.ToArrayLive(setup => setup.IsVerifiable && !setup.Invoked);
```
stakx: By moving setups from `InterceptorContext` into `Mock`, verification is no longer spread across three types (`Mock`, `Interceptor`, `InterceptorContext`) but happens completely within `Mock`.

Source/AsInterface.cs
```diff
@@ -107,6 +101,8 @@ public override Switches Switches
 			set => this.owner.Switches = value;
 		}
 
+		internal override Type TargetType => this.owner.TargetType;
```
stakx: This is *almost* identical to `MockedType`, except for the overrides in `AsInterface`.

PR #524, opened 2017-11-19
Rewrite `EmptyDefaultValueProvider` to use dictionary lookup for special type handling

stakx:
This is a lot more efficient (I observed speed-ups of simple mock method invocations by 67%) and makes it possible to easily add handling of additional types that are currently ignored (such as `List<T>`, `IReadOnlyList<T>`, etc.).

### DIFF
--- a/Source/EmptyDefaultValueProvider.cs
+++ b/Source/EmptyDefaultValueProvider.cs
@@ -56,6 +56,17 @@ namespace Moq
 	{
 		private Dictionary<Type, object> defaultValues = new Dictionary<Type, object>();
 
+		private static Dictionary<Type, Func<Type, object>> factories = new Dictionary<Type, Func<Type, object>>()
+		{
+			[typeof(Array)] = CreateArray,
+			[typeof(IEnumerable)] = CreateEnumerable,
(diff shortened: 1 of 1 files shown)


Source/EmptyDefaultValueProvider.cs
```diff
@@ -73,73 +84,79 @@ public virtual object ProvideDefault(MethodInfo member)
...
-		private static Task GetCompletedTaskForType(Type type)
+		private static object CreateTask(Type type)
 		{
-			var tcs = Activator.CreateInstance(typeof (TaskCompletionSource<>).MakeGenericType(type));
+			return Task.FromResult(false);
+		}
 
-			var setResultMethod = tcs.GetType().GetMethod("SetResult");
-			var taskProperty = tcs.GetType().GetProperty("Task");
+		private static object CreateTaskOf(Type type)
```
stakx: Interestingly enough, while creating a completed task via `Task.FromResult<T>` is about 16x as fast as via `TaskCompletionSource<T>`, the same is not true if done via reflection: `Task.FromResult<T>` then becomes a little slower even though less reflection calls are involved. So we don't change over to a `Task.FromResult`-based solution.

Issue #173, opened 2015-05-21
Loose mock should also return empty for IReadOnlyList<T>


hickford:
I think the behaviour described at https://github.com/Moq/moq4/wiki/Quickstart#user-content-customizing-mock-behavior and http://www.nudoq.org/#!/Packages/Moq/Moq/MockBehavior/F/Loose

> default behavior is "Loose" mock, which never throws and **returns default values or empty arrays, enumerables**

Should also apply to the new .NET 4.5 interfaces `IReadOnlyList<T>` and `IReadOnlyCollection<T>`

Relevant code: https://github.com/Moq/moq4/blob/92e8ed91820b572a524e7c746d030d673e2295eb/Source/EmptyDefaultValueProvider.cs#L78

kzu:
PR link? ;)

hickford:
Oh I'm not sure how to do it without breaking the library for .NET < 4.5

stakx:
Closed #330 as a duplicate of this issue, but there's a relevant discussion there about ways to make Moq more consistent. Simply changing the (long-standing) current behavior of the various default value settings could be dangerous because it could break a lot of existing user code. It might be better to extend the API so it allows custom, configurable default value providers.

stakx:
This present issue requests that `DefaultValue.Empty` should produce an empty collection for the `IReadOnlyList<T>` type. While it would now be easily possible to fulfill that request, taking it too literally would result in an incomplete change because there are so many more collection types in the `System.Collections.*` namespaces that `EmptyDefaultValueProvider` does not currently support.

Neither does it make much sense to just add *all* collection types because many of them are rarely used (e. g. `ArrayList`, `BitArray`, `Hashtable`, `Queue`, etc.). All that work to support those will likely not pay off.

What's worse, there are actually some old Moq unit tests that ensure `EmptyDefaultValueProvider` does *not* handle certain collection types specially (`List<T>`!). I don't know *why* those tests were put in place, but there they are.

Finally, just changing the behaviour or `EmptyDefaultValueProvider` could potentially break a lot of existing user code.

We could introduce a backward compatibility for that, but then there's the question whether it is on or off by default.

[...]

PR #536, opened 2017-11-28
New `LookupOrFallbackDefaultValueProvider` base class for custom default value providers

stakx:
This adds a new base class `LookupOrFallbackDefaultValueProvider`, which builds on top of `DefaultValueProvider`, to Moq's public API.

* This type becomes the new base class for both `EmptyDefaultValueProvider` and `MockDefaultValueProvider`.

* It has inbuilt support for task types `Task`, `Task<TResult>`, and `ValueTask<TResult>`.

* It also provides enough customizability for folks to easily create their own default value providers that work like the empty default value provider. Here's an example of a provider that will produce empty collections for `IReadOnlyList<T>`:

   ```csharp
   sealed class MyEmptyDefaultValueProvider : LookupOrFallbackDefaultValueProvider
   {
       public MyEmptyDefaultValueProvider()
       {
           base.Register(typeof(IReadOnlyList<>), CreateReadOnlyListOf);
       }

       private static CreateReadOnlyListOf(Type type, Mock _)
       {
           var listType = typeof(List<>).MakeGenericType(type.GetGenericArguments());
           return Activator.CreateInstance(listType);
       }
   }
   ```

For the rationale behind this new type, and the reasoning why it doesn't come preconfigured as an empty default value provider supporting all standard collection types from the .NET BCL, see #173.

Perhaps we can add a customizable or expanded empty default value provider at a later time if there's still widespread demand / need for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * Improved support for setting up and verifying protected members (including generic methods and methods having by-ref parameters) via the new duck-typing `mock.Protected().As<TDuck>()` interface (@stakx, #495, #501)
 * Support for `ValueTask<TResult>` when using the `ReturnsAsync` extension methods, similar to `Task<TResult>` (@AdamDotNet, #506)
 * Special handling for `ValueTask<TResult>` with `DefaultValue.Empty` (@stakx, #529)
-* Support for custom default value generators (besides `DefaultValue.Empty` and `DefaultValue.Mock`): subclass the new `DefaultValueProvider` and set up via `Mock[Repository].DefaultValueProvider` (@stakx, #533)
+* Support for custom default value generation strategies besides `DefaultValue.Empty` and `DefaultValue.Mock`:
+  Implement custom providers by subclassing either `DefaultValueProvider` or `LookupOrFallbackDefaultValueProvider`,
+  install them by setting `Mock[Repository].DefaultValueProvider` (@stakx, #533, #536)
 * Allow `DefaultValue.Mock` to mock `Task<TMockable>` and `ValueTask<TMockable>` (@stakx, #502)
 
 #### Changed
--- a/Moq.Tests/LookupOrFallbackDefaultValueProviderFixture.cs
+++ b/Moq.Tests/LookupOrFallbackDefaultValueProviderFixture.cs
@@ -1,0 +1,198 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Moq.Tests
+{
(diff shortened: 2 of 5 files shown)


Serg046:
But why is it sealed?
protected internal **sealed** override object GetDefaultValue(Type type, Mock mock)

I would like to override it to customize batch of types instead of Register(...) call for each type. Something like this:

```csharp
protected override object GetDefaultValue(Type type, Mock mock)
{
    var enumerableType = typeof(IEnumerable<>).MakeGenericType(type.GetGenericArguments());
    if (enumerableType.IsAssignableFrom(type))
    {
        var listType = typeof(List<>).MakeGenericType(type.GetGenericArguments());
        return Activator.CreateInstance(listType);
    }
    base.GetDefaultValue(type, mock);
}
```

stakx:
> But why is it sealed?

In order to guarantee a certain behavior with generic and array types.

(And because I believe that composition is generally better for composing behavior than inheritance, so I tend to define methods as either abstract, or sealed... but not virtual.)

If that's not what you want, you can always subclass the base class `DefaultValueProvider`.

Serg046:
Yes, DefaultValueProvider is good way but how to get empty values for other types that are not covered my `GetDefaultValue` implementation? I can get it from LookupOrFallbackDefaultValueProvider's `GetDefaultValue` method if it is public (by composition) or not sealed (by inheritance). It contains important logic like values for Task etc.

stakx:
Sorry I forgot to say, instead of trying to override `GetDefaultValue`, you *can* override `GetFallbackDefaultValue` which is what'll get invoked when there's no registered factory for a particular type.

Also, you can always use composition with / delegation to another default value provider (e.g. `DefaultValueProvider.Empty`) to cover those cases that your own provider does not handle itself.

Serg046:
I cannot use `DefaultValueProvider.Empty` because `GetDefaultValue` is protected internal.
But `GetFallbackDefaultValue` works for me, thank you!

stakx:
> I cannot use `DefaultValueProvider.Empty` because `GetDefaultValue` is protected internal.

(You cannot override, but you can use composition.)

> `GetFallbackDefaultValue` works for me, thank you!

Glad to hear. You're welcome!

Serg046:
Do you mean `DefaultValueProvider.Empty.GetDefaultValue(type, mock)` call?

stakx:
Sorry, you're right of course... I seem to have mis-remembered those methods' accessibility. Composition indeed doesn't seem possible in this case (which is probably a design flaw).

krizex:
> Sorry, you're right of course... I seem to have mis-remembered those methods' accessibility. Composition indeed doesn't seem possible in this case (which is probably a design flaw).

If so, how can I make a customized value provider which falls back to the Mock object if I did not register the type?

Issue #527, opened 2017-11-21
Verify that no unverified methods were called (alternative to Strict)


ashmind:
My goal in few specific tests is to verify that no unverified methods were called, for example:
```
var mailSender = new Mock<IMailSender>();

DoSomething(mailSender.Object);

mailSender.Verify(s => s.SendSpecificThing(), Times.Once);
mailSender.VerifyNotOtherCalls();
```

I know I can do this with `Strict`, but I don't like how `Strict` affects the test structure and moves some expectations into the "Arrange" part of the test (`Setup().Verifiable()`). It seems like `Verify(call)` is designed to provide a nicer alternative to `Strict`, so it would be nice to have that as well.

Alternatively, some kind of method matcher in `Verify` would work as well -- e.g.
`mailSender.VerifyAny(m => m.Name != nameof(SendSpecificThing), Times.Never);`

stakx:
> I don't like how `Strict` affects the test structure and moves some expectations into the "Arrange" part of the test (`Setup().Verifiable()`).

What does `.Verifiable()` have to do with mock behavior? All this does is flag a setup for inclusion in calls to `mock.Verify()` calls (as opposed to the non-discriminating `mock.VerifyAll()`, which will include *all* setups).

stakx:
```csharp
var mailSender = new Mock<IMailSender>(MockBehavior.Strict);
mock.Setup(s => s.SendSpecificThing()).AtMostOnce();
DoSomething(mailSender.Object);
mailSender.VerifyAll();
```

or, if you don't want to use the obsolete `.AtMostOnce()`:

```csharp
var mailSender = new Mock<IMailSender>(MockBehavior.Strict);
mock.Setup(s => s.SendSpecificThing());
DoSomething(mailSender.Object);
mailSender.Verify(s => s.SendSpecificThing(), Times.Once);
```

stakx:
> I don't like how `Strict` affects the test structure and moves some expectations into the "Arrange" part of the test [...].

Can you please elaborate a little? What expectations (say: setups) do you find acceptable in the "arrange" part, and which do you want to avoid?

You also mention `.Verifiable()`. That has nothing to do at all with a mock's behavior. It simply flags a setup so it's included in a verification via `mock.Verify()` (as opposed to the non-discriminating `mock.VerifyAll()` which always checks all setups).

ashmind:
[...]

PR #543, opened 2017-12-03
`VerifyNoOtherCalls`: Fix bugs with `Mock.Of<T>`

stakx:
This PR ensures that mocks created with `Mock.Of<T>` initially have no recorded invocations.

Due to the way `Mock.Of<T>` is currently implemented, mocks created using this method might already have recorded invocations. This never really mattered up until now, where it can interfere with `VerifyNoOtherCalls`.

This solves one of two issues with the initial `VerifyNoOtherCalls` implementation, as tracked in  https://github.com/moq/moq4/issues/527#issuecomment-348697314.

### DIFF
--- a/Source/Linq/Mocks.cs
+++ b/Source/Linq/Mocks.cs
@@ -94,7 +94,7 @@ namespace Moq
 		[Obsolete("Moved to Mock.Of<T>, as it's a single one, so no reason to be on Mocks.", true)]
 		public static T OneOf<T>() where T : class
 		{
-			return CreateMockQuery<T>().First<T>();
+			return Mock.Of<T>();
 		}
 
 		/// <summary>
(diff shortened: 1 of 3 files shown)


Source/Linq/Mocks.cs
```diff
@@ -108,7 +108,7 @@ public static class Mocks
 		[Obsolete("Moved to Mock.Of<T>, as it's a single one, so no reason to be on Mocks.", true)]
 		public static T OneOf<T>(Expression<Func<T, bool>> specification) where T : class
 		{
-			return CreateMockQuery<T>().First<T>(specification);
+			return Mock.Of<T>(specification);
```
stakx: Delegate to `Mock.Of<T>` (as claimed by the `[Obsolete]` attribute's message) so we don't need to duplicate the hack here.

Source/Linq/Mocks.cs
```diff
@@ -94,7 +94,7 @@ public static class Mocks
 		[Obsolete("Moved to Mock.Of<T>, as it's a single one, so no reason to be on Mocks.", true)]
 		public static T OneOf<T>() where T : class
 		{
-			return CreateMockQuery<T>().First<T>();
+			return Mock.Of<T>();
```
stakx: Delegate to Mock.Of<T> (as claimed by the [Obsolete] attribute's message).

Issue #547, opened 2017-12-08
`Mock.Of<T>()` is much slower than `new Mock<T>` (up to several orders of magnitude)


stakx:

informatorius:
I added a lambda compile cache to Moq and Mock.Of<T>() performance improves.
https://github.com/moq/moq4/issues/188

informatorius:
Is my understanding right that Mock.Of<Interface>() equals
```
var mock  = new Mock<Interface>();
mock.SetupAllProperties();
return mock.Object; 
```
If true can we just replace the

stakx:
@informatorius - I'd have to check whether the current implementation is equivalent to what you've shown above. If so, that's a possibility. But we cannot do the same for the parameterised `Mock.Of` overload.

I'll get back to you about that.

informatorius:
Is `Mock.Of<T>(predicate)` the same as 
```
var mock = new Mock<IFoo>();
mock.SetupAllProperties();
mock.Setup(predicate);
return mock.Object;
```

stakx:
Closing this issue as the performance of `Mock.Of<T>()` has now been taken care of. Thank you, @informatorius.

There's still the other method overload, `Mock.Of<T>(Func<T, bool>)` to deal with. Improving its performance would likely require some heavy refactoring regarding how Moq processes setups and LINQ setup expressions. This will probably deserve its own issue.

PR #548, opened 2017-12-10
Make `SetupAllProperties` semi-lazy

stakx:
This is a first step to improve the execution speed of `Mock.Of<T>` (see #547).


stakx:
Closed because it turned out that it wasn't so much eagerness / laziness of `SetupAllProperties` that contributes to `Mock.Of<T>`'s slowness, but copious use of reflection in `SetupAllProperties`. Making that method more lazy is still a good idea, but it won't really do anything for `Mock.Of<T>()`.

PR #549, opened 2017-12-10
Reduce use of reflection in `SetupAllProperties`

stakx:
Excessive reflection in `SetupAllProperties` contributes to `Mock.Of<T>`'s bad performance (#547).

This PR reduces use of reflection in this method by adding instances of specialized `MethodCall` types directly to `mock.Setups`, instead of installing setups by making calls to `mock.SetupProperty(m => m.Property, initialValue)` or `mock.SetupGet(m => m.Property).Returns(value)` through reflection.

### DIFF
--- a/Source/MethodCall.cs
+++ b/Source/MethodCall.cs
@@ -102,6 +102,19 @@ namespace Moq
 		private Exception throwExceptionResponse;
 		private bool verifiable;
 
+		/// <remarks>
+		///   Only use this constructor when you know that the specified <paramref name="method"/> has no `out` parameters,
+		///   and when you want to avoid the <see cref="MatcherFactory"/>-related overhead of the other constructor overload.
+		/// </remarks>
+		public MethodCall(Mock mock, Condition condition, Expression originalExpression, MethodInfo method, IMatcher[] argumentMatchers)
+		{
+			this.argumentMatchers = argumentMatchers;
+			this.condition = condition;
+			this.method = method;
+			this.mock = mock;
+			this.originalExpression = originalExpression;
+		}
+
 		public MethodCall(Mock mock, Condition condition, Expression originalExpression, MethodInfo method, params Expression[] arguments)
 		{
 			this.mock = mock;
--- a/Source/Mock.cs
+++ b/Source/Mock.cs
@@ -834,42 +834,22 @@ namespace Moq
 					   ProxyFactory.Instance.IsMethodVisible(p.GetGetMethod(), out _))
 				.Distinct();
 
-			var setupPropertyMethod = mock.GetType().GetMethods("SetupProperty")
-				.First(m => m.GetParameters().Length == 2);
-			var setupGetMethod = mock.GetType().GetMethods("SetupGet")
-				.First(m => m.GetParameters().Length == 1);
-
 			foreach (var property in properties)
 			{
 				var expression = GetPropertyExpression(mockType, property);
-				object initialValue = GetInitialValue(mock, mockedTypesStack, property);
+				object value = GetInitialValue(mock, mockedTypesStack, property);
 
-				var mocked = initialValue as IMocked;
+				var mocked = value as IMocked;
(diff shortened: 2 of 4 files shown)


Source/Mock.cs
```diff
@@ -834,42 +834,22 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
...
-			var setupPropertyMethod = mock.GetType().GetMethods("SetupProperty")
-				.First(m => m.GetParameters().Length == 2);
-			var setupGetMethod = mock.GetType().GetMethods("SetupGet")
-				.First(m => m.GetParameters().Length == 1);
-
 			foreach (var property in properties)
 			{
 				var expression = GetPropertyExpression(mockType, property);
-				object initialValue = GetInitialValue(mock, mockedTypesStack, property);
+				object value = GetInitialValue(mock, mockedTypesStack, property);
```
stakx: Renamed because this variable no longer just holds the initial value; it acts as the backing field for the set up property. It gets captured by both setups' getter / setter lambdas.

Source/MethodCall.cs
```diff
@@ -102,6 +102,19 @@ internal partial class MethodCall : ICallbackResult, IVerifies, IThrowsResult
 		private Exception throwExceptionResponse;
 		private bool verifiable;
 
+		/// <remarks>
+		///   Only use this constructor when you know that the specified <paramref name="method"/> has no `out` parameters,
+		///   and when you want to avoid the <see cref="MatcherFactory"/>-related overhead of the other constructor overload.
+		/// </remarks>
+		public MethodCall(Mock mock, Condition condition, Expression originalExpression, MethodInfo method, IMatcher[] argumentMatchers)
```
stakx: For the type of setups we want to install (i. e. setups that implement property getters and setters), we know in advance what kinds of matchers we'll need. The existing constructor would however force us to build an `It.IsAny<T>()` expression, which would then be put through the (slow!) `MatcherFactory.CreateMatcher` factory method. This new constructor allows us to skip these two unnecessary steps.

Source/Mock.cs
```diff
@@ -834,42 +834,22 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
...
 
-				var mocked = initialValue as IMocked;
+				var mocked = value as IMocked;
 				if (mocked != null)
 				{
 					SetupAllProperties(mocked.Mock, mockedTypesStack);
 				}
 
+				mock.Setups.Add(new PropertyGetterMethodCall(mock, expression, property.GetGetMethod(true), () => value));
```
stakx: The above LINQ query only returns properties that are readable, so we don't need to check `property.CanRead`.

kzu:
Glad to see this! Perf. should be comparable with v5, see the just-added [PropertyBehavior](https://github.com/moq/moq/blob/master/src/Moq/Moq.Sdk/PropertyBehavior.cs) :)

It's added wholesale to [every loose mock by default](https://github.com/moq/moq/blob/master/src/Moq/Moq/contentFiles/cs/netstandard1.3/Mock.cs#L25) so `SetupAllProperties` shouldn't be needed anymore even!

stakx:
@kzu - should something similar be done in Moq 4? Auto setting up property behavior for get-settable properties of loose mocks seems entirely feasible; we could modify the interception pipeline such that when a property is queried or set for the first time, a property setup is created on the fly. (This might be perceived as a breaking change, though.)

stakx:
Btw., regarding performance: Moq 4 performance is now back to what it was around version 4.2 (despite being more thread-safe and feature complete). I am expecting that it will get even faster once the remaining filed bugs are removed--essentially because fixing them will entail reducing expression tree compilation and doing plain expression tree rewriting instead, which is orders of magnitude faster. In particular, perf will greatly benefit on x64. I've prototyped the necessary changes, but it'll be quite difficult to pull this off without causing any minor breaking changes.

PR #550, opened 2017-12-10
`SetupAllProperties`: Defer property initialization

stakx:
Let `SetupAllProperties` defer initialization of set up properties to the point when they are queried for the very first time. Doing this has two advantages:

 1. There is no more immediate difference in execution speed between `DefaultValue.Empty` and `DefaultValue.Mock`. (Up until now, `SetupAllProperties` could be orders of magnitude slower when used with the latter default value mode.)

 2. Type recursion / loops in the object graph are no longer an issue (hence the changes to the unit tests). We no longer need to keep track of already mocked types, so we can get rid of `GetInitialValue` and call `GetDefaultValue` directly.

### DIFF
--- a/Source/Mock.cs
+++ b/Source/Mock.cs
@@ -803,19 +803,18 @@ namespace Moq
 		[DebuggerStepThrough]
 		internal static void SetupAllProperties(Mock mock)
 		{
-			PexProtector.Invoke(SetupAllPropertiesPexProtected, mock);
+			PexProtector.Invoke(SetupAllPropertiesPexProtected, mock, mock.DefaultValueProvider);
+			//                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^
+			// `SetupAllProperties` no longer performs eager recursive property setup like in previous versions.
+			// If `mock` uses `DefaultValue.Mock`, mocked sub-objects are now only constructed when queried for
+			// the first time. In order for `SetupAllProperties`'s new mode of operation to be indistinguishable
+			// from how it worked previously, it's important to capture the default value provider at this precise
+			// moment, since it might be changed later (before queries to properties of a mockable type).
 		}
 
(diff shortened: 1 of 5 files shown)


Source/Mock.cs
```diff
@@ -838,44 +837,47 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
...
+						}
+
+						if (initialValue is IMocked mocked)
+						{
+							SetupAllPropertiesPexProtected(mocked.Mock, defaultValueProvider);
+						}
 
-				mock.Setups.Add(new PropertyGetterMethodCall(mock, expression, getter, () => value));
+						value = initialValue;
+						valueNotSet = false;
```
stakx: In theory, there's room for a race condition here. The uninitialized property could be queried simultaneously several times, or it could be queried while the setter is running, too. In practice, this might not matter much with regards to correctness, but it *might* cause some inefficiency because multiple initial values are produced when only one is required.

Source/Mock.cs
```diff
@@ -838,44 +837,47 @@ private static void SetupAllProperties(Mock mock, Stack<Type> mockedTypesStack)
...
-
-			mockedTypesStack.Pop();
-		}
-
-		private static object SetupAllProperties_GetInitialValue(Mock mock, Stack<Type> mockedTypesStack, MethodInfo getter)
-		{
-			if (mockedTypesStack.Contains(getter.ReturnType))
-			{
-				// to deal with loops in the property graph
-				return mock.GetDefaultValue(getter, useAlternateProvider: DefaultValueProvider.Empty);
```
stakx: We no longer need to guard against "loops in the property graph" since `SetupAllProperties` only sets up the properties of the current level in the object's hierarchy. The setting up of deeper levels in the hierarchy must be triggered by querying a property.

Issue #504, opened 2017-10-24
Find out why Moq's performance has become worse since version approx. 4.5


stakx:
It was noted e.g. in https://github.com/moq/moq4/issues/388#issuecomment-334870636 that Moq's performance has suffered in recent releases. It would be nice to find the reason(s) for this, and perhaps apply some optimizations to start reversing that trend *before* version 4.8.0 gets released.

Suggestion:

* [ ] Create some Moq test code that does something non-trivial.
* [ ] Profile that code for a couple of Moq releases, going back to approx. version 4.5.0. (Perhaps using PerfView, or another capable, freely available .NET profiling tool.)
* [ ] Report back with conclusion what has affected Moq's performance negatively in more recent releases / what are the major performance bottlenecks.

/cc @jatouma, @JohanLarsson

toha73:
I noticed this since we're using `Mock.Of<>()` in alot of places in our unit tests.
Since NuGet version 4.7.99 it's extremely slow to use that method.

stakx:
@toha73: Have you by change done any profiling and found out why that is? Do you possibly have some code / use cases you could share to be used for profiling?

toha73:
I just did a simple script in LinqPad to test it out between different versions of Moq.

```
void Main()
{
	var iterations = Enumerable.Range(0, 500).ToList();
	
	var sw = Stopwatch.StartNew();
	iterations.ForEach(_ => { var x = Mock.Of<IMyInterface>(); });
	Console.WriteLine($"{sw.ElapsedMilliseconds} ms");

	sw.Restart();
	iterations.ForEach(_ => { var x = new Mock<IMyInterface>().Object; });
	Console.WriteLine($"{sw.ElapsedMilliseconds} ms");
}

public interface IMyInterface { string Property1 { get; set; } }
```

And the result on my machine is: 

**Moq 4.7.142**

> 5939 ms
> 3 ms

**Moq 4.7.99**

> 6444 ms
> 3 ms

**Moq 4.7.63**

> 287 ms
> 4 ms

JohanLarsson:
Ran a quick benchmark comparing with NSubstitute
``` ini

BenchmarkDotNet=v0.10.9, OS=Windows 7 SP1 (6.1.7601)
Processor=Intel Xeon CPU E5-2637 v4 3.50GHzIntel Xeon CPU E5-2637 v4 3.50GHz, ProcessorCount=16
[...]

Issue #98, opened 2014-02-20
Mocking Interfaces


kapdave:
My apologies if I missed an update or have done something wrong in my code.  I have updated my moq version and it seems as though I can no longer mock an interface without receiving a System.TypeLoadException?  Any help with this issue?  The following is a simple snippet of code that is generating the error:

```
using Moq;
namespace ConsoleApplication4 {
    class Program {
        static void Main(string[] args) {

            //Mock<ICrapGoat> asdf = new Mock<ICrapGoat>(); // <-- Throws System.TypeLoadException
            Mock<CrapGoat> asdf = new Mock<CrapGoat>(); // <-- Need to use concrete instance
            var asdfsdfvcsdafr = asdf.Object;
        }
    }

    public interface ICrapGoat {}
    public class CrapGoat : ICrapGoat {}
}
```

PR #551, opened 2017-12-11
Make interface proxy creation twice as fast

stakx:
Creating proxies with DynamicProxy is the single most expensive kind of operation that Moq does. Commit 54e132b changed how proxies for interface types are generated so that `object` methods on interfaces could also be intercepted (and thus set up). Unfortunately, this change also made interface proxy generation twice as costly!

This commit restores the previous method for interface proxy creation, and at the same time adds an `InterfaceProxy` class to keep `object` methods interceptable.

This should bring Moq's performance back to, or very close to, what it used to be before version 4.5.16.

/cc #504

### DIFF
--- a/Source/ProxyFactories/InterfaceProxy.cs
+++ b/Source/ProxyFactories/InterfaceProxy.cs
@@ -1,0 +1,116 @@
+﻿//Copyright (c) 2007. Clarius Consulting, Manas Technology Solutions, InSTEDD
+//https://github.com/moq/moq4
+//All rights reserved.
+
+//Redistribution and use in source and binary forms, 
+//with or without modification, are permitted provided 
+//that the following conditions are met:
+
+//    * Redistributions of source code must retain the 
+//    above copyright notice, this list of conditions and 
+//    the following disclaimer.
+
+//    * Redistributions in binary form must reproduce 
+//    the above copyright notice, this list of conditions 
+//    and the following disclaimer in the documentation 
+//    and/or other materials provided with the distribution.
+
+//    * Neither the name of Clarius Consulting, Manas Technology Solutions or InSTEDD nor the 
+//    names of its contributors may be used to endorse 
+//    or promote products derived from this software 
+//    without specific prior written permission.
+
+//THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+//CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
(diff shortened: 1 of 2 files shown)


Source/ProxyFactories/InterfaceProxy.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Reflection;
+
+namespace Moq.Internals
+{
+	/// <summary>Do not use. (Moq requires this class so that <see langword="object"/> methods can be set up on interface mocks.)</summary>
+	// NOTE: This type is actually specific to our DynamicProxy implementation of `ProxyFactory`.
+	// It is public because it needs to be visible to the framework's types in `Reflection.Emit`.
+	// Not sure what `[assembly: InternalsVisibleTo(<some framework assembly>)]` would have to look like.
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public abstract class InterfaceProxy
```
stakx: It's a pity this internal type needs to be `public` for now. See #98 and https://stackoverflow.com/q/47761193/240733.

Source/ProxyFactories/InterfaceProxy.cs
```diff
@@ -0,0 +1,116 @@
...
+		private static MethodInfo getHashCodeMethod = typeof(object).GetMethod("GetHashCode", BindingFlags.Public | BindingFlags.Instance);
+		private static MethodInfo toStringMethod = typeof(object).GetMethod("ToString", BindingFlags.Public | BindingFlags.Instance);
+
+		/// <summary/>
+		[DebuggerHidden]
+		public sealed override bool Equals(object obj)
+		{
+			// Forward this call to the interceptor, so that `object.Equals` can be set up.
+			var invocation = new Invocation(equalsMethod, obj);
+			((IInterceptor)((IMocked)this).Mock).Intercept(invocation);
```
stakx: When an `object` method is invoked on an interface proxy object, DynamicProxy will be bypassed, and we'll end up here, in the base class. The `Mock.Intercept(Invocation)` method doesn't care *who* intercepted a method, so we can just create an `Invocation` representing the `object` method call, and send it off to be intercepted.

Source/ProxyFactories/InterfaceProxy.cs
```diff
@@ -0,0 +1,116 @@
...
+//SUCH DAMAGE.
+
+//[This is the BSD license, see
+// http://www.opensource.org/licenses/bsd-license.php]
+
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Reflection;
+
+namespace Moq.Internals
```
stakx: Let's try not to pollute the `Moq` namespace with something that should be internal.

PR #569, opened 2018-01-07
Expose invocations on mock object

Tragedian:
First take at exposing invocation information.
Mock invocations exposed as `IReadOnlyList<Invocation>` property.
Internal abstract `Invocation` type exposed.

### DIFF
--- a/Source/Invocation.cs
+++ b/Source/Invocation.cs
@@ -39,13 +39,17 @@
 // http://www.opensource.org/licenses/bsd-license.php]
 
 using System.Collections;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Reflection;
 using System.Text;
 
 namespace Moq
 {
-	internal abstract class Invocation
+	/// <summary>
+	/// Represents the invocation of a method.
+	/// </summary>
+	public abstract class Invocation
 	{
 		private object[] arguments;
 		private MethodInfo method;
@@ -73,11 +77,16 @@ namespace Moq
 		/// <summary>
 		/// Gets the arguments of the invocation.
 		/// </summary>
+		public IReadOnlyList<object> Arguments => this.ArgumentsArray;
+
+		/// <summary>
+		/// Gets the arguments of the invocation.
+		/// </summary>
 		/// <remarks>
 		/// Arguments may be modified. Derived classes must ensure that by-reference parameters are written back
 		/// when the invocation is ended by a call to any of the three <c>Returns</c> methods.
 		/// </remarks>
-		public object[] Arguments => this.arguments;
+		internal object[] ArgumentsArray => this.arguments;
 
 		internal bool Verified => this.verified;
 
@@ -88,10 +97,10 @@ namespace Moq
 		/// Implementations may assume that this method is only called for a <see langword="void"/> method,
 		/// and that no more calls to any of the three <c>Return</c> methods will be made.
 		/// <para>
-		/// Implementations must ensure that any by-reference parameters are written back from <see cref="Arguments"/>.
+		/// Implementations must ensure that any by-reference parameters are written back from <see cref="ArgumentsArray"/>.
 		/// </para>
 		/// </remarks>
-		public abstract void Return();
+		internal abstract void Return();
 
 		/// <summary>
 		/// Ends the invocation as if a tail call to the base method were made.
@@ -100,10 +109,10 @@ namespace Moq
 		/// Implementations may assume that this method is only called for a method having a callable (non-<see langword="abstract"/>) base method,
 		/// and that no more calls to any of the three <c>Return</c> methods will be made.
 		/// <para>
-		/// Implementations must ensure that any by-reference parameters are written back from <see cref="Arguments"/>.
+		/// Implementations must ensure that any by-reference parameters are written back from <see cref="ArgumentsArray"/>.
 		/// </para>
 		/// </remarks>
-		public abstract void ReturnBase();
+		internal abstract void ReturnBase();
 
 		/// <summary>
 		/// Ends the invocation as if a <see langword="return"/> statement with the specified return value occurred.
@@ -112,7 +121,7 @@ namespace Moq
 		/// Implementations may assume that this method is only called for a non-<see langword="void"/> method,
 		/// and that no more calls to any of the three <c>Return</c> methods will be made.
 		/// <para>
-		/// Implementations must ensure that any by-reference parameters are written back from <see cref="Arguments"/>.
+		/// Implementations must ensure that any by-reference parameters are written back from <see cref="ArgumentsArray"/>.
 		/// </para>
 		/// </remarks>
 		public abstract void Return(object value);
@@ -136,7 +145,7 @@ namespace Moq
 			{
 				builder.Append(method.Name, 4, method.Name.Length - 4);
 				builder.Append(" = ");
-				builder.AppendValueOf(this.Arguments[0]);
+				builder.AppendValueOf(this.ArgumentsArray[0]);
 			}
 			else
 			{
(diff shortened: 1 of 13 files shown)


stakx:
@Tragedian, thanks for the PR! I'll review it in a short while.

Source/Invocation.cs
```diff
@@ -70,14 +74,19 @@ protected Invocation(MethodInfo method, params object[] arguments)
 		/// </summary>
 		public MethodInfo Method => this.method;
 
+		/// <summary>
+		/// Gets the arguments of the invocation.
+		/// </summary>
+		public IReadOnlyList<object> Arguments => this.ArgumentsArray;
```
kzu: Exposing the underlying array as-is doesn't make it magically read only heart because the interface says so. But I guess it's not critical anyway.

Tragedian: I think this is a failure of the interface's name; it's a readable list, rather than one which can only be read. If it needs to be immutable, a copy-on-read method could be used.

stakx: I don't see the practical benefits of making recorded invocations truly immutable; if someone feels the need to cast an interface back to the underlying runtime type, why not just let them? The fact that we're using an interface should be a strong enough signal against doing that. IMHO casting to an interface is good enough in practice.

stakx:
Hi @Tragedian, sorry for the delay. I've been looking at, and thinking about your PR a few times, and while it's generally solid, there are a few things that I think could use some adjustment. :-) I'd like to do a few experiments of my own before I submit a proper detailed review. Until then, let me quickly summarize the "pain points" I've found:

1. I don't particularly like the name changes from `Arguments` to `ArgumentsArray` and from `Invocations` to `InvocationCollection`. I can see the reason why you did it, but I'm wondering if there really isn't a better solution that would enable us to keep the shorter, non-technical names.

2. I think there could be some more consistency regarding the types that we expose publicly. Right now, it's `IReadOnlyList<Invocation>` and `Invocation` (an interface and a concrete class). Something about that bothers me. I am wondering whether we should either expose `IReadOnlyList<IInvocation>` and `Invocation` (that is, go all interfaces), or `InvocationCollection` and `Invocation` (that is, follow the current approach of Moq 4's public API and go all concrete classes). In the latter case, we'd run into one problem (that being `Invocation.Arguments` which should be exposed as read-only).

3. You're making some things public that should stay internal for now, e. g. `InvocationCollection.ToArray(Func<Invocation, bool>)`. I'll elaborate later why that should stay internal. But generally, I think it would be better to expose as few types and members as possible.

I'd like to think about this some more. Perhaps you have some ideas on the above points?

Tragedian:
Thanks for taking the time to compose your thoughts to give me such quality feedback. I'll address them as you've given them.

> 1. I don't particularly like the name changes from Arguments to ArgumentsArray and from Invocations to InvocationCollection. I can see the reason why you did it, but I'm wondering if there really isn't a better solution that would enable us to keep the shorter, non-technical names.

I don't particularly like them either. Really it's just a name-change to give the non-technical property names to the public API. I don't see a great answer for exposing a good read-only public API alongside the need to keep these writable properties internal. A heavier refactor could be to replace the setters with `Set` methods whilst still reading from the public properties. This may be a better answer, but I was afraid of making too big a change without further prompting.

> 2. I think there could be some more consistency regarding the types that we expose publicly. Right now, it's IReadOnlyList<Invocation> and Invocation (an interface and a concrete class). Something about that bothers me. I am wondering whether we should either expose IReadOnlyList<IInvocation> and Invocation (that is, go all interfaces), or InvocationCollection and Invocation (that is, follow the current approach of Moq 4's public API and prefer concrete classes). In the latter case, we'd run into one problem (that being Invocation.Arguments which should be exposed as read-only).

I personally don't have any discomfort in having a collection interface return a concrete class. Where APIs choose to expose a collection interface, it's usually whilst providing a collection of concrete values exposed by the API. I don't have any objections to exposing a read-only `InvocationCollection` class instead of an interface, and I'm quite happy to update the branch to go that way.

I do however feel uncomfortable creating an interface with a 1:1 mapping to a concrete class. Unless this is looking into the future and anticipating a scenario where you would want to provide multiple implementations of that interface, it seems like it's creating work just to side-step the work in making `Invocation` public.

> 3. You're making some things public that should stay internal for now, e. g. InvocationCollection.ToArray(Func<Invocation, bool>). I'll elaborate later why that should stay internal. But generally, I think it would be better to expose as few types and members as possible.

I don't think those types are any more public than they were already. They were marked `public` before changing their return-type. If we were to make `InvocationCollection` a public type, I would agree these would certainly need to be changed to `internal` methods to maintain their current level of access.

stakx:
@Tragedian: Thank you for replying, many good thoughts there, and I think I agree with much of what you're saying. I'd love to give you a very detailed response, but I fear I might go off on several tangents and completely lose focus, so I'll try to keep it brief:

> I don't see a great answer for exposing a good read-only public API alongside the need to keep these writable properties internal.

I believe we wouldn't have this problem if the public API consisted mostly of interface types (e. g. `IMock<T>`, `IInvocation`, `ISetup`), perhaps a single factory class (e. g. `Mock` or `MockFactory`), and a few abstract base classes for extensibility (e. g. `DefaultValueProvider`). That would give us a lot more flexibility shuffling around internals to our liking, then re-mapping them to the public interfaces. Unfortunately, that's not how Moq 4's API is built.

> A heavier refactor could be [...].

There are many ways how we could solve our present dilemma. Before we look at those, would it perhaps help to look at the *smallest* possible code change that achieves our goal?

I believe the smallest change that would expose invocations would be this:

1. Define the following public interface:

   ```csharp
   public interface IInvocation
   {
       MethodInfo Method { get; }
       IReadOnlyList<object> Arguments { get; }
   }
   ```

   (Not sure if this qualifies for the 1:1 correspondence that you'd be uncomfortable with? I personally think it would be totally fine, it's just an interface containing that read-only subset of `Invocation` which should be publicly visible.)

2. Let `Invocation` implement `IInvocation`. This is very straightforward (1 or 2 added LoC, plus some XML documentation comments).

3. Add the following method to `Mock<T>`:

   ```csharp
   public IReadOnlyList<IInvocation> GetInvocations() => this.Invocations.ToArray();
   ```

This method is essentially `mock.Invocations.GetSnapshot()`, but one level further up the object graph: `mock.GetInvocations()`. I am not a huge fan of it, having a `Invocations` property would be nicer. But perhaps that would only make sense when we actually return a regular .NET collection that you can use with LINQ etc. This is not currently the case.

Any thoughts on this?

Tragedian:
The footprint for adding `IInvocation` is very small, I must admit. But as a solution, it does have the problem I mentioned: there's only ever one implementation of this interface and it feels like there will only ever be one implementation, which makes its existence only to avoid making the more significant changes necessary to expose the existing type.

In other words, if designing from scratch, I would not make this interface; I would just expose the type directly. However that is not the scenario we're in right now. I am happy to disagree on this choice and commit to refactoring these changes behind such an interface. Better to disagree and commit if it's just the two of us 👍

stakx:
> [...] which makes its existence only to avoid making the more significant changes necessary to expose the existing type.

You do have a point there. It boils down to the question of what bears the bigger cost in the long term: having two types where one would possibly suffice; or doing a much larger refactoring to be able to expose the present type, which unfoetunately has to include renaming things to less-than-ideal names (at least internally).

There isn't a right or wrong solution here, but I definitely tend towards the present, nicer names (which we encounter in many parts of Moq's source code and have to deal with over and over again) and a "sort-of-redundant" role interface that just sits there for the public API, but doesn't really hurt anyone... or does it? Is it actually harmful besides appearing to be slightly redundant?

> I am happy to disagree on this choice and commit to refactoring these changes behind such an interface. Better to disagree and commit if it's just the two of us 👍

Fine with me, if you want to see this happen quickly.

(I find this a surprisingly difficult design decision to make, because both of our proposals have their own respespective disadvantages. Perhaps some more drafting would lead us to a solution that is better than either one we've looked at so far. I'd be willing to continue this discussion if we want to aim for the best possible solution but I understand if you're OK with a more pragmatic solution, too.)

stakx:
(Previous post of mine deleted. I'd really like to think about your point of view some more.)

Tragedian:
Any further comments or direction? I'd love to get this stuff closer to production.

stakx:
@Tragedian, sorry for being slow in responding, I've been down with a cold during the past 2 weeks or so, and now on vacation for another 2 weeks. I've been looking more deeply into this two weeks ago, I'll try to give you a response as soon as I get a chance. Sorry again for the delay.

Tragedian:
The Moq 5 API is frozen against further change. Happy to move my efforts to Moq 5.

Tragedian:
New pull request: https://github.com/moq/moq4/pull/627

I suggest we resume our conversation over there.

Issue #567, opened 2018-01-05
After v4.8.0 upgrade and tests failing with missing netstandard, version=2.0.0.0


salixzs:
After upgrading to Moq 4.8 on build server (where we do not have anything netstandard 2.0) started to fail with this:
```
Information: [xUnit.net 00:00:10.0169373]       System.TypeInitializationException : The type initializer for 'Moq.DefaultValueProvider' threw an exception.
Information: [xUnit.net 00:00:10.0170013]       ---- System.IO.FileNotFoundException : Could not load file or assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
Information: [xUnit.net 00:00:10.0174190]       Stack Trace:
Information: [xUnit.net 00:00:10.0175399]         C:\projects\moq4\Source\DefaultValueProvider.cs(59,0): at Moq.DefaultValueProvider.get_Empty()
Information: [xUnit.net 00:00:10.0176097]         C:\projects\moq4\Source\Mock.Generic.cs(144,0): at Moq.Mock`1..ctor(MockBehavior behavior, Object[] args)
Information: [xUnit.net 00:00:10.0176671]         C:\projects\moq4\Source\Mock.Generic.cs(128,0): at Moq.Mock`1..ctor(MockBehavior behavior)
Information: [xUnit.net 00:00:10.0177262]         C:\projects\moq4\Source\Mock.Generic.cs(112,0): at Moq.Mock`1..ctor()
```

Could this be related to recently added
``
<dependency id="System.Threading.Tasks.Extensions" version="4.4.0" />
``

packages.config:
```
  <package id="Castle.Core" version="4.2.1" targetFramework="net462" />
  <package id="Moq" version="4.8.0" targetFramework="net462" />
  <package id="System.Threading.Tasks.Extensions" version="4.4.0" targetFramework="net462" />
  <package id="xunit.abstractions" version="2.0.1" targetFramework="net462" />
  <package id="xunit.core" version="2.3.1" targetFramework="net462" />
  <package id="xunit.extensibility.core" version="2.3.1" targetFramework="net462" />
  <package id="xunit.extensibility.execution" version="2.3.1" targetFramework="net462" />
  <package id="xunit.runner.visualstudio" version="2.3.1" targetFramework="net462" developmentDependency="true" />
```
Solution targets Full .Net Framework 4.6.2

stakx:
[...]

Issue #566, opened 2018-01-03
MissingMethodException / "Method not found" after update to Moq 4.8.0


JHotterbeekxBannerconnect:
Hi,

Today I've tried updating the NuGet package from 4.7.145.0 to 4.8.0.0 and I'm running into an issue with a specific test case failing now on a strange error message. To give you some context, in our tests we create a HttpActionContext to test our authentication. This is native .net functionality. I've isolated the issue to the following code:

    [Test]
    public void OnActionExecuting_ShouldFillMarketInSession() {
      var fakeActionContext = _CreateFakeHttpActionContext();
    }

    private HttpActionContext _CreateFakeHttpActionContext() {
      return new HttpActionContext(null, new ReflectedHttpActionDescriptor()) {
        Response = new HttpResponseMessage(HttpStatusCode.OK)
      };
    }

When running this test, it fails with the following error:

> System.MissingMethodException : Method not found: 'Void System.Web.Http.Controllers.HttpActionContext.set_Response(System.Net.Http.HttpResponseMessage)'.
   at SMP.Api.Test.Filter.MoqTest._CreateFakeHttpActionContext()
   at SMP.Api.Test.Filter.MoqTest.OnActionExecuting_ShouldFillMarketInSession() in C:\Projects\Athena\Athena\SMP.Api.Test\Filter\MoqTest.cs:line 14

Could you help me figure out what is wrong, or fix this if it is an actual bug?

Thanks,
John

stakx:
I see nothing related to Moq in your example code. Could you please provide a short but complete code example that demonstrates this issue?

willykc:
Hi, I think I'm seeing the same problem. Out of 405 tests using Moq, 16 started failing after upgrading to Moq 4.8.0 with exception "System.MissingMethodException: Method not found". It seems to be related to the "System.Net.Http" namespace since all the "missing methods" in the failing tests have types from that namespace in their signature (eg. "System.Net.Http.HttpResponseMessage").

stakx:
@JHotterbeekxBannerconnect, @willykc: Could it be that you updated not just Moq, but any other Microsoft ASP.NET packages at the same time without noticing it? Because quite frankly, Moq has no dependency on, nor any special logic related to, any `System.Net.Http.*` types. I don't see how this issie could be caused by Moq, especially given that the exception gets triggered in a piece of code that doesn't even involve Moq.

JHotterbeekxBannerconnect:
[...]

PR #571, opened 2018-01-08
Downgrade System.ValueTuple & System.Threading.Tasks.Extensions

stakx:
Hopefully, this will take care of the better part of #566 and #567 and restore better compatibility of Moq with the full .NET Framework <4.7.1.

### DIFF
--- a/Moq.nuspec
+++ b/Moq.nuspec
@@ -15,7 +15,7 @@
 		<dependencies>
 			<group targetFramework=".NETFramework4.5">
 				<dependency id="Castle.Core" version="4.2.1" />
-				<dependency id="System.Threading.Tasks.Extensions" version="4.4.0" />
+				<dependency id="System.Threading.Tasks.Extensions" version="4.3.0" />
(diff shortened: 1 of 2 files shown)


abatishchev:
Is it possible to do not reference other/Core packages on Full .NET?

stakx:
Yes, but then Moq loses `ValueTask<>` and `ValueTuple<>` support.

abatishchev:
For me it would be fine. I'd like to see this support via an additional package, i.e. Moq.ValueTuple, or via a major version bump, I believe adding new major dependencies is a good reason for that. Personally hate to install additional packages which I don't need.

stakx:
See the discusssion in #566 about a separate package. While possible, the split would be a little awkward. I'd rather just remove `ValueX<>` support completely until .NET tooling is reasonably reliable again.

abatishchev:
Thanks, will read it. Meanwhile I'm all for the removal of such until a better time.

Issue #572, opened 2018-01-09
Setup cannot invoke callback with different number of parameters in Moq 4.8.1


Caraul:
Hi,

looks like adding ValidateReturnDelegate makes extension methods unusable in zero argument methods such as property getters. Here is small repro that fails at Setup phase with ArgumentException:

```
using Moq;

namespace TestMoq
{
	internal class Program
	{
		internal static void Main()
		{
			var a = new A();
			var mock = new Mock<A>();
			mock.SetupGet(aa => aa.P).Returns(a.Pext);
		}
	}

	public class A
	{
		public virtual string P { get; set; }
	}

	internal static class Aext
	{
		public static string Pext(this A a)
		{
			return "42";
		}
	}
}
```
Not sure if this is expected effect - please could anyone tell is such behavior correct?

stakx:
@Caraul, thanks for reporting. Without looking into this in detail, this looks like a regression introduced in 4.8.0. This used to work in Moq 4.7.145, but I'm not sure whether that was by design or incidental. Will try to look into this in the next few days.

stakx:
It's indeed a regression caused by 104c6464eccf9f2b643bf077adcf6da8d612e7d3 (which aimed to make delegate validation more strict, as you've already found out). We didn't have a test for the extension method case you've provided. I'll fix this over the next few days... or would you like to provide a PR yourself?

Caraul:
Well, fix won't take weeks but one week maybe cause I need to go deeper inside Moq - if next week ASAP is acceptable then I'm in with PR.

stakx:
@Caraul, that is fine. :+1:

I expect that a local fix in just that validation method should be possible; check whether the callback is an extension method (`IsStatic` & `IsDefined(typeof(ExtensionAttribute))`), and if so, subtract 1 from the parameter count when matching. Something like that.

Please don't forget to also add one or more unit tests to the test suite. :)

Thank you!

Caraul:
Thanks for moving Moq forward - [Good job!](https://youtu.be/S0UvJZmGTsk) 👍

stakx:
Moq 4.8.2 (which includes your patch, see [changelog](https://github.com/moq/moq4/blob/master/CHANGELOG.md#482-2018-02-23)) has just been published [on NuGet](https://www.nuget.org/packages/Moq/4.8.2).

Caraul:
Great news, thank you very much!

PR #575, opened 2018-01-17
Issue #572 Setup cannot invoke callback with different number of parameters in Moq 4.8.1

Caraul:
Add more parameters to test method as 0 and 1 parameters are special cases for parameterless and extension methods.
Add tests for extension methods with different number of parameters.

### DIFF
--- a/Source/MethodCallReturn.cs
+++ b/Source/MethodCallReturn.cs
@@ -43,6 +43,7 @@ using System.Globalization;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+
 using Moq.Language;
 using Moq.Language.Flow;
 using Moq.Properties;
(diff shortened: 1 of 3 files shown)


Source/MethodCallReturn.cs
```diff
@@ -42,7 +42,7 @@
 using System.Globalization;
 using System.Linq.Expressions;
 using System.Reflection;
-
+using System.Runtime.CompilerServices;
```
stakx: Please move this line a little further up so that `using System.*` are properly sorted. Also, restore the blank line between `using System.*` and `using Moq.*`.

Caraul: About sorting - Runtime is definitely after Reflection, isn't it?
Blank line is my fault - suspect VS dit that :)

stakx: Uhhh... yes indeed. You're right, of course. :-)

stakx:
Merged. Thanks again for taking the time to submit a fix! :+1:

PR #583, opened 2018-01-25
Ignore CallBase for members of additional interfaces

### LINKED ISSUE #582: NotImplementedException DynamicProxy2 error when CallBase = true while mocking interfaces again
This code:
```
    public class Tests
    {
        public interface IFoo
        {
            string Foo { get; }
        }

        public class Bar
        {
        }

        [Fact]
        public void Test()
        {
            // arrange
            var bar = new Mock<Bar>() { CallBase = true };
            bar.As<IFoo>();

            var foo = (IFoo)bar.Object;
            Mock.Get(foo).DefaultValue = DefaultValue.Mock; // this can also be done on nodeMock above, issue still reproduces in 4.8.1 and still doesn't in 4.8.0

            //Mock.Get(foo).SetupGet(obj => obj.Foo).Returns("foo"); // uncomment me to work around this issue

            // act
            var dummy = foo.Foo;
        }
    }
```

Throws:
> Result StackTrace:	
> at Castle.DynamicProxy.AbstractInvocation.ThrowOnNoTarget()
>    at Castle.DynamicProxy.Internal.CompositionInvocation.EnsureValidTarget()
>    at Castle.Proxies.Invocations.IFoo_get_Foo.InvokeMethodOnTarget()
>    at Castle.DynamicProxy.AbstractInvocation.Proceed()
>    at Moq.CastleProxyFactory.Invocation.ReturnBase() in C:\projects\moq4\Source\ProxyFactories\CastleProxyFactory.cs:line 201
>    at Moq.Return.Handle(Invocation invocation, Mock mock) in C:\projects\moq4\Source\Interception\InterceptionAspects.cs:line 351
>    at Moq.Mock.Moq.IInterceptor.Intercept(Invocation invocation) in C:\projects\moq4\Source\Interception\Mock.cs:line 58
>    at Moq.CastleProxyFactory.Interceptor.Intercept(IInvocation invocation) in C:\projects\moq4\Source\ProxyFactories\CastleProxyFactory.cs:line 176
>    at Castle.DynamicProxy.AbstractInvocation.Proceed()
>    at Castle.Proxies.BarProxy.get_Foo()
>    at ConsoleApplication1.Tests.Test() in C:\dev\ConsoleApplication1\ConsoleApplication1\Program.cs:line 30
> Result Message:	System.NotImplementedException : This is a DynamicProxy2 error: The interceptor attempted to 'Proceed' for method 'System.String get_Foo()' which has no target. When calling method without target there is no implementation to 'proceed' to and it is the responsibility of the interceptor to mimic the implementation (set return value, out arguments etc)

This is a close match to issue #128 but it's not the same issue -- the reproducing case posted there does not reproduce in 4.8.1.

Similar to what is described in #128, my code above works up through 4.2.1402, and throws the exception starting in 4.2.1408. Unlike the earlier issue which was apparently fixed in https://github.com/moq/moq4/releases/tag/v4.2.1502.911 my code continues throwing exceptions up through 4.7.145 (I did not check every release, but spot checked several in the range between 4.2.1408 and 4.8). It works in 4.8.0. It throws again in 4.8.1.

Uncommenting the explicit `Mock.Get(foo).SetupGet(obj => obj.Foo).Returns("foo");` before acting on the mock makes it work in 4.8.1, but that wasn't necessary in 4.8.0, or before 4.2.1408.

stakx:
This fixes #582. (See commit messages for details.)

### DIFF
--- a/Source/Interception/InterceptionAspects.cs
+++ b/Source/Interception/InterceptionAspects.cs
@@ -345,11 +345,23 @@ namespace Moq
 						Debug.Assert(mock.ImplementsInterface(declaringType));
 
 						// Case 2: Explicitly implemented interface method of a class proxy.
-						// Only call base method if it isn't an event accessor.
-						if (!method.LooksLikeEventAttach() && !method.LooksLikeEventDetach())
+
+						if (mock.InheritedInterfaces.Contains(declaringType))
 						{
-							invocation.ReturnBase();
-							return InterceptionAction.Stop;
+							// Case 2a: Re-implemented interface.
+							// The base class has its own implementation. Only call base method if it isn't an event accessor.
+							if (!method.LooksLikeEventAttach() && !method.LooksLikeEventDetach())
+							{
+								invocation.ReturnBase();
+								return InterceptionAction.Stop;
+							}
+						}
+						else
+						{
+							Debug.Assert(mock.AdditionalInterfaces.Contains(declaringType));
+
+							// Case 2b: Additional interface.
+							// There is no base method to call, so fall through.
 						}
 					}
 				}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 #### Fixed
 
 * Wrong parameters count for extension methods in `Callback` and `Returns` (@Caraul, #575)
+* `CallBase` regression with members of additional interfaces (@stakx, #583)
 
 
 ## 4.8.1 (2018-01-08)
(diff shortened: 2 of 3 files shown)


Source/Interception/InterceptionAspects.cs
```diff
@@ -345,11 +345,23 @@ public override InterceptionAction Handle(Invocation invocation, Mock mock)
 						Debug.Assert(mock.ImplementsInterface(declaringType));
 
 						// Case 2: Explicitly implemented interface method of a class proxy.
-						// Only call base method if it isn't an event accessor.
-						if (!method.LooksLikeEventAttach() && !method.LooksLikeEventDetach())
+
+						if (mock.InheritedInterfaces.Contains(declaringType))
 						{
```
stakx: It would be nice to add a `Debug.Assert(!method.IsAbstract)` here someday.

PR #594, opened 2018-02-26
Added `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results

### LINKED ISSUE #592: Usage of ReturnsExtensions.ThrowsAsync() can cause UnobservedTaskException
Usage of ReturnsExtensions.ThrowsAsync() will cause UnobservedTaskException if:
1) Set up a mock method with **ReturnsExtensions.ThrowsAsync()**
```
            var mock = new Mock<A>();
            mock.Setup(a => a.Foo()).ThrowsAsync(new ArgumentException());
```
2) **Don't** call the method you set up

Scenario where it can cause troubles is when you want to check that your code doesn't throw any UnobservedExceptions. It will looks like this 
Sample project [MoqThrowsAsync.zip](https://github.com/moq/moq4/files/1754744/MoqThrowsAsync.zip) or code:
```
    [TestClass]
    public class UnitTest1
    {
        private UnobservedTaskExceptionEventArgs _unobservedEventArgs;

        [TestInitialize]
        public void Init()
        {
            TaskScheduler.UnobservedTaskException += OnUnobservedTaskException;
        }

        private void OnUnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
        {
            _unobservedEventArgs = e;
        }

        [TestCleanup]
        public void Clenup()
        {
            GC.Collect();
            GC.WaitForPendingFinalizers();
            TaskScheduler.UnobservedTaskException -= OnUnobservedTaskException;
            if (_unobservedEventArgs != null && !_unobservedEventArgs.Observed)
            {
                throw _unobservedEventArgs.Exception;
            }
        }

        [TestMethod]
        public void TestMethod1()
        {
            var mock = new Mock<A>();
            mock.Setup(a => a.Foo()).ThrowsAsync(new ArgumentException());
        }
    }

    public interface A
    {
        Task Foo();
    }
```

snrnats:
Added new method overloads:
```
public interface ISetupSequentialResult<TResult>
{
    ...
    ISetupSequentialResult<TResult> Returns(Delegate valueFunction);

    ISetupSequentialResult<TResult> Returns(Func<TResult> valueExpression);
    ...
}
```
 They are needed to resolve #592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Added
+
+* Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
+
 ## 4.8.2 (2018-02-23)
 
 #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
 
+
 ## 4.8.2 (2018-02-23)
 
 #### Changed
--- a/Moq.Tests/SequenceExtensionsFixture.cs
+++ b/Moq.Tests/SequenceExtensionsFixture.cs
@@ -14,10 +14,12 @@ namespace Moq.Tests
 			mock.SetupSequence(x => x.Do())
 				.Returns(2)
 				.Returns(3)
+				.Returns(() => 4)
 				.Throws<InvalidOperationException>();
 
 			Assert.Equal(2, mock.Object.Do());
 			Assert.Equal(3, mock.Object.Do());
+			Assert.Equal(4, mock.Object.Do());
 			Assert.Throws<InvalidOperationException>(() => mock.Object.Do());
 		}
 
--- a/Moq.Tests/SequenceExtensionsFixture.cs
+++ b/Moq.Tests/SequenceExtensionsFixture.cs
@@ -155,6 +155,30 @@ namespace Moq.Tests
 			mock.Verify(m => m.Do(), Times.Exactly(2));
 		}
 
+		[Fact]
+		public void Func_are_invoked_deferred()
+		{
+			var mock = new Mock<IFoo>();
+			var i = 0;
+			mock.SetupSequence(m => m.Do())
+				.Returns(() => i);
+
+			i++;
+
+			Assert.Equal(i, mock.Object.Do());
+		}
+
+		[Fact]
+		public void Func_can_be_treated_as_return_value()
+		{
+			var mock = new Mock<IFoo>();
+			Func<int> func = () => 1;
+			mock.SetupSequence(m => m.GetFunc())
+				.Returns(func);
+
+			Assert.Equal(func, mock.Object.GetFunc());
+		}
+
 		public interface IFoo
 		{
 			string Value { get; set; }
@@ -162,6 +186,8 @@ namespace Moq.Tests
 			int Do();
 
 			Task<int> DoAsync();
+
+			Func<int> GetFunc();
 		}
 
 		public class Foo
--- a/Moq.Tests/SequenceExtensionsFixture.cs
+++ b/Moq.Tests/SequenceExtensionsFixture.cs
@@ -179,6 +179,22 @@ namespace Moq.Tests
 			Assert.Equal(func, mock.Object.GetFunc());
 		}
 
+		[Fact]
+		public void Keep_Func_as_return_value_when_setup_method_returns_implicitly_casted_type()
+		{
+			var mock = new Mock<IFoo>();
+			Func<object> funcObj = () => 1;
+			Func<Delegate> funcDel = () => new Action(() => { });
+			Func<MulticastDelegate> funcMulticastDel = () => new Action(() => { });
+			mock.SetupSequence(m => m.GetObj()).Returns(funcObj);
+			mock.SetupSequence(m => m.GetDel()).Returns(funcDel);
+			mock.SetupSequence(m => m.GetMulticastDel()).Returns(funcMulticastDel);
+
+			Assert.Equal(funcObj, mock.Object.GetObj());
+			Assert.Equal(funcDel, mock.Object.GetDel());
+			Assert.Equal(funcMulticastDel, mock.Object.GetMulticastDel());
+		}
+
 		public interface IFoo
 		{
 			string Value { get; set; }
@@ -188,6 +204,12 @@ namespace Moq.Tests
 			Task<int> DoAsync();
 
 			Func<int> GetFunc();
+
+			object GetObj();
+
+			Delegate GetDel();
+
+			MulticastDelegate GetMulticastDel();
 		}
 
 		public class Foo
--- a/Source/Language/Flow/SetupSequencePhrase.cs
+++ b/Source/Language/Flow/SetupSequencePhrase.cs
@@ -96,8 +96,8 @@ namespace Moq.Language.Flow
 			// If `TResult` is `Func<>`, that is someone is setting up the return value of a method
 			// that returns a `Func<>`, then we need to make sure that the passed Func will be treated 
 			// as a return value. We don't want to invoke the passed Func to get a return value; the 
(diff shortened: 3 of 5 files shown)


Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -93,6 +95,38 @@ public ISetupSequentialResult<TResult> Returns(TResult value)
 			this.setup.AddReturns(value);
 			return this;
 		}
+		public ISetupSequentialResult<TResult> Returns(Delegate valueFunction)
```
stakx: Please insert a blank line before this one.

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -40,6 +40,8 @@
 
 using System;
 using System.ComponentModel;
+using System.Reflection;
+using Moq.Properties;
```
stakx: Please insert a blank line before this one (such that namespace imports are grouped by main namespace).

Source/Language/Flow/SetupSequencePhrase.cs
stakx: The parameter type `Delegate` doesn't seem right. It wouldn't make sense to pass *any* kind of delegate to `Returns`; how would Moq know what arguments to pass, and in case anything other than a `TResult` comes back, how to return it? I'd say the parameter type should be `Func<TResult>`. (When I quoted the code that you've copied almost verbatim (below), I only meant to demonstrate that you'd have to keep in mind the possibility that the passed `Func<TResult>` might be the desired result instead of the value returned by the `Func<>`. Like I said in the issue, I'm not sure this is necessary at all, perhaps try to write a test pretending to call `Returns` with a `Func` argument that should be returned as the result instead of being called.

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -93,6 +95,38 @@ public ISetupSequentialResult<TResult> Returns(TResult value)
...
+			}
+
+			// The following may seem overly cautious, but we don't throw an `ArgumentNullException`
+			// here because Moq has been very forgiving with incorrect `Returns` in the past.
+			if (valueFunction == null)
+			{
+				return this.Returns(() => default(TResult));
+			}
+
+			if (valueFunction.GetMethodInfo().ReturnType == typeof(void))
```
stakx: If the parameter type is changed to `Func<TResult>`, then this `if` will become redundant.

Source/Language/ISetupSequentialResult.cs
```diff
@@ -18,6 +18,18 @@ public interface ISetupSequentialResult<TResult>
 		/// </summary>
 		ISetupSequentialResult<TResult> Returns(TResult value);
 
+		/// <summary>
+		/// Uses delegate to get return value
+		/// </summary>
+		/// <returns></returns>
+		ISetupSequentialResult<TResult> Returns(Delegate valueFunction);
```
stakx: As above, I'd say this overload isn't necessary at all, it doesn't make sense to pass any `Delegate`. Moq won't know how to invoke *any* arbitrary delegate.

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -93,6 +95,38 @@ public ISetupSequentialResult<TResult> Returns(TResult value)
...
+			// We don't want to invoke the passed delegate to get a return value; the passed delegate
+			// already is the return value.
+			if (typeof(TResult) == typeof(Delegate))
+			{
+				return this.Returns(() => (TResult)(object)valueFunction);
+			}
+
+			// The following may seem overly cautious, but we don't throw an `ArgumentNullException`
+			// here because Moq has been very forgiving with incorrect `Returns` in the past.
+			if (valueFunction == null)
```
stakx: I think this `if` clause could be replaced with a simple `Guard.NotNull(valueFunction, nameof(valueFunction))`.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Added
+
+* Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
+
 ## 4.8.2 (2018-02-23)
```
stakx: (Detail, but if it's not too much hassle, please insert a second blank line before this one.)

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -93,6 +95,38 @@ public ISetupSequentialResult<TResult> Returns(TResult value)
...
+			{
+				throw new ArgumentException(Resources.InvalidReturnsCallbackNotADelegateWithReturnType, nameof(valueFunction));
+			}
+
+			this.setup.AddReturns(valueFunction);
+			return this;
+		}
+
+		public ISetupSequentialResult<TResult> Returns(Func<TResult> valueExpression)
+		{
```
stakx: Perhaps add a `Guard.NotNull` (argument validation) here.

stakx: Perhaps add a `Guard.NotNull` (argument validation) here.

stakx: Like I mentioned in the issue, I'm not sure if it's 100 % safe to add a new overload to that method. What if there's user code out there that passes a `Func` to the previous only method with the intent that the `Func` itself is the result? If we now add a method overload that takes a `Func`, the compiler might suddenly pick that new overload instead of the original function, causing the sequence setup to produce a different value (the result of the function instead of the function itself). Perhaps try to write a test that provokes this situation.

Source/Language/ISetupSequentialResult.cs
stakx: As above, I don't think this overload is necessary. I'd say remove it.

Source/Language/Flow/SetupSequencePhrase.cs
stakx: :+1:

Moq.Tests/SequenceExtensionsFixture.cs
```diff
@@ -153,13 +155,39 @@ public void Verify_can_verify_invocation_count_for_sequences()
...
+
 		public interface IFoo
 		{
 			string Value { get; set; }
 
 			int Do();
 
 			Task<int> DoAsync();
+
+			Func<int> GetFunc();
```
stakx: Suggestion: Add another test for a method with return type `object`. This might be relevant because when `TResult := object`, both overloads `Returns(object)` and `Returns(Func<object>)` become relevant. (See also long review comment below.)

Source/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -90,10 +93,33 @@ public ISetupSequentialResult<TResult> CallBase()
 
 		public ISetupSequentialResult<TResult> Returns(TResult value)
 		{
+			// If `TResult` is `Func<>`, that is someone is setting up the return value of a method
+			// that returns a `Func<>`, then we need to make sure that the passed Func will be treated 
+			// as a return value. We don't want to invoke the passed Func to get a return value; the 
+			// passed func already is the return value. To prevent it we need to wrap up this func because 
+			// `SequenceMethodCall.Execute` invokes every func it encounters
+			if (typeof(TResult).IsConstructedGenericType && typeof(TResult).GetGenericTypeDefinition() == typeof(Func<>))
```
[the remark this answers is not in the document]
snrnats: In the latest changes I tried to cover slightly different case of breaking changes:

After my changes `SequenceMethodCall.Execute()` treats all `Delegate` as a function that need to be evaluated. Because of these an old code that mocked method like `Func<AnyType> GetFunc()` would be broken. So it's why I have a check inside `Returns(TResult)`. Although my checks didn't cover all cases.

Now I am considering that it's better to have a new enum value `SequenceMethodCall.ResponseKind.InvokeFunc` for all funcs that need to be invoked to get the return value. It will allow to remove checks that I was trying to accomplish.

Anyway I am still missing the scenario that you described. If I understand it right we also need to handle some other cases:
```
                public interface IFoo
 		{ 
 			object GetObj();

			Delegate GetDel();

                        MulticastDelegate GetMulticastDel()
                 }
```

Moq.Tests/SequenceExtensionsFixture.cs
stakx: :+1:

stakx: :+1:

PR #595, opened 2018-02-27
 Fixed UnobservedTaskException after using ThrowsAsync. Fixed ReturnsAsync and ThrowsAsync with delay

### LINKED ISSUE #592: Usage of ReturnsExtensions.ThrowsAsync() can cause UnobservedTaskException

### LINKED ISSUE #593: ReturnsAsync and ThrowsAsync with delay parameter starts timer at setup
`ReturnsAsync` and `ThrowsAsync` **with delay parameter** start timer at setup, not at mock method invocation.

The reason is that `Task.Delay` is called inside `DelayedResult/DelayedException` but it needs to be deferred until mock method is invoked.
```
public class Issue593
{
	[Fact]
	public void ReturnsAsync_ThrowsAsync_start_delay_timer_at_mock_invocation()
	{
		var mock = new Mock<IFoo>();
		mock.Setup(a => a.Boo()).ReturnsAsync(true, TimeSpan.FromMilliseconds(100));
		mock.Setup(a => a.Foo()).ThrowsAsync(new ArgumentException(), TimeSpan.FromMilliseconds(100));

		//Wait for the delay greater then specified for Boo setup
		Task.Delay(200).GetAwaiter().GetResult();

		Assert.False(mock.Object.Boo().IsCompleted);
		Assert.False(mock.Object.Foo().IsCompleted);
	}

	public interface IFoo
	{
		Task<bool> Foo();
		Task<bool> Boo();
	}
}
```

snrnats:
This closes #592, closes #593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
 
+#### Fixed
+
+*  Usage of ReturnsExtensions.ThrowsAsync() can cause UnobservedTaskException (@snrnats, #595)
+*  ReturnsAsync and ThrowsAsync with delay parameter starts timer at setup (@snrnats, #595)
+
 
 ## 4.8.2 (2018-02-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
 
+#### Changed
+
+* **Breaking change:** `sequenceSetup.ReturnsAsync(...)`, `sequenceSetup.ThrowsAsync(...)`, `setup.ReturnsAsync(..., TimeSpan)`, `setup.ReturnsAsync(..., TimeSpan, TimeSpan)`, `setup.ReturnsAsync(..., TimeSpan, TimeSpan, Random)` and all overloads of `setup.ThrowsAsync(...)` now return a different `Task` on each invocation, which wasn't previously the case (@snrnats, #595) 
+
 #### Fixed
 
 *  Usage of ReturnsExtensions.ThrowsAsync() can cause UnobservedTaskException (@snrnats, #595)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Changed
 
-* **Breaking change:** `sequenceSetup.ReturnsAsync(...)`, `sequenceSetup.ThrowsAsync(...)`, `setup.ReturnsAsync(..., TimeSpan)`, `setup.ReturnsAsync(..., TimeSpan, TimeSpan)`, `setup.ReturnsAsync(..., TimeSpan, TimeSpan, Random)` and all overloads of `setup.ThrowsAsync(...)` now return a different `Task` on each invocation, which wasn't previously the case (@snrnats, #595) 
+* **Breaking change:** All ReturnsAsync and ThrowsAsync setup methods now consistently return a new Task on each invocation (@snrnats, #595) 
 
 #### Fixed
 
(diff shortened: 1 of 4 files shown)


snrnats:
@stakx I have have a concern about changed behavior that we have discussed. We chose to keep the behavior that `ReturnsAsync` and `ThrowsAsync` keeps returning the same task because it's how it worked before. But it's not possible with delayed `ReturnsAsync` or 'ThrowsAsync'. New task need to be created to start the timer over. Would it be better to implement a new consistent behavior for all `ReturnsAsync` and `ThrowsAsync` where each invocation will return a new task?

stakx:
@snrnats: Thanks for this new PR. Thank you also for raising above question.

> Would it be better to implement a new consistent behavior for all ReturnsAsync and ThrowsAsync where each invocation will return a new task?

Yes, after thinking about this a little longer, I believe this would be the better solution; for two reasons:

1. The majority of `ReturnsAsync` and `ThrowsAsync` methods already seem to cause the setup methods to return new `Task`s for each invocation. (I haven't finished verifying this, you're welcome to verify this claim independently.)

2. When you manually implement a method returning some `Task<TResult>`, you'll also end up with an implementation that returns a new `Task` for each invocation *unless* you make a deliberate effort to the contrary:

```csharp
Task<int> IFoo.GetSomeNumberAsync()  // returns a new Task each time
{
    return Task.FromResult(42);
}

Task<int> IFoo.GetSomeNumberAsync()  // returns a new Task each time
{
    var tcs = new TaskCompletionSource<int>();
    tcs.SetException(new Exception("..."));
    return tcs.Task;
}

async Task<int> IFoo.GetSomeNumberAsync()  // returns a new Task each time
{
    await Task.Delay(0);
    return 42;
}

async Task<int> IFoo.GetSomeNumberAsync()  // returns a new Task each time
{
    await Task.Delay(0);
    throw new Exception("...");
}
```

To return the same `Task` on each invocation, you'd have to hack together something more complicated, e.g. this:

```csharp
private Lazy<Task<int>> cachedTask = new Lazy<Task<int>>(() =>
{
    return Task.FromResult(42);
});

Task<int> IFoo.GetSomeNumberAsync() => cachedTask.Value;
```

And I dare say this isn't exactly what most people will do in practice, so why should Moq?

And a third reason specific to `SequenceExtensions.ThrowsAsync`: Come to think of it, it doesn't even matter whether the same or a new `Task` is returned upon every invocation, because with a sequence, one `ThrowsAsync` configuration will only affect one single invocation; then the sequence moves on to the next configured step. So I was wrong in originally suggesting the `Lazy` business, it isn't required.

So I think I'd be OK with a breaking change here, as long as this is explicitly marked in the changelog entry, e.g.:

```
#### Changed

* **Breaking change:** `setup.ReturnsAsync(..., TimeSpan)` and `setup.ThrowsAsync(..., TimeSpan)` now return a different `Task` on each invocation, which wasn't previously the case (@snrnats, #nnn) 
```

So you'd have to find out exactly which methods are affected, and mention them all in the changelog.

snrnats:
I wanted to make two separate PRs. But I wasn't comfortable changing the code when I know that it still has issues:smile:

stakx:
@snrnats - I just pushed Moq 4.8.3 to NuGet, which includes your changes. Sorry it's taken so long.

PR #596, opened 2018-02-27
Add convenience method ReturnsSequence

Kralizek:
A simple convenience method to simply setup so that a sequence of items is returned, once per call.

### DIFF
--- a/Source/SequenceExtensions.cs
+++ b/Source/SequenceExtensions.cs
@@ -73,5 +73,18 @@ namespace Moq
 
 			return setup.Returns(tcs.Task);
 		}
+		
+		/// <summary>
+		/// Returns a sequence of items, once per call.
+		/// </summary>
+		public static ISetupSequentialResult<T> ReturnsSequence<T>(this ISetupSequentialResult<T> setup, IEnumerable<T> results)
+        {
+            foreach (var item in results)
+            {
+                setup.Returns(item);
+            }
+
+            return setup;
+        }
 	}
 }
--- a/Source/SequenceExtensions.cs
+++ b/Source/SequenceExtensions.cs
@@ -39,6 +39,7 @@
 // http://www.opensource.org/licenses/bsd-license.php]
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Threading.Tasks;
 


stakx:
Hi @Kralizek, thanks for working on this!

Unfortunately, we are currently no longer accepting any pull requests that effect an API change in Moq 4 due to Moq 5's imminent initial release&mdash;see [this message by @kzu (project owner) in Moq's Gitter chat](https://gitter.im/moq/moq?at=5a883d726fba1a703a62cdf0). The idea is that kzu can finalize Moq 5 without having to hit a moving target.

I'll leave this PR open for now, perhaps we can merge this at a later time. I'll ask kzu how to proceed with this and other similar PRs at an appropriate time (sometime during the next few weeks).

Kralizek:
@stakx thanks for your message.

do you mind if I attach this other variation to the PR?

```csharp
public static ISetupSequentialResult<Task<T>> ReturnsSequenceAsync<T>(this ISetupSequentialResult<Task<T>> setup, IEnumerable<T> results)
{
    foreach (var item in results)
    {
        setup.ReturnsAsync(item);
    }

    return setup;
}
```

stakx:
I don't mind, but I'll withhold any in-depth review and decision until I know that Moq 4 API extensions are OK to be merged.

What I can tell you now is that there's one thing I believe could be solved better: Your convenience method tries to eagerly consume the whole sequence and convert each item to a correspong sequence setup step. This approach obviously won't work for infinite sequences, and it'll have a suboptimal memory footprint for large sequences because they'll be fully buffered in memory.

It would perhapd be better to implement this functionality inside `SequenceMethodCall` so that the original sequence is consumed on demand, one element at a time. This would add support for infinite sequences and drastically reduce memory footprint (you'd have to buffer only one element instead of all of them). Thread safety would become a concern, however.

That's all for now. Please be patient. I'll try to get back to you as soon as I can.

stakx:
P.S.: I tried implementing this method in a memory-efficient and thread-safe manner out of curiosity what would be involved in doing so. This is the draft code that I ended up with: https://github.com/stakx/moq4/commit/4fe3e7cc473d14afc6b493320309e57c0fda2f9a. It appears to work, but I haven't written any proper tests, nor have I measured how much runtime performance is affected by replacing `ConcurrentDictionary` with `lock`s.

Kralizek:
Oh wow, I feel almost ashamed of my naive implementation... 😲

Kralizek:
Would make it easier to move from ‘IEnumerable<T>‘ to ’IReadOnlyList<T>’?

Obviously the memory consumption should be taken in check but at least we can strike out the risk of infinite sequences.

stakx:
@Kralizek 

> I feel almost ashamed of my naive implementation.

Please don't. Your implementation would make for a perfectly good extension method e. g. in an external contrib package where you cannot change Moq's internals, i. e. where you are constrained to use only the public API.

> Would make it easier to move from `IEnumerable` to `IReadOnlyList`?
>
> Obviously the memory consumption should be taken in check but at least we can strike out the risk of infinite sequences.

No, `IEnumerable` is actually the right abstraction: you'll only ever look at one element after the other; the sequence setup doesn't ever have to know how many elements there are in total, nor will it ever need to perform random accesses.

Infinite sequences aren't a problem as long as you don't try to buffer the whole sequence in memory.

stakx:
@Kralizek, are you still interested in working on this?

* If not, I suggest we close this issue.

* If yes, I think it would be good to iterate on your current code some more, in particular to address the memory efficiency (and thread-safety) concerns that I raised above.

How would you like to proceed?

Kralizek:
Hi @stakx, I thought this PR was on freeze until v5 is out. Has anything changed? 

Anyway I'm off for vacation soon and I cannot guarantee much coding since I'll be abroad for the whole time.

stakx:
This PR has gone stale (no activity for half a year or more), shall I close it?

stakx:
@Kralizek - Thanks for replying. If you're currently too busy to work on this, I suggest we close this PR. If and when you're ready to resume work on this addition (either here, on over at the Moq v5 repository), you can always submit a new PR.

Issue #388, opened 2017-06-21
Add benchmarks


JohanLarsson:
As it is now it is not uncommon for moq to show up in the profiler. It will probably never be a fast library with all the use of expressions but it can be good to track performance as it is pretty relevant when used in tests.

[nodatime](https://github.com/nodatime/nodatime) has a nice setup I think. Infrastructure for running them is perhaps tricky, not sure running them on CI is worth much.

JohanLarsson:
Stuff like this is not helping perf:
```c#
public static void NotNull<T>(Expression<Func<T>> reference, T value)
{
	if (value == null)
	{
		throw new ArgumentNullException(GetParameterName(reference));
	}
}
```

Using C#7 it can be replaced by:

```c#
this.underlyingCreateMocks = underlyingCreateMocks ?? throw new ArgumentNullException(nameof(underlyingCreateMocks));
```

stakx:
(Yep. Now that we have `nameof`, the whole `Guard` helper class could be rewritten to not use expression trees just to get at the parameter name.)

stakx:
The idea of setting up some kind of strategy for measuring the effects of optimizations is a good one, btw.!

Speaking about priorities, I think right now it's most important to first deal with extant bug reports and long-standing feature requests. Next, there are some convoluted bits of code that need to be rewritten to make their intent more clear. Finally, optimizations. (["Make it work, make it right, make it fast."](http://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast))

jatouma:
Tried updating from Moq 4.2.1402 to Moq 4.99 hoping for a performance improvement, but our unit tests now take 31 minutes, up from 17 minutes.

stakx:
Yep, Moq has lately suffered performance-wise. I haven't verified this, but I am suspecting these three main cuases:

* There have been improvements regarding thread safety, which was achieved through `lock`s (aka the `Monitor` class). Those are fairly expensive. There might be some room for improvement by switching to concurrent collections, possibly at the expense of higher memory consumption.

* Moq does a lot of expression tree lambda compilation, which is expensive under x64. Not sure if it did less of that, though.

* There's an expensive feature for better debugging, namely capturing call stack information (file & line number) for all setup sites. Perhaps this feature wasn't yet in 4.2.

I'd very much love to get to performance optimizations, but before we can really start on those, we need to refactor a lot of code such that e.g. we can make a shift away from expression tree lambda compilation towards expression tree inspection only.

[...]

PR #598, opened 2018-02-28
Replace Mock.Of code with 10x faster equivalent

informatorius:
https://github.com/moq/moq4/issues/547

PR not finished yet ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Changed
 
-* Changed Mock.Of<T> implementation to use much faster simple approach
+* Changed Mock.Of<T> implementation to use much faster simple approach (@informatorius, #547)
 
 ## 4.8.2 (2018-02-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
 
+#### Changed
+
+* Changed Mock.Of<T> implementation to use much faster simple approach
 
 ## 4.8.2 (2018-02-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Changed
 
-* Changed Mock.Of<T> implementation to use much faster simple approach (@informatorius, #547)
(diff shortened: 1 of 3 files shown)


stakx:
https://github.com/moq/moq4/blob/9c618125989d40f19bd8afa07ee21510b875e8d9/Source/Linq/Mock.cs#L53-L61

;-)

informatorius:
Fixing ...
What about the benchmarks you did, should I include them somehow?
Or is this done with https://github.com/moq/moq4/issues/388

stakx:
The time required to create a mock depends on the complexity of the type you want to mock. So including one single benchmark for one single mocked type wouldn't really give the full picture. All I wanted to see with that benchmark is an approximate speed difference (order of magnitude).

I'll review your PR a little later!

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
 
+#### Changed
+
+* Changed `Mock.Of<T>()` implementation to use much faster simple approach (@informatorius, #547)
```
stakx: In principle, this is a perfectly fine changelog entry. But may I humbly suggest that you shorten & rephrase it to be even more relevant from the user's perspective? For example:

> * Speed up `Mock.Of<T>()` by approx. one order of magnitude (...)

Source/Linq/Mock.cs
```diff
@@ -57,7 +57,12 @@ public partial class Mock
 		/// <returns>The mocked object created.</returns>
 		public static T Of<T>() where T : class
 		{
-			return Mocks.CreateMockQuery<T>().First<T>();
+			//Disabled because of bad performance
+			//return Mocks.CreateMockQuery<T>().First<T>();
+			//and replaced by equivalent which is more than 10 times faster currently
```
stakx: Hmm, IMHO using the word "bad" isn't quite fair. I totally agree that the current implementation has *worse* performance than your new one; but does that make it *bad*?

I would've chosen a slightly different wording, for example:

```csharp
// This method was originally implemented as follows:
//
//     return Mocks.CreateMockQuery<T>().First<T>();
//
// which involved a lot of (unnecessary) `IQueryable` query provider overhead and
// lambda compilation. What it really boils down to is this (much faster) code:
...
```

CHANGELOG.md
stakx: Please add one more blank line after this one.

stakx: Oh btw., please replace `#547` (issue) with `#598` (this PR).

informatorius:
Done! Thanks for your perfect feedback.

stakx:
Thank you for contributing this much faster reimplementation of `Mock.Of<T>()`! :+1:

stakx:
@informatorius - I just pushed Moq 4.8.3 to NuGet, which includes your changes. Sorry it's taken so long.

PR #602, opened 2018-03-06
'Returns' regression with null function callback

Caraul:
Restore 4.7 behavior, add tests and update CHANGELOG.md

### DIFF
--- a/Source/MethodCallReturn.cs
+++ b/Source/MethodCallReturn.cs
@@ -100,14 +100,7 @@ namespace Moq
 				return this.Returns(() => (TResult)(object)valueFunction);
 			}
 
-			// The following may seem overly cautious, but we don't throw an `ArgumentNullException`
-			// here because Moq has been very forgiving with incorrect `Returns` in the past.
-			if (valueFunction == null)
-			{
-				return this.Returns(() => default(TResult));
-			}
-
-			if (valueFunction.GetMethodInfo().ReturnType == typeof(void))
+			if (valueFunction != null && valueFunction.GetMethodInfo().ReturnType == typeof(void))
 			{
 				throw new ArgumentException(Resources.InvalidReturnsCallbackNotADelegateWithReturnType, nameof(valueFunction));
 			}
@@ -154,12 +147,18 @@ namespace Moq
 
 		private void SetReturnDelegate(Delegate value)
 		{
-			if (value != null)
+			if (value == null)
+			{
+				// The following may seem overly cautious, but we don't throw an `ArgumentNullException`
+				// here because Moq has been very forgiving with incorrect `Returns` in the past.
+				this.valueDel = (Func<TResult>)(() => default(TResult));
+			}
+			else
 			{
 				ValidateReturnDelegate(value);
+				this.valueDel = value;
 			}
 
-			this.valueDel = value;
 			this.returnValueKind = ReturnValueKind.Explicit;
 		}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 *  Usage of `ReturnsExtensions.ThrowsAsync()` can cause `UnobservedTaskException` (@snrnats, #595)
 *  `ReturnsAsync` and `ThrowsAsync` with delay parameter starts timer at setup (@snrnats, #595)
+*  `Returns` regression with null function callback (@Caraul, #600)
 
 
 ## 4.8.2 (2018-02-23)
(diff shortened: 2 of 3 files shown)


Source/MethodCallReturn.cs
```diff
@@ -154,12 +147,18 @@ public IReturnsResult<TMock> CallBase()
 
 		private void SetReturnDelegate(Delegate value)
 		{
-			if (value != null)
+			if (value == null)
+			{
+				// The following may seem overly cautious, but we don't throw an `ArgumentNullException`
+				// here because Moq has been very forgiving with incorrect `Returns` in the past.
```
stakx: Let's rewrite these two comment lines so it becomes clear that we're not just being "overly cautious", but that this is in fact a required piece of code. For example:

```csharp
// A `null` reference (instead of a valid delegate) is interpreted as the actual return value.
// This is necessary because the compiler might have picked the wrong overload for calls like
// `Returns(null)`, and instead of in `Returns(TResult)`, we ended up in `Returns(Delegate)` or
// `Returns(Func<TResult>)`, which likely isn't what the user intended. So here we do what
// we would've done in `Returns(TResult)`:
```

(We're essentially dealing with the same problem&mdash;C# picking the "wrong" overload&mdash;as already mentioned here:)

https://github.com/moq/moq4/blob/f86759225c46b072c4a721b200b31992d4c54b8b/Source/MethodCallReturn.cs#L92-L101

Caraul: Well, the comment is subject to discuss :)

I've already mentioned "`Returns(null)` doesn't compile due to ambiguous invocation" so there is no wrong pick - anyway the user should clearly choose one of Returns overloads with typecasting or "closing" generics etc. Rock on which we split is `Returns<TResult>(null)` which seems like `Returns((TResult)null)` but in fact it's `IReturns.Returns<T>((Func<T, TResult>)null)` - surprisingly, isn't it? So I'd propose sightly modified version:
```
// A `null` reference (instead of a valid delegate) is interpreted as the actual return value.
// This is necessary because the user might have picked an ambiguous overload like `Returns<TResult>(null)`,
// but instead of in `Returns(TResult)`, we ended up in `Returns<T>(Func<T, TResult>)`,
// which likely isn't what the user intended. So here we do what we would've done in `Returns(TResult)`
```
What do you think?

stakx: > "`Returns(null)` doesn't compile due to ambiguous invocation" so there is no wrong pick

`Returns(null)` compiles just fine in some quick scenarios that I've tried, so I can't really confirm this.

> anyway the user should clearly choose one of Returns overloads with typecasting or "closing" generics etc.

That may be so, but we have no control over whether they think so, too, nor whether users will actually do so. If everyone explicitly chose the right overload, then passing `null` instead of a delegate should trigger an `ArgumentNullException`. But let's stay on the safe side and assume that some people let the compiler pick the right overload. And because that can go wrong, we need to "correct" those situations.

Regarding your new comment, that's principally also fine with me. But let's be clear that to the compiler, the overload isn't ambiguous at all. It's simply that users might expect a different overload to be called than what the compiler will choose.

Caraul: > Returns(null) compiles just fine in some quick scenarios that I've tried, so I can't really confirm this.

The truth appears to be in between - it's ambiguous for references but for value types compiler chooses `Returns(Func<TResult>)`.

> users might expect a different overload

My point exactly. The only note that reasons are "compiler chooses unexpected overload" and "user chooses an overload that behaves unexpectedly". I'd also try to avoid the word "wrong" as it may be incorrectly interpreted as "bug in compiler". So what about version:

```
// A `null` reference (instead of a valid delegate) is interpreted as the actual return value.
// This is necessary because the compiler might have picked the unexpected overload for calls like
// `Returns(null)`, or the user might have picked an overload like `Returns<TResult>(null)`,
// and instead of in `Returns(TResult)`, we ended up in `Returns(Delegate)` or
// `Returns(Func<TResult>)`, which likely isn't what the user intended. So here we do what
// we would've done in `Returns(TResult)`:
```

stakx: Agreed, avoiding the word `wrong` is a good idea.

Fine with me, too, except that when the user picks a specific overload e.g. by explicitly specifying the generic type argument, then IIRC we cannot end up in any other overloads. So I'd remove the part that says `or the user might ...`.

I really don't mean to be too fussy about the comment, main thing is that we're more specific than my original `overly cautious` comment, which clearly no longer applies.

stakx:
... and merged! Thank you for fixing this regression!

stakx:
@Caraul - I just pushed Moq 4.8.3 to NuGet, which includes your changes. Sorry it's taken so long.

Caraul:
Great job, thank you very much!

2018-06-09 22:59 GMT+03:00 stakx <notifications@github.com>:

PR #625, opened 2018-06-09
Verify that `in` parameter modifier is supported

### LINKED ISSUE #605: C# 7.2 `in ValueType` parameters are not supported with .NET Standard / Core
See this discussion over at DynamicProxy's repo: https://github.com/castleproject/Core/issues/339#issuecomment-372705522.

What needs to be done to add C# 7.2 reference semantics support in Moq for value types is this:

* [x] Upgrade Castle.Core package dependency to vNext once it's available (see https://github.com/castleproject/Core/issues/354).
* [X] ~~Add .NET Standard 2.0 as a target framework.~~ Not required, see below.
* [X] ~~Possibly update delegate generator logic, as per https://github.com/moq/moq4/issues/605#issuecomment-373309955.~~ Not required, see below.
* [X] Check whether `DynamicInvoke` e.g. for `Callback` and `Returns` is already lenient enough given differences in delegate signatures (presence of `in` parameter modifier), as per https://github.com/moq/moq4/issues/605#issuecomment-373343104

stakx:
This adds a few tests, including those mentioned in https://github.com/castleproject/Core/issues/339#issue-304443306 and https://github.com/castleproject/Core/issues/339#issuecomment-372605885, to document the state of `in` parameter modifier support.

As it turns out, the previous update of Castle.Core to 4.3.0 makes `in` parameter modifiers work on .NET Core even though Moq doesn't specifically target .NET Standard 1.5 or above... it suffices that Castle.Core does that.

(.NET Standard 1.3 vs 1.5 is important because the latter includes the System.Reflection.Emit facilities required for mocking `in`; more specifically, the ability to emit custom modifiers.)

This closes #605.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 #### Added
 
 * Add `ISetupSequentialResult<TResult>.Returns` method overload that support delegate for deferred results (@snrnats, #594)
+* Support for C# 7.2's `in` parameter modifier (@stakx, #624, #625)
 
 #### Changed
 
--- a/Moq.Tests/Moq.Tests.csproj
+++ b/Moq.Tests/Moq.Tests.csproj
@@ -9,6 +9,7 @@
 		<GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
 		<PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
 		<SignAssembly>true</SignAssembly>
+		<LangVersion>7.2</LangVersion>
 	</PropertyGroup>
 
 	<PropertyGroup Condition=" '$(TargetFramework)' == 'net46' ">
--- a/Moq.Tests/Regressions/IssueReportsFixture.cs
+++ b/Moq.Tests/Regressions/IssueReportsFixture.cs
@@ -1906,6 +1906,71 @@ namespace Moq.Tests.Regressions
 
 		#endregion
 
+		#region 605
+
+		public class Issue605
+		{
+			// Taken from https://github.com/castleproject/Core/issues/339.
+
+			public readonly struct Struct
+			{
+			}
+
+			public interface IStructByRefConsumer
+			{
+				void Consume(in Struct message);
+			}
+
+			public interface IGenericStructByRefConsumer<T>
+			{
+				T Consume(in Struct message);
+			}
+
+			public interface IStructByValueConsumer
+			{
+				void Consume(Struct message);
+			}
+
+			[Fact]
+			public void Struct_ByRef_Moq_Test()
+			{
+				_ = Mock.Of<IStructByRefConsumer>();
+			}
+
+			[Fact(Skip = "Currently not supported due to a bug in the runtime, see https://github.com/dotnet/corefx/issues/29254.")]
+			public void Struct_ByRef_Generic_Moq_Test()
+			{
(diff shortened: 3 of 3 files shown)


zvirja:
@stakx And what about delegates with `in` modifiers? Do you also support them? We have the same situation in NSubstitute and I'm guessing that'll make us to target 1.5 directly..

zvirja:
@stakx Oh, disregard - you have a test for that. That's really interesting - will investigate later how that works, given that we emit own interface...

stakx:
@zvirja - In Moq, think the important bits occur here:

https://github.com/moq/moq4/blob/396a9e5f2196ee20a7df0a74b0495db2835af943/Source/ProxyFactories/CastleProxyFactory.cs#L143-L149

Note that in the call to `DefineMethod`, the original delegate's `ParameterInfo`s are supplied. `ParameterInfo` carries with it the means to query custom attributes as well as custom modifiers, so (I guess) System.Reflection.Emit is smart enough to copy those.

https://github.com/moq/moq4/blob/396a9e5f2196ee20a7df0a74b0495db2835af943/Source/ProxyFactories/CastleProxyFactory.cs#L151-L154

I'm not actually sure the above loop is needed at all&mdash;quite possibly it does little more than copy the parameter names (and perhaps even that has already happened).

zvirja:
> Note that in the call to DefineMethod, the original delegate's ParameterInfos are supplied. 

I dont see that. You pass parameter types, but they don't carry modifiers..

I suspect that it happens because runtime allow delegate and method to differ by `in` modifier for parameters, so it doesn't matter that generated method is without it. But it's just a guess.

stakx:
@zvirja - Oh boy, you're right. Now I feel like a fool. 🦆😆 The tests I've already got there might be inadequate then. I've done some more REPL-style testing, Moq indeed doesn't reproduce the delegate method's signature accurately, yet invoking the mocked delegate still works. If, during your own investigations, you find a way of making mocked delegate invocation fail, I'd love to hear what your tests look like!

stakx:
@zvirja - I found out why delegate mocking works even when custom modifiers aren't replicated. It seems that for delegates, the CLR performs a signature check that's less strict than usual:

```csharp
public delegate void WithIn(in AttributeTargets arg);

public interface WithoutIn
{
	void Invoke(ref AttributeTargets arg);
}

public class Impl : WithoutIn
{
	public void Invoke(ref AttributeTargets arg)
	{
		Console.WriteLine($"Called with {arg}.");
	}
}

class Program
{
	static void Main(string[] args)
	{
		// The following will print 1 and 0, meaning that the metho signatures differ!
		Console.WriteLine(typeof(WithIn).GetMethod("Invoke").GetParameters()[0].GetRequiredCustomModifiers().Length);
		Console.WriteLine(typeof(WithoutIn).GetMethod("Invoke").GetParameters()[0].GetRequiredCustomModifiers().Length);

		// Despite the difference in signature, the methods are compatible when
		// going through a delegate invocation:
		var interfaceMethod = typeof(WithoutIn).GetMethod("Invoke");
		var deleg = (WithIn)interfaceMethod.CreateDelegate(typeof(WithIn), new Impl());
		deleg.Invoke(AttributeTargets.All);
	}
}
```

I might raise an issue over at dotnet/coreclr later to ask whether that's accidental or intentional, but for now I'll add custom modifier replication code to Moq. Thanks for double-checking this PR! :+1:

zvirja:
@stakx Thanks for raising the investigation. It's indeed a critical issue, as you [can now rewrite](https://github.com/dotnet/coreclr/issues/18401#issuecomment-396223909) the `in` value, which supposed to be read-only.

Let's see what guys think about this.

PR #627, opened 2018-06-09
Expose invocations

Tragedian:
Second take at exposing invocation information.
Invocation information exposed via `IReadOnlyInvocation` interface.
Mock invocations exposed via property `IReadOnlyList<IReadOnlyInvocation> Invocations` on `Mock` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+#### Added
+
+* Add `Mock.Invocations` property to support inspection of invocations on a mock (@Tragedian, #560)
 
 ## 4.8.3 (2018-06-09)
 
--- a/Moq.Tests/VerifyFixture.cs
+++ b/Moq.Tests/VerifyFixture.cs
@@ -1097,7 +1097,7 @@ namespace Moq.Tests
 			var mock = new Mock<IFoo>();
 			mock.Object.Submit();
 
-			var invocation = mock.Invocations.ToArray()[0];
+			var invocation = mock.MutableInvocations.ToArray()[0];
 			Assert.False(invocation.Verified);
 
 			mock.Verify(m => m.Submit());
@@ -1112,7 +1112,7 @@ namespace Moq.Tests
 			mock.Object.Echo(2);
 			mock.Object.Echo(3);
 
-			var invocations = mock.Invocations.ToArray();
+			var invocations = mock.MutableInvocations.ToArray();
 			Assert.False(invocations[0].Verified);
 			Assert.False(invocations[1].Verified);
 			Assert.False(invocations[2].Verified);
@@ -1131,7 +1131,7 @@ namespace Moq.Tests
 			mock.Object.Echo(2);
 			mock.Object.Echo(3);
 
-			var invocations = mock.Invocations.ToArray();
+			var invocations = mock.MutableInvocations.ToArray();
 			Assert.False(invocations[0].Verified);
 			Assert.False(invocations[1].Verified);
 			Assert.False(invocations[2].Verified);
--- a/Source/AsInterface.cs
+++ b/Source/AsInterface.cs
@@ -65,7 +65,7 @@ namespace Moq
 			get { return this.owner.InnerMocks; }
 		}
 
-		internal override InvocationCollection Invocations => this.owner.Invocations;
+		internal override InvocationCollection MutableInvocations => this.owner.MutableInvocations;
 
 		internal override bool IsObjectInitialized => this.owner.IsObjectInitialized;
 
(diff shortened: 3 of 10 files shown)


stakx:
Hi @Tragedian, nice to hear back from you! I think your PR is in a mergable state as it is. Some feedback:

> I performed a Refactor->Rename operation on the property to call it `MutableInvocations`; I believe I have caught all references, to the property, unless there's some scary string-based reflection happening somewhere that I would've missed.

Looks good to me. I believe you've caught all occurrences of `Invocation` with the rename.

> Added implementation to `Invocation`; I'm not clear whether we need to copy-on-read the `Arguments` property.

Looks good to me. Copying `Arguments` on read shouldn't be necessary, since `Invocation` objects in `InvocationCollection` can be considered completed and therefore "frozen".

> I've made `InvocationCollection` implement `IReadOnlyList<IReadOnlyInvocation>`. I've used a copy-on-read strategy to make sure the enumerator is tolerant of changes to the collection. This should make any Linq extensions safe to call, including `ToArray()`.

Quick question, is there a reason why you opted *not* to make `Invocations` non-copying? (I grant you that your approach is absolutely fine from a practical perspective. I would prefer a non-copying approach however, because I think `mock.Invocations` should ideally act like any other collection in .NET and throw when a change happens during enumeration. That's why I suggested the additional interface with a `.ToArray()` method.)

A small look ahead: If Moq 5 is going to take much longer than expected, I might clean up the Moq 4 API a little bit. This may encompass, among other things, replacing the `mock.ResetCalls()` extension method in favor of a new `mock.Invocations.Clear()` method. In that case, I might opt to rework some of what you're adding here (e.g. introducing some `IInvocationList` interface with the necessary methods, in place of `IReadOnlyList<IReadOnlyInvocation>`). However, that's of yet undecided, so I think we're good to go. I just thought I'd mention it so you won't think I'm changing things behind your back just to get it my way. 😄

Tragedian:
> Quick question, is there a reason why you opted _not_ to make Invocations non-copying?

It's the `lock` statements.

I'm assuming the collection is heavily synchronised internally because there's an expectation it will be modified and read from multiple concurrent executions. On this assumption, the simplest implementation is a copy-on-read model which allows consumers to read their copy of the collection whilst allowing modifications to proceed on the synchronised collection.

This is much better to manage for a consumer than an enumerator which is always racing with potential modifications and has to be coded around failing at any moment.

It is a sub-optimal implementation of this pattern because it will produce allocations which aren't really needed. There are alternatives we could explore around using a copy-on-write method, or using some form of immutable collection internally (which would significantly help on allocations). It's implemented this way for simplicity, with an eye on potential future improvements without breaking behaviour contracts.

stakx:
> It's the `lock` statements.

Agreed, I am not exactly thrilled to have them, either. The reason they are there (for now) is because (IIRC) something like a `ConcurrentQueue<Invocation>` actually performs worse than a combination of `List<Invocation>` and `lock`s in single-threaded scenarios. The benefits of .NET's concurrent collections might only become apparent when there actually *is* some degree of concurrency.

I just ran a quick experiment with `ConcurrentQueue<Invocation>` again. I only discovered now that my expectation that all collections throw when a change happens during enumeration was naïve; this is no longer true for concurrent collections. Enumerating over a `ConcurrentQueue<>` sees a snapshot of the collection made at the time when enumeration started&mdash;which is exactly what you implemented, also.

So in summary, I'm convinced. Your implementation will allow us to later get rid of the `lock`s by moving to a `ConcurrentQueue<Invocation>` without any functional change.

stakx:
P.S.: Perhaps my look towards `ConcurrentQueue<>` is also naïve? If you can recommend a better way to get rid of the `lock`s in `InvocationCollection`, I'd be interested to learn about it.

stakx:
Merging... thank you! I'll try to release Moq 4.9.0 reasonably quickly; say, in approx. 3 weeks' time. (I'd like to allow some time to gather feedback on possible issues in 4.8.3.)

Like I mentioned above, if you have any insights regarding concurrency that you'd like to share here, I'm happy to listen.

Tragedian:
I'd have to ask why you'd want to remove those `lock` statements. I personally have no problems with them; they're simple to understand and (just as you say) very fast when resources aren't contested.

I do have slight "this-could-be-better" pains with allocating memory every time a read is performed, especially if there's no concurrency in that scenario. I'd need to understand what scenario is most common and optimise for that. If the 99% case is where there is no concurrency, it would be better to not create a copy on every read. If the collection only has to support append and clear functions, using an mutable collection protected by locks and implementing a `Range` enumeration over the collection would probably give the best balance of speed and memory use.

public IEnumerator<IReadOnlyInvocation> GetEnumerator()
{
    lock (this.invocations)
    {
        return this.invocations.Take(this.invocations.Count).GetEnumerator();
    }
}

The `Clear()` method would have to replace the existing collection with a new one, so existing enumerators could still proceed.

Tragedian:
If you still want those unit tests, I will try to put together some.

Tragedian:
Tests and collection changes: https://github.com/moq/moq4/pull/628

PR #628, opened 2018-06-10
Invocations collection improvements

Tragedian:
Add tests for `Mock.Invocations` property.
Improved `InvocationCollection` to no longer allocate a copy of the collection on enumeration.

### DIFF
--- a/Moq.Tests/InvocationsFixture.cs
+++ b/Moq.Tests/InvocationsFixture.cs
@@ -1,8 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace Moq.Tests
@@ -58,9 +54,31 @@ namespace Moq.Tests
 		    mock.Object.CompareTo(0);
 		    mock.Object.CompareTo(1);
 
-		    var count = mock.Invocations.Count();
+		    var count = 0;
 
-			Assert.Equal(mock.Invocations.Count, count);
+		    using (var enumerator = mock.Invocations.GetEnumerator())
+		    {
+			    while (enumerator.MoveNext())
+			    {
+				    Assert.NotNull(enumerator.Current);
+
+				    count++;
+			    }
+		    }
+
+			Assert.Equal(3, count);
+	    }
+
+	    [Fact]
+	    public void MockInvocationsCanBeCleared()
+	    {
+		    var mock = new Mock<IComparable>();
+
+		    mock.Object.CompareTo(new object());
+
+			mock.ResetCalls();
+
+		    Assert.Equal(0, mock.Invocations.Count);
 	    }
 	}
 }
--- a/Moq.Tests/InvocationsFixture.cs
+++ b/Moq.Tests/InvocationsFixture.cs
@@ -80,5 +80,29 @@ namespace Moq.Tests
 
 		    Assert.Equal(0, mock.Invocations.Count);
 	    }
+
+		[Fact]
+	    public void MockInvocationsCanBeRetrievedByIndex()
+	    {
+			var mock = new Mock<IComparable>();
+
+		    mock.Object.CompareTo(-1);
+		    mock.Object.CompareTo(0);
+		    mock.Object.CompareTo(1);
+
+		    var invocation = mock.Invocations[1];
+
+		    var arg = invocation.Arguments[0];
+
+			Assert.Equal(0, arg);
+	    }
+
+		[Fact]
+	    public void MockInvocationsIndexerThrowsIndexOutOfRangeWhenCollectionIsEmpty()
+	    {
+			var mock = new Mock<IComparable>();
+
+			Assert.Throws<IndexOutOfRangeException>(() => mock.Invocations[0]);
+	    }
 	}
 }
--- a/Moq.Tests/InvocationsFixture.cs
+++ b/Moq.Tests/InvocationsFixture.cs
@@ -48,5 +48,19 @@ namespace Moq.Tests
 
 			Assert.Equal(expectedArguments, invocation.Arguments);
 	    }
+
+	    [Fact]
+	    public void MockInvocationsCanBeEnumerated()
+	    {
+			var mock = new Mock<IComparable>();
+
+		    mock.Object.CompareTo(-1);
+		    mock.Object.CompareTo(0);
+		    mock.Object.CompareTo(1);
+
+		    var count = mock.Invocations.Count();
+
+			Assert.Equal(mock.Invocations.Count, count);
+	    }
 	}
 }
--- a/Moq.Tests/InvocationsFixture.cs
+++ b/Moq.Tests/InvocationsFixture.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Moq.Tests
+{
+    public class InvocationsFixture
+    {
+	    [Fact]
+	    public void MockInvocationsAreRecorded()
+	    {
+		    var mock = new Mock<IComparable>();
+
+		    mock.Object.CompareTo(new object());
+
+		    Assert.Equal(1, mock.Invocations.Count);
+	    }
+
+		[Fact]
+	    public void MockInvocationsIncludeInvokedMethod()
+		{
+			var mock = new Mock<IComparable>();
+
+			mock.Object.CompareTo(new object());
+
+			var invocation = mock.Invocations[0];
+
+			var expectedMethod = typeof(IComparable).GetMethod(nameof(mock.Object.CompareTo));
+
+			Assert.Equal(expectedMethod, invocation.Method);
+		}
+
+	    [Fact]
+	    public void MockInvocationsIncludeArguments()
+	    {
+			var mock = new Mock<IComparable>();
+
+		    var obj = new object();
+
+			mock.Object.CompareTo(obj);
+
+		    var invocation = mock.Invocations[0];
+
+		    var expectedArguments = new[] {obj};
+
+			Assert.Equal(expectedArguments, invocation.Arguments);
+	    }
+	}
+}
--- a/Source/InvocationCollection.cs
+++ b/Source/InvocationCollection.cs
@@ -69,7 +69,6 @@ namespace Moq
 		{
 			get
 			{
-				// Test 
 				if (this.invocations == null)
 				{
 					throw new IndexOutOfRangeException();
@@ -95,14 +94,6 @@ namespace Moq
 
 				this.invocations[this.count] = invocation;
 				this.count++;
-			}
-		}
-
-		public bool Any()
-		{
-			lock (this.invocationsLock)
-			{
-				return this.count > 0;
 			}
 		}
 
--- a/Source/InvocationCollection.cs
+++ b/Source/InvocationCollection.cs
@@ -69,6 +69,12 @@ namespace Moq
 		{
 			get
 			{
+				// Test 
+				if (this.invocations == null)
+				{
+					throw new IndexOutOfRangeException();
+				}
+
 				lock (this.invocationsLock)
 				{
 					return this.invocations[index];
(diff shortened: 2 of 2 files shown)


Moq.Tests/InvocationsFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+	    [Fact]
+	    public void MockInvocationsCanBeEnumerated()
+	    {
+			var mock = new Mock<IComparable>();
+
+		    mock.Object.CompareTo(-1);
+		    mock.Object.CompareTo(0);
+		    mock.Object.CompareTo(1);
+
+		    var count = mock.Invocations.Count();
```
stakx: I understand what you're doing here, but I wonder if this could be implemented more clearly. What if LINQ's `.Count()` operator has special handling for `IReadOnlyList<>` (as requested by e.g. https://github.com/dotnet/corefx/issues/24773)? Then this possibly wouldn't enumerate at all! If the test is about enumeration, it would be good to have the enumeration more visible.

Moq.Tests/InvocationsFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+
+			Assert.Equal(mock.Invocations.Count, count);
+	    }
+	}
+}
```
stakx: Could you perhaps add an additional test that verifies that `mock.ResetCalls()` empties `mock.Invocations`?

Source/InvocationCollection.cs
```diff
@@ -47,20 +47,20 @@ namespace Moq
 {
 	internal sealed class InvocationCollection : IReadOnlyList<IReadOnlyInvocation>
 	{
-		private List<Invocation> invocations;
+		private Invocation[] invocations;
```
stakx: I suppose I might be missing something obvious here, but I'd really like to avoid us reinventing the wheel and programming another version of `List<>`. Why is this necessary?

Source/InvocationCollection.cs
```diff
@@ -78,49 +78,120 @@ public int Count
...
-			lock (this.invocations)
+			lock (this.invocationsLock)
 			{
-				return this.invocations.Where(predicate).ToArray();
+				if (this.count == 0)
+				{
+					return new Invocation[0];
+				}
+
+				return this.invocations.Take(this.count).Where(predicate).ToArray();
```
stakx: I wonder if you could use `Array.Copy` or `List<T>.CopyTo` instead of a LINQ query?

Source/InvocationCollection.cs
```diff
@@ -78,49 +78,120 @@ public int Count
...
 			}
 		}
 
 		public IEnumerator<IReadOnlyInvocation> GetEnumerator()
 		{
-			lock (this.invocations)
+			lock (this.invocationsLock)
 			{
-				return this.invocations.ToList().GetEnumerator();
+				return new Enumerator(this.invocations, this.count);
```
stakx: I'm wondering, why not the LINQ range-based approach you suggested earlier?

Tragedian: `Range` creates an enumerator which operates on the underlying collection. If the collection is modified on another thread whilst working with this enumerator, at *best* you'll get an `InvalidOperationException` and at worse will end up in an unknown state. Previously, copying was used to prevent modification by another thread. This custom `Enumerator` instead takes advantage of append-only nature of the collection and allows enumeration over the known state of the collection even when the collection itself is appending new items to the collection.

Source/InvocationCollection.cs
Tragedian: If we continue to use `List<T>` you also have to use `List<T>.Enumerator` which isn't going to behave the way we want in concurrent-execution scenarios. Arrays, being primitives, have a much simpler access model, so we can know and take advantage of what operations are safe to do concurrently and which are not.

Moq.Tests/InvocationsFixture.cs
Tragedian: Sounds like a good idea to me.

Moq.Tests/InvocationsFixture.cs
Tragedian: Fair point. I'll break it out into a `foreach` to be certain.

Tragedian: Actually to be 100% certain, I'll call the enumerator manually. I know `foreach` has some special behaviours for lists.

Source/InvocationCollection.cs
stakx: Wouldn't we only end up with `List<T>.Enumerator` if we used LINQ or a `foreach` loop? What if we went for a simple `for` loop that accesses the list by index?

Source/InvocationCollection.cs
stakx: OK, fair enough, that wouldn't work. But see my comment above regarding a plain indexing `for` loop.

Source/InvocationCollection.cs
Tragedian: I did consider it, but it has the same problem. When you call `List<T>.Add(T)`, it's not concurrency-safe with its indexer. A path exists where you wind up in an unknown state, like if the underlying array has to be resized. We could try to work out how the implementation of `List<T>` works and try to handle these edge-cases, but I think it's considerably easier to just use an array. The only method with any real complexity is our `Add(Invocation)` and most of that is delegated to the fantastic `Array.Resize` method.

Source/InvocationCollection.cs
Tragedian: I think it could be re-written to use those calls, but the predicate makes it a little harder. I'm not sure if there's a benefit to do so.

Source/InvocationCollection.cs
stakx: I can't see the problem yet. Due to the locks in `InvocationCollection`, only one thread can ever access the list at a time. If an `Add` causes a resize, then that resize happens inside the `lock`. The indexer access also happens inside a `lock`, so while that happens, noone can modify the list.

stakx: Sorry if the following might be a little naïve&mdash;not sure, a single `lock` around the whole method body might be sufficient, not sure how the C# compiler transforms `lock` blocks when `yield return` gets involved:

```csharp
IEnumerator<IReadOnlyInvocation> GetEnumerator()
{
    List<Invocation> invocations;
    int count;
    lock (this.invocationsLock)
    {
        invocations = this.invocations;
        count = invocations.Count;
    }
    for (int i = 0; i < count; ++i)
    {
        lock (this.invocationsLock)
        {
            yield return invocations[i];
        }
    }
}
```

Source/InvocationCollection.cs
stakx: Less unnecessary overhead, basically. This class is pretty low-level, I usually try to avoid LINQ in such places.

Source/InvocationCollection.cs
Tragedian: A `lock` around the `for` loop would actually make it so the enumerator would acquire the collection until all elements had been iterated over. What you have should acquire a lock to the collection every time it yields an item, then release the lock.

I do like how much less code there is here. I also like that my implementation only acquires a lock once, versus once-per-item + 1 times. I could go either way.

stakx: Perhaps there is a middle way. I feel it should be possible to let the compiler generate the enumerator class, which would help reduce the amount of additional code. I'll look into this more closely a little later.

Tragedian: Added an implementation using `yield return`. It should be functionally identical to the original, except it doesn't support `IEnumerator.Reset()`.

Source/InvocationCollection.cs
```diff
@@ -69,7 +69,7 @@ public int Count
 		{
 			get
 			{
-				lock (this.invocations)
+				lock (this.invocationsLock)
 				{
 					return this.invocations[index];
```
stakx: Could you please add an explicit index-in-range check here / throw `IndexOutOfRangeException` otherwise?

Source/InvocationCollection.cs
```diff
@@ -47,20 +47,20 @@ namespace Moq
...
 		public int Count
 		{
 			get
 			{
-				lock (this.invocations)
+				lock (this.invocationsLock)
 				{
-					return this.invocations.Count;
+					return count;
 				}
```
stakx: I believe this `lock` is superfluous... please remove. Reading an `int` is an atomic operation, there's no danger of struct tearing or anything of the sort. If there is a concurrent mutation happening during a read, it doesn't really matter whether external code sees the original or updated `count`.

(In fact, we could question the whole point of having a `Count` property in a concurrent collection, since user code isn't supposed to do non-atomic accesses like `if (invocations.Count > 0) firstInvocation = invocations[0];`... but having `Count` can still be useful in single-threaded scenarios.)

Source/InvocationCollection.cs
```diff
@@ -78,46 +78,101 @@ public int Count
...
 			}
 		}
 
 		public bool Any()
 		{
-			return this.invocations.Count > 0;
+			lock (this.invocationsLock)
+			{
+				return this.count > 0;
+			}
```
stakx: This lock isn't necessary (for the same reason mentioned at `Count`), please remove.

(In fact, now that we're implementing `IEnumerable`, I wonder if this method is still needed at all: We're only using it in a single place&mdash;`VerifyNoOtherCalls`&mdash;which in typical user code is perhaps run on the order of once per unit test method, so we could probably affort just letting Moq use LINQ's `.Any()` operator and going through `.GetEnumerator()`. What do you think?)

Source/InvocationCollection.cs
```diff
@@ -78,46 +78,101 @@ public int Count
...
+					var invocation = this.invocations[i];
+					if (predicate(invocation))
+					{
+						result[resultIndex] = invocation;
+						resultIndex++;
+						resultSize++;
+					}
+				}
+
+				Array.Resize(ref result, resultSize);
```
stakx: Haha, I know I previously said that I tend to avoid LINQ queries in low-level places such as this class... but please feel free to just use a `List<>` with an initial capacity of `this.count`, then returning `.ToArray()` in the end. It's a common pattern that would save you some lines of code at only very little additional overhead.

Source/InvocationCollection.cs
Tragedian: Won't we already just get one from the array anyway? Did you want to add some specific detail?

Source/InvocationCollection.cs
Tragedian: It's not there to avoid a torn read, but to protect against reordering optimisations and variable caching. If you want to go down that avenue, it might require marking the backing field as volatile, or otherwise setting up a memory barrier. I don't advise pulling on this thread unless you think it's very important.

Source/InvocationCollection.cs
stakx: There may be no array; you set the field to `null`. I believe that an out-of-range index access should reliably throw `IndexOutOfRangeException` instead of leaking implementation details.

Tragedian: Ah, I hadn't thought of this scenario. I wouldn't want a `NullReferenceException` either.

Source/InvocationCollection.cs
Tragedian: I think LINQ's `Any()` might just call `Count` in this case. It's probably fine to remove the method.

Source/InvocationCollection.cs
```diff
@@ -69,7 +69,13 @@ public int Count
 		{
 			get
 			{
-				lock (this.invocations)
+				// Test 
+				if (this.invocations == null)
+				{
+					throw new IndexOutOfRangeException();
+				}
```
stakx: Unfortunately, this still isn't sufficient. The array can have a capacity that exceeds 'count' so you are not going to get an out-of-range exception in some cases. I suggest you just add a straightforward check for `0 <= index && index < this.count` instead. ;-)

Source/InvocationCollection.cs
stakx: That's fine with me, let's remove it.

Source/InvocationCollection.cs
stakx: I didn't know reordering happened across different methods. But yeah, this isn't super-important. Let's leave it as is, then.

Source/InvocationCollection.cs
stakx: (Sorry, I was an idiot above. I was looking at the wrong lines of code when I wrote the above, I completely overlooked the predicate.)

stakx:
Looks like you have made all the requested changes, so let's merge this. Thanks for your great work!

PR #646, opened 2018-07-19
migrate to .net standard 2.0

tomchavakis:
support .net core 2.1

### DIFF
--- a/Moq.nuspec
+++ b/Moq.nuspec
@@ -27,6 +27,15 @@
 				<dependency id="System.Threading.Tasks.Extensions" version="4.3.0" />
 				<dependency id="System.ValueTuple" version="4.4.0" />
 			</group>
+			<group targetFramework=".NETStandard2.0">
+				<dependency id="NETStandard.Library" version="2.0" />
+				<dependency id="System.Linq.Queryable" version="4.3.0" />
+				<dependency id="System.Reflection.Emit" version="4.3.0" />
+				<dependency id="System.Reflection.TypeExtensions" version="4.3.0" />
+				<dependency id="Castle.Core" version="4.3.1" />
+				<dependency id="System.Threading.Tasks.Extensions" version="4.3.0" />
+				<dependency id="System.ValueTuple" version="4.4.0" />
+			</group>
 		</dependencies>
 	</metadata>
 	<files>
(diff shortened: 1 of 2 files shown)


stakx:
Hi @tomchavakis, thanks for the PR.

You might have seen #630. Something like your pull request is definitely planned for the next minor release (4.10.0), but there'll be a patch release first (4.9.1) which is the reason why we cannot merge this into `master` just yet. Please be patient. ;-)

Also, I am starting to think that it might be preferable to actually target .NET Core 2.x directly, instead of .NET Standard. Please feel free to participate in the discussion over at #630 and update this PR accordingly once we reach a conclusion there.

Source/Moq.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net45;netstandard1.3</TargetFrameworks>
+		<TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
```
stakx: We first need to decide which frameworks we want to target before dropping older versions. This will likely need to be changed. I'll give you further feedback once we've figured out things.

Moq.nuspec
```diff
@@ -27,6 +27,15 @@
 				<dependency id="System.Threading.Tasks.Extensions" version="4.3.0" />
 				<dependency id="System.ValueTuple" version="4.4.0" />
 			</group>
+			<group targetFramework=".NETStandard2.0">
```
stakx: If we do indeed target `netcoreapp2.x` instead the dependencies will likely look a bit more minimal. This will likely need to be changed. Stay tuned for more feedback later.

Source/Moq.csproj
tomchavakis: It would be great to support the latest netstandard2.0 version in order to use it at .net core 2.x projects and also the previous versions if it's possible.

stakx:
@tomchavakis - Are you still interested in working on this? If so, please confirm briefly by posting here. If not, that's perfectly fine as well.

This PR has been dormant for a while because it wasn't sufficiently clear which platform targets Moq should support in the future. In the meantime, some new information has become available (see e.g. https://github.com/moq/moq4/issues/630#issuecomment-445060252). My current thinking is that we should probably add `net461` and `netstandard2.0` targets to Moq. (But I'm not 100% sure I'm reading the [.NET cross-platform targeting guide](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting) correctly, and that both a `net45` and `net461` target is necessary. This might warrant some testing.)

Supporting `netstandard2.0` would mean that all conditional compilation symbols have to be updated, they will no longer make much sense, nor be named correctly in all instances, in the case of .NET Standard 2.0.

tomchavakis:
@stakx Thank's for the detailed comment. I agree with you that net461 and netstandard2.0 should targets to Moq. In case you need any help in some testing for both net45 and net461 I am willing to help.

stakx:
@tomchavakis, sorry for having disregarded this  PR in favour of #784. I actually forgot about your PR until I was already done with mine. 😲

tomchavakis:
@stakx  no worries!!! I am happy that the project is updated!!!

PR #648, opened 2018-07-21
Simplify the build process & scripts

stakx:


NuGet.Restore.targets
```diff
@@ -1,91 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
kzu: So glad to see this one go...

build.cmd
```diff
@@ -1,42 +1 @@
-:: Optional batch file to quickly build with some defaults.
```
kzu: I'm not only in favor of making this minimal, I'm actually conviced it's pointless. I will most likely totally remove `build.cmd` from the moq/moq repo. I think any .NET developer coming across a root `build.proj` will know what to do. Moreover, with corebuild.io, you even get dynamic help for its props/targets! See https://github.com/moq/moq/blob/master/build.proj

build/Build.proj
```diff
@@ -0,0 +1,44 @@
+<Project DefaultTargets="Build;Test;Pack">
```
kzu: I would not move this to a less discoverable location, for the reasons metioned in the previous comment. I would also make this a corebuild.io standard project and get rid of the custom restore.

build/Build.proj
```diff
@@ -0,0 +1,44 @@
...
+  <Target Name="Test" DependsOnTargets="Build">
+    <MSBuild Projects="@(TestProjects)" Targets="Test" Properties="$(CommonBuildProperties)" BuildInParallel="False" />
+  </Target>
+
+  <Target Name="Pack" DependsOnTargets="Build">
+    <MSBuild Projects="@(Source)" Targets="Pack" Properties="$(CommonBuildProperties)" BuildInParallel="False" />
+  </Target>
+
+  <Target Name="Configure">
+    <MSBuild Projects="$(RootDirectory)Moq.sln" Targets="Restore" Properties="Step=1;$(CommonBuildProperties)" />
```
kzu: Not sure what those `Step` properties passed down are for.

build/SignAssembly.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+
+  <PropertyGroup>
```
kzu: This may be a bit too much granularity ;)

build/SignAssembly.props
```diff
@@ -0,0 +1,9 @@
...
+    <SignAssembly>True</SignAssembly>
+    <AssemblyOriginatorKeyFile>$(RootDirectory)Moq.snk</AssemblyOriginatorKeyFile>
+    <PublicSign Condition="'$(OS)' != 'Windows_NT'">True</PublicSign>
```
kzu: Can't we public sign on Windows? Really?

build/xUnit.props
```diff
@@ -0,0 +1,31 @@
+<Project TreatAsLocalProperty="TestReportFileName;DotnetErrorCode">
```
kzu: xunit is usually written in all-lowercase ;)

build/xUnit.props
```diff
@@ -0,0 +1,31 @@
...
+  </ItemGroup>
+
+  <Target Name="Test">
+    <!-- See if dotnet CLI is available: -->
+    <Exec ContinueOnError="True" Command="dotnet --info">
+      <Output TaskParameter="ExitCode" PropertyName="DotnetErrorCode"/>
+    </Exec>
+    <Warning Condition="'$(DotnetErrorCode)' != '0'" Text="Error in executing dotnet." />
+
+    <Exec Command="dotnet xunit -nobuild -configuration $(Configuration) -noshadow -appveyor -html &quot;$(OutputDirectory)$(TestReportFileName).html&quot; -xml &quot;$(OutputDirectory)$(TestReportFileName).xml&quot;"
```
kzu: Not sure why we need this instead of just using `xunit.runners.msbuild` and passing in the test assembly...

packages.config
```diff
@@ -1,8 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
kzu: yay, gone!

src/Moq/Moq.csproj
```diff
@@ -144,4 +150,11 @@
 		</None>
 	</ItemGroup>
 
+	<Target Name="SetNuspecProperties" BeforeTargets="GenerateNuspec" DependsOnTargets="GitVersion">
```
kzu: Good upgrade to versioning!

src/Moq/Moq.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
```
kzu: Next step should be to move to NuGetizer ;)

build/Build.proj
stakx: Right, agreed. If `build.cmd` indeed is allowed to go away, there should be a `build.proj` in the root.

build/Build.proj
stakx: They are a hack to prevent MSBuild from caching information about the invoked projects. The thing is that a `Restore` will create a `.props` file under `obj/` which includes build files from the NuGet packages (e.g. GitInfo). If you invoke `Restore` together with another task, MSBuild won't read the auto-generated `.props` file and won't find the NuGet package-provided targets.

I'd love to get rid of this.

build/SignAssembly.props
stakx: No idea, didn't check that one yet. But it's been there before, it's no new addition.

build/xUnit.props
stakx: Because that runner targets only `net452`. That is, won't be usable on Linux boxes without resorting to Mono. I would love to have a build script that runs on Linux, too.

Btw. I'm looking at replacing `dotnet xunit` with `xunit.runner.console`. The former is going away with xUnit 4.

src/Moq/Moq.nuspec
stakx: Will look into it!

build.cmd
stakx: Tadaa! It's gone completely now. 🍾

build/SignAssembly.props
stakx: @kzu, after reading https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/public-signing.md:

> To enable you to build assemblies that have a matching identity to what Microsoft would build, we leverage a new signing mechanism called *Public Signing*.

Why would we even want to public-sign at all? Moq's keypair used for strong-naming is available to everyone as it's included in the repository, so public-signing simply shouldn't be necessary.

It seems to me that the `<PublicSign>True</PublicSign>` should be removed.

build.cmd
stakx: It gets even better: It's perfectly possible to get rid of `build.proj` as well, meaning there will only be a `Moq.sln` left that MSBuild will picks up automatically (because it is the only project in that directory). So people can build Moq using a simple `msbuild`, `msbuild /t:Configure`, `msbuild /t:Test`, `msbuild /t:Pack`, whatever. It doesn't get any simpler.

kzu: Not a fan of that, because it's not easy (even possible?) to add solution-level targets like `Configure` or `Test`. I'd keep `build.proj` since it's the thing that you can run to get reproducible builds that look and run (almost) exactly as what the CI builds, including all steps it runs, like testing and packaging.

People building from the cmdline aren't the same (I think) than those opening `Moq.sln` straight-away. So it's fine if they have different "paths". If the sln is inside a `src` dir, then both can find their sweet spot.

build/xUnit.props
kzu: hm. I wonder how many request to run this from Linux, honestly... But fine then!

kzu:
Looks great @stakx. You are awesome 👍

build.cmd
stakx: > it's not easy (even possible?) to add solution-level targets like `Configure` or `Test`

It is. You can define them in `Before.Moq.sln.targets`.

You can also define dummy targets in `Directory.Build.props` that each solution project will "inherit" and be able to override as needed.

> I'd keep `build.proj` since it's the thing that you can run to get reproducible builds that look and run (almost) exactly as what the CI builds,

build/xUnit.props
stakx: Well, we're supporting .NET Core so why not do it thoroughly. :)

And it's not just Linux. It's also relevant for those who (accidentally) try to build using `dotnet msbuild` instead of `msbuild`. Both should work equally well, IMO.

Issue #652, opened 2018-08-06
"Different number of parameters" error when passing to Returns a dynamically compiled Expression


MaMazav:
When trying to mock the following interface:

    public interface IMyInterface
    {
        void MyAction(int x);

        int MyFunc(int x);
    }

By the following code:

            Mock<IMyInterface> mock = new Moq.Mock<IMyInterface>();
            var setupAction = mock.Setup(my => my.MyAction(It.IsAny<int>()));
            var setupFunc = mock.Setup(my => my.MyFunc(It.IsAny<int>()));

            int a = 5;
            Expression<Action<int>> callback = x => Console.WriteLine(a);
            Action<int> callbackCompiled = callback.Compile();

            int b = 4;
            Expression<Func<int, int>> funcResult = x => b;
            Func<int, int> funcResultCompiled = funcResult.Compile();

            setupAction.Callback(callbackCompiled); // OK
            setupFunc.Callback(callbackCompiled); // OK
            setupFunc.Returns(funcResultCompiled); // Fail

While setting callbacks using Callback method succeeds for method compiled on runtime, trying to mock using Returns method fail with "System.ArgumentException: 'Invalid callback. Setup on method with 1 parameter(s) cannot invoke callback with different number of parameters (2).'".

**Preliminary investigation:**
All compiled method has a "redundant" first argument of type System.Runtime.CompilerServices.Closure. In general, overcoming this redundant parameter is by calling to DynamicInvoke instead of simple Invoke, thus if I understand correctly it should not be a problem in a Moq (some answers on the web talk about that e.g. https://stackoverflow.com/questions/7935306/compiling-a-lambda-expression-results-in-delegate-with-closure-argument/7940617#7940617).
So seems that the problem is only the error method. And indeed, seems that the argument checks differ between the Callback and the Returns method internal implementations: While the former (in file src/Moq/MethodCall.cs, method SetCallbackWithArguments) calls extension method HasCompatibleParameterList to check correctness, the latter (in file src/Moq/MethodCallReturn.cs, method ValidateNumberOfCallbackParameters) have a different check of arguments which seems less stable.

(A side note, this problem was seen by another user: https://stackoverflow.com/questions/49641537/incorrect-number-of-arguments-supplied-for-call-to-method-expression-callmetho).

stakx:
[...]

PR #655, opened 2018-08-08
Align Returns with Callback method signature matching by being permissive about when HasCompatibleParameterList returned true

idigra:
Opened following a discussion in https://github.com/moq/moq4/issues/652.
Can be pulled depends on the decision in this discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * More precise `out` parameter detection for mocking COM interfaces with `[in,out]` parameters (@koutinho, #645)
 * Prevent false 'Different number of parameters' error with `Returns` callback methods that have been compiled from `Expression`s (@stakx, #654)
+* Align `Returns` with `Callback` method signature matching by being permissive about when `HasCompatibleParameterList` returned true (@idigra, #654).
 
 
 ## 4.9.0 (2018-07-13)
--- a/src/Moq/MethodCallReturn.cs
+++ b/src/Moq/MethodCallReturn.cs
@@ -171,6 +171,13 @@ namespace Moq
 		{
 			var callbackMethod = callback.GetMethodInfo();
 
+			var expectedParams = this.Method.GetParameters();
+			var actualParams = callback.GetMethodInfo().GetParameters();
+			if (callback.HasCompatibleParameterList(expectedParams))
+			{
+				return;
+			}
+
 			// validate number of parameters:
 
 			var numberOfActualParameters = callbackMethod.GetParameters().Length;


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * More precise `out` parameter detection for mocking COM interfaces with `[in,out]` parameters (@koutinho, #645)
 * Prevent false 'Different number of parameters' error with `Returns` callback methods that have been compiled from `Expression`s (@stakx, #654)
+* Align `Returns` with `Callback` method signature matching by being permissive about when `HasCompatibleParameterList` returned true (@idigra, #654).
```
stakx: The target audience of the changelog are Moq users. They likely have no clue about an internal method like `HasCompatibleParameterList`. This would have to be rephrased to be more readily understandable for users. Heck, I know Moq internals quite well, but I'm still not sure what change we're really making here!

src/Moq/MethodCallReturn.cs
```diff
@@ -171,6 +171,13 @@ private void ValidateReturnDelegate(Delegate callback)
 		{
 			var callbackMethod = callback.GetMethodInfo();
 
+			var expectedParams = this.Method.GetParameters();
+			var actualParams = callback.GetMethodInfo().GetParameters();
```
stakx: You aren't using `actualParams` anywhere... are you? Looks like a copy & paste error.

src/Moq/MethodCallReturn.cs
```diff
@@ -171,6 +171,13 @@ private void ValidateReturnDelegate(Delegate callback)
...
+			if (callback.HasCompatibleParameterList(expectedParams))
+			{
+				return;
```
stakx: I still don't understand why we are making `Returns` more permissive by possibly returning early and skipping sensible sanity checks here. *Why* are you proposing this change? Which specific use case would profit from it? Have you seen the 3 previous issues/PRs I linked to, over in the issue?

stakx: (Btw., it would be ideal to document the use cases that profit from the proposed change in the form of one or more tests.)

idigra: 1. I'm going to be more permissive exactly in places where it's OK to be, i.e. when the method is runnable. The HasCompatibleParameterList method returns true only if it can be DynamicInvoked or Invoked, so when it returns true we are sure that it's OK. I've seen the old three issues and I think that they all points to the direction I'm going to.
2. The sanity checks were proved as wrong. We had one case (in https://github.com/moq/moq4/issues/652) for which it wrongly failed. The fix in the previous PR targets exactly the problem that was caught, but we don't know if there are other cases that .NET manipulates the method signature like the Closure argument in https://github.com/moq/moq4/issues/652. IMO Checking the standard .NET Invoke method is a more stable way to eliminate such manipulations than implementing a heuristics and extend it on each case we encounter (as was really proved in the Closure bug).
3. That's also the answer for the use case question - the use case is https://github.com/moq/moq4/issues/652 . Although was already targeted by another way in a previous PR, it may be only one example among others (still unknown) cases in which the existing heuristics doesn't cover.

PR #658, opened 2018-08-19
WIP: Mock only the explicitly implemented method when another method with the same signature is present

### LINKED ISSUE #657: Moq does not mock 'Explicit Interface Implementation' and 'protected virtual' correctly
If we use [Explicit Interface Implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation), with same name as a protected virtual method, Moq will mock the wrong method.
Protected methods should not be possible to mock, due to their protection level.

We found this issue in one of our Custom HttpHandlers that inherits from HttpClientHandler.
The method we implement is a overridden protected virtual method SendAsync. To be able to mock this, we used a Interface method as a layer in between for calling base (do an actual httprequest instead for just calling the mock).

### Steps to Reproduce

Code example of code with this behaviour:
```
void Main()
{
    var mock = new Moq.Mock<Contract>() { CallBase = true };
    mock.As<IContractOveride>().Setup(m => m.DoWork()).Returns(3);
    Console.WriteLine(mock.Object.PublicWork()); //Writes out "3". Should be "1". Wrong method is mocked.
}

public class Contract : IContractOveride
{
    public int PublicWork()
    {
        return this.DoWork();
    }
    
    protected virtual int DoWork()
    {
        //Should not be able to mock this method: '...' is inaccessible due to its protection level
        ((IContractOveride)this).DoWork(); //Does not return the value of interface method.
        return 1;
    }

    int IContractOveride.DoWork()
    {
        Console.WriteLine("IFace");
        return 2;
    }
}

public interface IContractOveride
{
    int DoWork();
}
```

Working code:
```
void Main()
{
    var mock = new Moq.Mock<Contract>() { CallBase = true };
    mock.As<IContractOveride>().Setup(m => m.DoWorkOverride()).Returns(3);
    Console.WriteLine(mock.Object.PublicWork()); //Writes out "1" as expected
}

public class Contract : IContractOveride
{
    public int PublicWork()
    {
        return this.DoWork();
    }
    
    protected virtual int DoWork()
    {
        //Should not be able to mock this method: '...' is inaccessible due to its protection level
        ((IContractOveride)this).DoWorkOverride(); //Does not return the value of interface method.
        return 1;
    }

    int IContractOveride.DoWorkOverride()
    {
        Console.WriteLine("IFace");
        return 2;
    }
}

public interface IContractOveride
{
    int DoWorkOverride(); //This name is the only change.
}
```

### Expected Behavior
It should write out "1" as i am overriding the interfaced method and not the virtual one.
The virtual method should not be mocked in any circumstances, as this one is protected.

### Actual Behavior
It writes out "3" since it mocks the protected virtual method, and not the interfaced method.

stakx:
**This is a work in progress; not to be merged just yet.**

Given two mockable methods with the same signature (same name and same parameters), where one implements an interface method explicitly, setting up that one via `.As<TInterface>()` leads to the other method being mocked, too. Seems like Moq incorrect matches methods somewhere.

This will fix #657.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * More precise `out` parameter detection for mocking COM interfaces with `[in,out]` parameters (@koutinho, #645)
 * Prevent false 'Different number of parameters' error with `Returns` callback methods that have been compiled from `Expression`s (@stakx, #654)
+* Mock only the explicitly implemented method when another method with the same signature is present (@stakx, #658)
 
 
 ## 4.9.0 (2018-07-13)
--- a/src/Moq/MethodCall.cs
+++ b/src/Moq/MethodCall.cs
@@ -394,7 +394,10 @@ namespace Moq
 				return true;
 			}
 
-			if (method.DeclaringType.IsAssignableFrom(invocationMethod.DeclaringType))
+			var methodDeclaringType = method.DeclaringType;
+			var invocationMethodDeclaringType = invocationMethod.DeclaringType;
+
+			if (methodDeclaringType.IsAssignableFrom(invocationMethodDeclaringType))
 			{
 				if (!method.Name.Equals(invocationMethod.Name, StringComparison.Ordinal) ||
 					method.ReturnType != invocationMethod.ReturnType ||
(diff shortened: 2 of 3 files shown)


stakx:
Superceded by #904.

PR #659, opened 2018-08-19
Make `VerifyNoOtherCalls` play along nicely with parameterless `Verify()` and `VerifyAll()`

### LINKED ISSUE #607: Verifiable vs VerifyNoOtherCalls
Using Moq 4.8.1 and NUnit 3.9.0, with the following code:

    public interface ICat
    {
        string Purr(int amount);
    }

    [TestFixture]
    public class Tests
    {
        [Test]
        public void Purring1()
        {
            var cat = new Mock<ICat>();
            cat.Setup(x => x.Purr(15)).Returns("happy").Verifiable();

            var mood = cat.Object.Purr(15);

            cat.Verify();
            Assert.That(mood, Is.EqualTo("happy"));
            cat.VerifyNoOtherCalls();
        }

        [Test]
        public void Purring2()
        {
            var cat = new Mock<ICat>();
            cat.Setup(x => x.Purr(15)).Returns("happy");

            var mood = cat.Object.Purr(15);

            cat.Verify(x => x.Purr(15));
            Assert.That(mood, Is.EqualTo("happy"));
            cat.VerifyNoOtherCalls();
        }
    }

Purring2 passes as expected.  Purring1 unexpectedly fails with:

    Moq.MockException : The following invocations were not verified:
    ICat.Purr(15)
       at Moq.Mock.VerifyNoOtherCalls(Mock mock) in C:\projects\moq4\Source\Mock.cs:line 393

Why doesn't `Verify()` count as having verified the `Verifiable()` calls?

stakx:
This fixes #607. See commit messages for details.


src/Moq/Invocation.cs
```diff
@@ -161,5 +195,13 @@ public override string ToString()
 
 			return builder.ToString();
 		}
+
+		private enum VerificationState : byte
+		{
+			Invoked = 0,
+			InvokedAndMatchedBySetup,
+			InvokedAndMatchedByVerifiableSetup,
+			Verified,
```
stakx: Note that the logic implemented in the new `Invocation` methods ensures that we only ever progress from `Invoked` towards `Verified`, but it should never be possible to regress back to an earlier flag.

src/Moq/Mock.cs
```diff
@@ -252,6 +252,11 @@ public void Verify()
 
 		private bool TryVerify(out MockException error)
 		{
+			foreach (Invocation invocation in this.MutableInvocations)
```
stakx: There is some hidden complexity here. This returns an `IEnumerator<IInvocation>` that triggers one initial `lock` and enters a lock for each iteration. There's also a type cast here from `IInvocation` to `Invocation` so we get access to the internal `MarkAsVerifiedIf...` method.

This could be more efficient, but it's probably good enough for now.

src/Moq/Mock.cs
```diff
@@ -282,6 +287,11 @@ public void VerifyAll()
 
 		private bool TryVerifyAll(out MockException error)
 		{
+			foreach (Invocation invocation in this.MutableInvocations)
```
stakx: (The above comment applies here as well.)

tests/Moq.Tests/VerifyFixture.cs
```diff
@@ -1309,6 +1309,33 @@ public void VerifyNoOtherCalls_can_tell_apart_transitive_and_nontransitive_usage
...
+
+			var mood = cat.Object.Purr(15);
+
+			cat.Verify(x => x.Purr(15));
+			Assert.Equal("happy", mood);
+			cat.VerifyNoOtherCalls();
+		}
+
+		[Fact]
+		public void VerifyNoOtherCalls_works_together_with_parameterless_Verify()
```
stakx: A similar test for `VerifyAll` should be added as well.

stakx:
@uecasm, may I ask you to briefly review this? I'd be glad if you could take a look at the tests I added in the first commit. Do we need to test anything else?

PR #675, opened 2018-09-04
Add SkipVerifyAll on MethodCall

BlythMeister:
This enables a setup to state that a VerifyAll will not fail when called if no invocation has occured

My current workaround

image

Which is not nice as when i actually want to use this mock, i have to reset the value.

### DIFF
--- a/src/Moq/Language/Flow/SetupPhrase.cs
+++ b/src/Moq/Language/Flow/SetupPhrase.cs
@@ -199,5 +199,10 @@ namespace Moq.Language.Flow
 		{
 			this.setup.Verifiable(failMessage);
 		}
+
+		public void SkipVerifyAll()
+		{
+			this.setup.SkipVerifyAll();
+		}
 	}
 }
--- a/src/Moq/Language/IVerifies.cs
+++ b/src/Moq/Language/IVerifies.cs
@@ -77,5 +77,18 @@ namespace Moq.Language
 		/// </code>
 		/// </example>
 		void Verifiable(string failMessage);
+
+		/// <summary>
+		/// Marks the expectation as skip-able for <see cref="Mock.VerifyAll()"/>
+		/// </summary>
+		/// <example>
+		/// The following example marks the expectation as skip-able:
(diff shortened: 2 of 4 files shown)


stakx:
Hi @BlythMeister - I think we shouldn't do this. This would make the verification API more complex with little practical benefit. We already have `.Verifable()` + `.Verify()` after all, do we really need the same kind of mechanism in its "negative" form `.SkipVerifyAll()` + `.VerifyAll()`?. If you don't want to include a setup in verification, then use `.Verify()` instead of `.VerifyAll()`.

BlythMeister:
@stakx I'm using verify all to verify that only already verified actions and counts are included...is there a better way to do this?

stakx:
> @stakx I'm using verify all to verify that only already verified actions and counts are included...is there a better way to do this?

This doesn't make any sense to me. Whether you've previously verified anything doesn't matter at all for `.Verify()` nor `.VerifyAll()`. If you want to make sure that there have been no calls other than those you've previously verified, then use `.VerifyNoOtherCalls()` as the last verification step&mdash;or use a mock with `MockBehavior.Strict` and set up all allowed calls before you use your mock.

stakx:
> I am assuming from what you have said that my verify all is pointless (unless I have set something up and forgotten to add a verify in)

Sounds like it, yes.

Regarding the possible use (abuse?) of `VerifyAll` to tell you when you've forgotten to verify a setup... that's not how it's intended to be used, and whoever will read your test code probably won't recognize this intent behind it, so I'd recommend that you don't use `VerifyAll` for that purpose.

> It would also appear I could achieve the same thing by not using a strict mock, then verify what I expect to be called, followed by a call to VerifyNoOtherCalls.

Yes, that would perhaps be the best approach in this particular case.

stakx:
Given the above discussion, it appears that the proposed change is not actually necessary to resolve the issue you've described. May we close this PR?

BlythMeister:
I agree, looks like I misunderstood the usage.  Thanks for your time.

PR #676, opened 2018-09-04
Replace parameter list comparison extension methods

stakx:
There are currently several different ways in the Moq code base for comparing ordered lists of (parameter) types:

* `Enumerable.SequenceEqual`
* `Enumerable.SequenceEqual` + `AssignmentCompatibilityTypeComparer`
* `Extensions.HasSameParameterTypesAs`
* `Extensions.HasCompatibleParameterTypes`
* `Extensions.HasCompatibleParameterList`

This refactoring replaces all of the above with three new composable building blocks:

* `ParameterTypes` &mdash; wrapper for `ParameterInfo[]` so it can be treated like `Type[]`
* `Extensions.GetParameterTypes` &mdash; gets a `MethodInfo`'s parameter types as a `ParameterTypes`
* `Extensions.CompareTo` &mdash; compares any pair of `Type[]` or `ParameterTypes` using either of two comparison modes: equality, or assignment-from-compatibility.

Care has been taken to keep them fast, and free of any heap allocations.

### DIFF
--- a/src/Moq/InvocationShape.cs
+++ b/src/Moq/InvocationShape.cs
@@ -116,7 +116,7 @@ namespace Moq
 
 			if (method.IsGenericMethod)
 			{
-				if (!invocationMethod.GetGenericArguments().SequenceEqual(method.GetGenericArguments(), AssignmentCompatibilityTypeComparer.Instance))
+				if (!method.GetGenericArguments().CompareTo(invocationMethod.GetGenericArguments(), exact: false))
 				{
 					return false;
 				}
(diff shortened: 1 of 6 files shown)


src/Moq/InvocationShape.cs
```diff
@@ -116,7 +116,7 @@ private bool IsOverride(MethodInfo invocationMethod)
 
 			if (method.IsGenericMethod)
 			{
-				if (!invocationMethod.GetGenericArguments().SequenceEqual(method.GetGenericArguments(), AssignmentCompatibilityTypeComparer.Instance))
+				if (!method.GetGenericArguments().CompareTo(invocationMethod.GetGenericArguments(), exact: false))
```
stakx: Swapping the order of `method` and `invocationMethod` is no mistake; this swap previously happened in `AssignmentCompatibilityTypeComparer`, too (see below).

PR #682, opened 2018-09-05
Add VerifyNoOtherCalls to MockFactory

BlythMeister:
Enables single call to verify no other calls on all mocks generated by the factory

Based on the discussion with @stakx in #675 it appeared that i was not using the framework correctly.
The reason i chose to do it this was way the inability to perform a `VerifyNoOtherCalls` on all mocks generated from a single factory.

This new method means you can create many mocks from a single factory, verify each expectation in turn and then at the end call `Factory.VerifyNoOtherCalls` to ensure that only the items you have verified occurred for every mock generated in a given test.

Current rather nasty workaround

```
    public class MyMockRepository : MockRepository
    {
        public MyMockRepository(MockBehavior defaultBehavior) : base(defaultBehavior)
        {
        }

        public virtual void VerifyNoOtherCalls()
        {
            VerifyMocks(m =>
            {
                dynamic objectAsType = Convert.ChangeType(m, m.GetType());
                objectAsType.VerifyNoOtherCalls();
            });
        }
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * `ExpressionCompiler`: An extensibility point for setting up alternate LINQ expression tree compilation strategies (@stakx, #647)
 * `setup.CallBase()` for `void` methods (@stakx, #664)
+* `VerifyNoOtherCalls` for `MockRepository` (@BlythMeister, #682)
 
 #### Changed
 
--- a/src/Moq/MockException.cs
+++ b/src/Moq/MockException.cs
@@ -193,14 +193,29 @@ namespace Moq
 		/// <summary>
 		///   Returns the exception to be thrown when <see cref="Mock.VerifyNoOtherCalls(Mock)"/> finds invocations that have not been verified.
 		/// </summary>
-		internal static MockException UnverifiedInvocations(IEnumerable<Invocation> invocations)
+		internal static MockException UnverifiedInvocations(Mock mock, IEnumerable<Invocation> invocations)
 		{
 			return new MockException(
 				MockExceptionReason.UnverifiedInvocations,
 				string.Format(
 					CultureInfo.CurrentCulture,
 					Resources.UnverifiedInvocations,
-					string.Join(Environment.NewLine, invocations)));
+					mock.ToString(),
+					invocations.Aggregate(new StringBuilder(), (builder, setup) => builder.AppendLine(setup.ToString())).ToString()));
+		}
+
+		/// <summary>
+		///   Returns the exception to be thrown when <see cref="MockFactory.VerifyNoOtherCalls()"/> finds invocations that have not been verified.
+		/// </summary>
+		public static Exception UnverifiedInvocations(List<MockException> errors)
+		{
+			Debug.Assert(errors.All(e => e.Reason == MockExceptionReason.UnverifiedInvocations));
+
+			return new MockException(
+				MockExceptionReason.UnverifiedInvocations,
+				string.Join(
+					Environment.NewLine,
+					errors.Select(error => error.Message)));
 		}
 
 		private MockExceptionReason reason;
--- a/src/Moq/Obsolete/MockFactory.cs
+++ b/src/Moq/Obsolete/MockFactory.cs
@@ -363,8 +363,35 @@ namespace Moq
 		}
 
 		/// <summary>
-		/// Invokes <paramref name="verifyAction"/> for each mock 
-		/// in <see cref="Mocks"/>, and accumulates the resulting 
+		/// Calls <see cref="Mock{T}.VerifyNoOtherCalls()"/> on all mocks created by this factory.
+		/// </summary>
+		/// <seealso cref="Mock{T}.VerifyNoOtherCalls()"/>
+		/// <exception cref="MockException">One or more mocks had invocations that were not verified.</exception>
+		public void VerifyNoOtherCalls()
+		{
+			var errors = new List<MockException>();
+
+			foreach (var mock in mocks)
+			{
+				try
+				{
+					Mock.VerifyNoOtherCalls(mock);
+				}
+				catch (MockException error) when (error.Reason == MockExceptionReason.UnverifiedInvocations)
+				{
+					errors.Add(error);
+				}
+			}
+
+			if (errors.Count > 0)
+			{
+				throw MockException.UnverifiedInvocations(errors);
+			}
+		}
+
+		/// <summary>
+		/// Invokes <paramref name="verifyAction"/> for each mock
+		/// in <see cref="Mocks"/>, and accumulates the resulting
 		/// verification exceptions that might be
 		/// thrown from the action.
 		/// </summary>
--- a/src/Moq/Properties/Resources.resx
+++ b/src/Moq/Properties/Resources.resx
@@ -345,7 +345,7 @@ Expected invocation on the mock once, but was {4} times: {1}</value>
 		<value>Invalid callback. Setup on method with {0} parameter(s) cannot invoke callback with different number of parameters ({1}).</value>
 	</data>
 	<data name="UnverifiedInvocations" xml:space="preserve">
-		<value>The following invocations were not verified:
-{0}</value>
+		<value>The following invocations on mock '{0}' were not verified:
+{1}</value>
 	</data>
 </root>
--- a/tests/Moq.Tests/MockRepositoryFixture.cs
+++ b/tests/Moq.Tests/MockRepositoryFixture.cs
@@ -53,6 +53,61 @@ namespace Moq.Tests
 		}
 
 		[Fact]
+		public void ShouldVerifyNoOtherCalls()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			repository.Create<IFoo>();
+
+			repository.VerifyNoOtherCalls();
+		}
+
+		[Fact]
+		public void ShouldVerifyNoOtherCalls_WhenUnmatched()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var mock = repository.Create<IFoo>();
+
+			mock.Object.Do();
+
+			var mex = Assert.Throws<MockException>(() => repository.VerifyNoOtherCalls());
+			Assert.Equal(MockExceptionReason.UnverifiedInvocations, mex.Reason);
+		}
+
+		[Fact]
+		public void ShouldVerifyNoOtherCalls_WhenVerified()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var mock = repository.Create<IFoo>();
+
+			mock.Object.Do();
+
+			mock.Verify(foo => foo.Do(), Times.Once);
+			repository.VerifyNoOtherCalls();
+		}
+
+		[Fact]
+		public void VerifyNoOtherCalls_should_correctly_aggregate_unmatched_calls_from_more_than_one_mock()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var fooMock = repository.Create<IFoo>();
+			var barMock = repository.Create<IBar>();
+
+			fooMock.Object.Do();
+			barMock.Object.Redo();
+
+			var ex = Record.Exception(() => repository.VerifyNoOtherCalls());
+
+			var expectedErrorMessage = $"The following invocations on mock \'{fooMock}\' were not verified:{Environment.NewLine}" +
+									   $"IFoo.Do(){Environment.NewLine}" +
+									   Environment.NewLine +
+									   $"The following invocations on mock \'{barMock}\' were not verified:{Environment.NewLine}" +
+									   $"IBar.Redo(){Environment.NewLine}";
+
+			Assert.NotNull(ex);
+			Assert.Equal(expectedErrorMessage, ex.Message);
+		}
+
+		[Fact]
 		public void ShouldVerifyVerifiables()
 		{
 			try
@@ -76,25 +131,23 @@ namespace Moq.Tests
 		[Fact]
 		public void ShouldAggregateFailures()
 		{
-			try
-			{
-				var repository = new MockRepository(MockBehavior.Loose);
-				var foo = repository.Create<IFoo>();
-				var bar = repository.Create<IBar>();
+			var repository = new MockRepository(MockBehavior.Loose);
+			var foo = repository.Create<IFoo>();
+			var bar = repository.Create<IBar>();
 
-				foo.Setup(f => f.Do());
-				bar.Setup(b => b.Redo());
+			foo.Setup(f => f.Do());
+			bar.Setup(b => b.Redo());
 
-				repository.VerifyAll();
-			}
-			catch (MockException mex)
-			{
-				Expression<Action<IFoo>> fooExpect = f => f.Do();
-				Assert.Contains(fooExpect.ToString(), mex.Message);
+			var ex = Record.Exception(() => repository.VerifyAll());
 
-				Expression<Action<IBar>> barExpect = b => b.Redo();
-				Assert.Contains(barExpect.ToString(), mex.Message);
-			}
+			var expectedErrorMessage = $"The following setups on mock \'{foo}\' were not matched:{Environment.NewLine}" +
+									   $"IFoo f => f.Do(){Environment.NewLine}" +
+									   Environment.NewLine +
+									   $"The following setups on mock \'{bar}\' were not matched:{Environment.NewLine}" +
+									   $"IBar b => b.Redo(){Environment.NewLine}";
+
+			Assert.NotNull(ex);
+			Assert.Equal(expectedErrorMessage, ex.Message);
 		}
 
 		[Fact]
--- a/src/Moq/Mock.cs
+++ b/src/Moq/Mock.cs
@@ -415,7 +415,7 @@ namespace Moq
 				var remainingUnverifiedInvocations = unverifiedInvocations.Where(i => i != null);
 				if (remainingUnverifiedInvocations.Any())
 				{
-					throw MockException.UnverifiedInvocations(remainingUnverifiedInvocations);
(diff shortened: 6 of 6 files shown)


src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,18 @@ public virtual void VerifyAll()
 		}
 
 		/// <summary>
-		/// Invokes <paramref name="verifyAction"/> for each mock 
-		/// in <see cref="Mocks"/>, and accumulates the resulting 
+		/// Calls <see cref="Mock.VerifyNoOtherCalls"/> on all mocks created by this factory.
```
stakx: Please replace the word `factory` with `repository`.

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -52,6 +52,44 @@ public void ShouldVerifyAll()
 			}
 		}
 
+		[Fact]
+		public void ShouldVerifyNoOtherCalls()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var mock = repository.Create<IFoo>();
+
+			mock.Setup(foo => foo.Do());
```
stakx: Please remove this call to `Setup`. (`VerifyNoOtherCalls` only looks at actual invocations, it doesn't care about setups at all, so the setup is not necessary for this test.)

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -52,6 +52,44 @@ public void ShouldVerifyAll()
...
+			repository.VerifyNoOtherCalls();
+		}
+
+		[Fact]
+		public void ShouldVerifyNoOtherCalls_WhenUnmatched()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var mock = repository.Create<IFoo>();
+
+			mock.Setup(foo => foo.Do());
```
stakx: Same here: Please remove this call to `Setup`. It is unnecessary for this test.

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -52,6 +52,44 @@ public void ShouldVerifyAll()
...
+			Assert.Equal(MockExceptionReason.UnverifiedInvocations, mex.Reason);
+		}
+
+		[Fact]
+		public void ShouldVerifyNoOtherCalls_WhenVerified()
+		{
+			var repository = new MockRepository(MockBehavior.Default);
+			var mock = repository.Create<IFoo>();
+
+			mock.Setup(foo => foo.Do());
```
stakx: Same here: Please remove this call to `Setup`. It is unnecessary for this test.

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -52,6 +52,44 @@ public void ShouldVerifyAll()
...
+
+			mock.Object.Do();
+
+			mock.Verify(foo => foo.Do(), Times.Once);
+			repository.VerifyNoOtherCalls();
+		}
```
stakx: Please insert an additional blank line after this one.

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,18 @@ public virtual void VerifyAll()
...
+		/// </summary>
+		/// <seealso cref="Mock.VerifyNoOtherCalls"/>
+		/// <exception cref="MockException">One or more mocks had setups that were not satisfied.</exception>
+		public virtual void VerifyNoOtherCalls()
```
stakx: By default, methods should not be `virtual`. Do you see a reason why this method should be overridable? (If not, please make this method non-`virtual`.)

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,18 @@ public virtual void VerifyAll()
...
+		{
+			VerifyMocks(Mock.VerifyNoOtherCalls);
```
stakx: I suspect that this call won't do proper error aggregation. Let's say I'm adding another test for a repository with more than one mock:

```csharp
[Fact]
public void VerifyNoOtherCalls_should_correctly_aggregate_unmatched_calls_from_more_than_one_mock()
{
    var repository = new MockRepository(MockBehavior.Default);
    repository.Create<IFoo>().Object.Do();
    repository.Create<IBar>().Object.Redo();

    var ex = Record.Exception(() => repository.VerifyNoOtherCalls());

    Assert.Contains("The following invocations were not verified:", ex.Message);
    Assert.Contains("IFoo.Do()", ex.Message);
    Assert.Contains("IBar.Redo()", ex.Message);
}
```

This test will fail because the error message is this:

```
The following invocations were not verified:
IFoo.Do()
```

No mention of the `mock2.Redo()` call.

Could you please add a test such as the above and look into error aggregation for `VerifyNoOtherCalls`?

As an inspiration, here's how errors are aggregated for e.g. `VerifyAll`. Take this code:

```csharp
var repo = new MockRepository(MockBehavior.Default);
repo.Create<IFoo>().Setup(m => m.Do());
repo.Create<IBar>().Setup(m => m.Redo());

repo.VerifyAll();
```

This will produce an error message such as the following:

```
The following setups on mock 'Mock<IFoo:00000001>' were not matched:
IFoo m => m.Do()

The following setups on mock 'Mock<IBar:00000001>' were not matched:
IBar m => m.Redo()
```

BlythMeister: This threw up an interesting issue of multiple types of exception  needing to be grouped, so i added a new exception reason for this.

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,18 @@ public virtual void VerifyAll()
...
-		/// Invokes <paramref name="verifyAction"/> for each mock 
-		/// in <see cref="Mocks"/>, and accumulates the resulting 
+		/// Calls <see cref="Mock.VerifyNoOtherCalls"/> on all mocks created by this factory.
+		/// </summary>
+		/// <seealso cref="Mock.VerifyNoOtherCalls"/>
+		/// <exception cref="MockException">One or more mocks had setups that were not satisfied.</exception>
+		public void VerifyNoOtherCalls()
+		{
+			VerifyMocks(Mock.VerifyNoOtherCalls);
+		}
```
stakx: I would feel more comfortable if you could just implement the logic to call `VerifyNoOtherCalls` and aggregate exceptions right here rather than using `VerifyMocks`. Like you've noticed, `VerifyMocks` as it is is very limited in the way how it aggregates exceptions. It's a pity that method has been made `protected virtual` as it really only works for `Verify` and `VerifyAll`, overriding it is pretty pointless.

I greatly appreciate the effort that you made by trying to make `VerifyMocks` more general (by declaring a new `MockExceptionReason` enumeration value for multiple various verification errors etc.), but I am not sure that it would pay off. First, because Moq's verification error messages might not be precise enough in each case to still make sense when they're blindly concatenated. Second, because `MockException` simply isn't designed to carry several verification errors as a payload, and this cannot be solved cleanly with an additional enumeration value.

I'd prefer if you could leave `VerifyMocks` untouched and do something like this in `VerifyNoOtherCalls` instead:

```csharp
var errors = new List<MockException>();

foreach (var mock in mocks)
{
    try
    {
        Mock.VerifyNoOtherCalls(mock);
    }
    catch (MockException error) when (error.Reason == MockExceptionReason.UnverifiedInvocations)
    {
        errors.Add(error);
    }
}

if (errors.Count > 0)
{
    throw MockException.UnverifiedInvocations(errors);
}
```

src/Moq/MockException.cs
```diff
@@ -203,7 +203,35 @@ internal static MockException UnverifiedInvocations(IEnumerable<Invocation> invo
...
+		internal static MockException UnverifiedInvocations(IEnumerable<MockException> errors)
+		{
+			Debug.Assert(errors.All(e => e.Reason == MockExceptionReason.UnverifiedInvocations));
+
+			return new MockException(
+				MockExceptionReason.UnverifiedInvocations,
+				string.Format(
+					CultureInfo.CurrentCulture,
+					Resources.UnverifiedInvocations,
+					string.Join(Environment.NewLine, errors.Select(error => error.Message))));
```
stakx: There's two problems here:

1. Using `string.Format` on `Resources.UnverifiedInvocations` here will lead to duplicated text in the error message (since the same resource string will already be reproduced at the beginning of each `error.Message`.

   It would be better to follow the pattern that can be seen in the `MockException.UnmatchedSetups(IEnumerable<MockException> errors)` method:

   ```csharp
   return new MockException(
       MockExceptionReason.UnverifiedInvocations,
       string.Join(
           Environment.NewLine,
           errors.Select(error => error.Message)));
   ```

2. It would be important to include the failed mock's name in each original `error.Message` so that when multiple error messages are concatenated, the user can still tell which reported failures were caused by which mocks. This could be done as follows:

   * First, change the value of `Resources.UnverifiedInvocations` so it includes a placeholder for the mock's name: `The following invocations on mock {0} were not verified:\n{1}`.
   * In the `MockException.UnverifiedInvocations(IEnumerable<Invocation> invocations)` method, include the mock's name when `string.Format`-ting (which means that method needs an additional parameter for the mock:

      ```csharp
      string.Format(
          CultureInfo.CurrentCulture,
          Resources.UnverifiedInvocations,
          mock.ToString(),
          string.Join(Environment.NewLine, invocations)));
      ```

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,18 @@ public virtual void VerifyAll()
 		}
 
 		/// <summary>
-		/// Invokes <paramref name="verifyAction"/> for each mock 
-		/// in <see cref="Mocks"/>, and accumulates the resulting 
+		/// Calls <see cref="Mock.VerifyNoOtherCalls"/> on all mocks created by this factory.
+		/// </summary>
+		/// <seealso cref="Mock.VerifyNoOtherCalls"/>
```
stakx: Like above, please replace the `cref` value with `Mock{T}.VerifyNoOtherCalls()`.

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -384,13 +395,27 @@ protected virtual void VerifyMocks(Action<Mock> verifyAction)
...
+				catch (MockException error) when (error.Reason == MockExceptionReason.UnverifiedInvocations)
+				{
+					unverifiedInvocations.Add(error);
 				}
 			}
 
-			if (errors.Count > 0)
+			if (unmatchedSetups.Count > 0 && unverifiedInvocations.Count > 0)
+			{
+				throw MockException.MultipleVerificationErrors(unmatchedSetups.Concat(unverifiedInvocations));
```
stakx: *If* we made `VerifyMocks` truly general, such that it could aggregate any number and type of verification error, we shouldn't even attempt to special-case `UnmatchedSetups` and `UnverifiedInvocations` (further below); we should just always do `MultipleVerificationErrors`. In that scenario, it would also be better to have only one `catch` clause (further above) and do a `when (error.IsVerificationError)`.

*However*, I don't think such a generic aggregation would pay off in the end, nor would it be a terribly clean as long as `MockException` nor the various verification error messages are designed with aggregation in mind. (Better to do that in a separate PR, as it would touch many parts of the code that aren't directly related to `VerifyNoOtherCalls`.)

I've always thought `VerifyMocks` was mis-designed, it's `protected virtual` but there's probably no point in ever overriding it. I think it would be better to just leave this method as is and accept that it's only really useful as an implementation detail of `Verify` and `VerifyAll`. That's why I suggested in another review comment (above) to just do `VerifyNoOtherCalls`-specific aggregation in `MockFactory.VerifyNoOtherCalls()` directly.

src/Moq/Properties/Resources.Designer.cs
```diff
@@ -326,6 +326,16 @@ internal class Resources {
 				return ResourceManager.GetString("MoreThanOneCall", resourceCulture);
 			}
 		}
+        
+		/// <summary>
+		///   Looks up a localized string similar to The following verification errors occured:
+		///{0}.
+		/// </summary>
+		internal static string MultipleVerificationErrors {
```
stakx: Please remove this, along with the change in `Resources.resx`.

src/Moq/Properties/Resources.resx
```diff
@@ -346,6 +346,10 @@ Expected invocation on the mock once, but was {4} times: {1}</value>
 	</data>
 	<data name="UnverifiedInvocations" xml:space="preserve">
 		<value>The following invocations were not verified:
+{0}</value>
+	</data>
+	<data name="MultipleVerificationErrors" xml:space="preserve">
```
stakx: Please remove this, along with the change in `Resources.Designer.cs`.

Instead, as mentioned in another review comment, change the value of `UnverifiedInvocations` (right above these lines) to include a placeholder for the mock's name. (See other review comment for details.)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * `ExpressionCompiler`: An extensibility point for setting up alternate LINQ expression tree compilation strategies (@stakx, #647)
 * `setup.CallBase()` for `void` methods (@stakx, #664)
+* `VerifyNoOtherCalls` for `VerifyNoOtherCalls` (@BlythMeister #682)
```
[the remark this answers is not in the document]
BlythMeister: That is an epic fail on my typing.  Was meant to say VerifyNoOtherCalls for MockRepository

BlythMeister:
@stakx I see where your coming from and I too was dubious over the way the usage was going, so I'll revert a bit and implement the aggregate exception nicer.

CHANGELOG.md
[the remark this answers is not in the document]
stakx: That corrected version sounds fine. :laughing:

BlythMeister:
@stakx all updated (i rebased with a squash to clean the history, hope that's OK)

stakx:
@BlythMeister - Rebasing, squashing, fixing up, etc. your PR to keep it clean and tidy is very much appreciated! :+1: Will review again in a bit.

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -363,8 +363,35 @@ public virtual void VerifyAll()
 		}
 
 		/// <summary>
-		/// Invokes <paramref name="verifyAction"/> for each mock 
-		/// in <see cref="Mocks"/>, and accumulates the resulting 
+		/// Calls <see cref="Mock{T}.VerifyNoOtherCalls()"/> on all mocks created by this factory.
+		/// </summary>
+		/// <seealso cref="Mock{T}.VerifyNoOtherCalls()"/>
+		/// <exception cref="MockException">One or more mocks had setups that were not satisfied.</exception>
```
stakx: Sorry for not noticing earlier, but this should read `invocations that were not verified`, not `setups that were not satisfied`.

BlythMeister:
Thank you for keeping bearing with me, I really appreciate it, and have already learnt a lot about moq!
I will take a look at the error messages in the morning.

BlythMeister:
@stakx updates made, i did also update the test for the verify to make it match the new test which ensures the full error message is correct.

stakx:
@BlythMeister - did you see the [two test failures](https://ci.appveyor.com/project/MobileEssentials/moq4/build/4.9.0.78+sha.1a87d7b#L464)? (They appear to be failing due to different line endings (CRLF vs LF). Instead of using `"\r\n"` (CRLF) in your test's string literals, try using `Environment.NewLine` (which will get resolved to whatever line separator your current platform uses).

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -76,25 +131,23 @@ public void ShouldVerifyVerifiables()
...
+			var ex = Record.Exception(() => repository.VerifyAll());
 
-				Expression<Action<IBar>> barExpect = b => b.Redo();
-				Assert.Contains(barExpect.ToString(), mex.Message);
-			}
+			var expectedErrorMessage = $"The following setups on mock \'{foo}\' were not matched:\r\n" +
+										"IFoo f => f.Do()\r\n" +
+										"\r\n" +
+									   $"The following setups on mock \'{bar}\' were not matched:\r\n" +
+										"IBar b => b.Redo()\r\n";
```
stakx: This and one other test currently fails on AppVeyor. I suspect this is due to your use of `"\r\n"` (CRLF) which isn't used on all system. Please try replacing it with `Environment.NewLine`.

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -52,6 +52,61 @@ public void ShouldVerifyAll()
...
+			fooMock.Object.Do();
+			barMock.Object.Redo();
+
+			var ex = Record.Exception(() => repository.VerifyNoOtherCalls());
+
+			var expectedErrorMessage = $"The following invocations on mock \'{fooMock}\' were not verified:\r\n" +
+									   "IFoo.Do()\r\n" +
+									   "\r\n" +
+									   $"The following invocations on mock \'{barMock}\' were not verified:\r\n" +
+									   "IBar.Redo()\r\n";
```
stakx: This and one other test currently fail on AppVeyor. I suspect this is due to your use of `"\r\n"` (CRLF) which isn't used on all system. Please try replacing it with `Environment.NewLine`.

BlythMeister: Facepalm!

stakx: Seems I wasn't completely right about this, obviously `Resources.resx` embeds string resources with different line endings than what `Environment.NewLine` reports. 😒 Since I've sort-of led you down the wrong rabbit hole, I'll take responsibility and fix this (and other tests that compare multi-line strings).

stakx:
@BlythMeister - I've just pushed Moq 4.10.0 (which includes your addition) to NuGet. It should be available shortly.

PR #683, opened 2018-09-05
Switch to MIT license

### LINKED ISSUE #669: License update to project
I was hoping the project owner would consider updating the project license to use spdx identifier for the "BSD license" license. Your license in the github api only shows as "Other".
See https://api.github.com/repos/moq/moq4/license for your current license.
Also note your current license committed in the repo is already "[This is the BSD license, see
 http://opensource.org/licenses/BSD-3-Clause]" https://raw.githubusercontent.com/moq/moq4/master/License.txt
This is no change to the license at all, but then allows anyone using the github license API to get an exact match of the licensed used and not have to use any string magic to come up with your license.

If your unfamiliar with SPDX see spdx.org.

Pretty sure this can be done in your settings somewhere. Here is an example of your API json and your
spdx_id is null.
{
"name": "License.txt",
"path": "License.txt",
"sha": "c284be82c2f6c82cce6a1b182311b3b7be308dd7",
"size": 1735,
"url": "https://api.github.com/repos/moq/moq4/contents/License.txt?ref=master",
"html_url": "https://github.com/moq/moq4/blob/master/License.txt",
"git_url": "https://api.github.com/repos/moq/moq4/git/blobs/c284be82c2f6c82cce6a1b182311b3b7be308dd7",
"download_url": "https://raw.githubusercontent.com/moq/moq4/master/License.txt",
"type": "file",
"content": "Q29weXJpZ2h0IChjKSAyMDA3LiBDbGFyaXVzIENvbnN1bHRpbmcsIE1hbmFz\nIFRlY2hub2xvZ3kgU29sdXRpb25zLCBJblNURURECmh0dHA6Ly93d3cubW9x\ndGhpcy5jb20vCkFsbCByaWdodHMgcmVzZXJ2ZWQuCgpSZWRpc3RyaWJ1dGlv\nbiBhbmQgdXNlIGluIHNvdXJjZSBhbmQgYmluYXJ5IGZvcm1zLCAKd2l0aCBv\nciB3aXRob3V0IG1vZGlmaWNhdGlvbiwgYXJlIHBlcm1pdHRlZCBwcm92aWRl\nZCAKdGhhdCB0aGUgZm9sbG93aW5nIGNvbmRpdGlvbnMgYXJlIG1ldDoKCiAg\nICAqIFJlZGlzdHJpYnV0aW9ucyBvZiBzb3VyY2UgY29kZSBtdXN0IHJldGFp\nbiB0aGUgCiAgICBhYm92ZSBjb3B5cmlnaHQgbm90aWNlLCB0aGlzIGxpc3Qg\nb2YgY29uZGl0aW9ucyBhbmQgCiAgICB0aGUgZm9sbG93aW5nIGRpc2NsYWlt\nZXIuCiAgICAKICAgICogUmVkaXN0cmlidXRpb25zIGluIGJpbmFyeSBmb3Jt\nIG11c3QgcmVwcm9kdWNlIAogICAgdGhlIGFib3ZlIGNvcHlyaWdodCBub3Rp\nY2UsIHRoaXMgbGlzdCBvZiBjb25kaXRpb25zIAogICAgYW5kIHRoZSBmb2xs\nb3dpbmcgZGlzY2xhaW1lciBpbiB0aGUgZG9jdW1lbnRhdGlvbiAKICAgIGFu\nZC9vciBvdGhlciBtYXRlcmlhbHMgcHJvdmlkZWQgd2l0aCB0aGUgZGlzdHJp\nYnV0aW9uLgogICAgCiAgICAqIE5laXRoZXIgdGhlIG5hbWUgb2YgQ2xhcml1\ncyBDb25zdWx0aW5nLCBNYW5hcyBUZWNobm9sb2d5IFNvbHV0aW9ucyBvciBJ\nblNURUREIG5vciB0aGUgCiAgICBuYW1lcyBvZiBpdHMgY29udHJpYnV0b3Jz\nIG1heSBiZSB1c2VkIHRvIGVuZG9yc2UgCiAgICBvciBwcm9tb3RlIHByb2R1\nY3RzIGRlcml2ZWQgZnJvbSB0aGlzIHNvZnR3YXJlIAogICAgd2l0aG91dCBz\ncGVjaWZpYyBwcmlvciB3cml0dGVuIHBlcm1pc3Npb24uCgpUSElTIFNPRlRX\nQVJFIElTIFBST1ZJREVEIEJZIFRIRSBDT1BZUklHSFQgSE9MREVSUyBBTkQg\nCkNPTlRSSUJVVE9SUyAiQVMgSVMiIEFORCBBTlkgRVhQUkVTUyBPUiBJTVBM\nSUVEIFdBUlJBTlRJRVMsIApJTkNMVURJTkcsIEJVVCBOT1QgTElNSVRFRCBU\nTywgVEhFIElNUExJRUQgV0FSUkFOVElFUyBPRiAKTUVSQ0hBTlRBQklMSVRZ\nIEFORCBGSVRORVNTIEZPUiBBIFBBUlRJQ1VMQVIgUFVSUE9TRSBBUkUgCkRJ\nU0NMQUlNRUQuIElOIE5PIEVWRU5UIFNIQUxMIFRIRSBDT1BZUklHSFQgT1dO\nRVIgT1IgCkNPTlRSSUJVVE9SUyBCRSBMSUFCTEUgRk9SIEFOWSBESVJFQ1Qs\nIElORElSRUNULCBJTkNJREVOVEFMLCAKU1BFQ0lBTCwgRVhFTVBMQVJZLCBP\nUiBDT05TRVFVRU5USUFMIERBTUFHRVMgKElOQ0xVRElORywgCkJVVCBOT1Qg\nTElNSVRFRCBUTywgUFJPQ1VSRU1FTlQgT0YgU1VCU1RJVFVURSBHT09EUyBP\nUiAKU0VSVklDRVM7IExPU1MgT0YgVVNFLCBEQVRBLCBPUiBQUk9GSVRTOyBP\nUiBCVVNJTkVTUyAKSU5URVJSVVBUSU9OKSBIT1dFVkVSIENBVVNFRCBBTkQg\nT04gQU5ZIFRIRU9SWSBPRiBMSUFCSUxJVFksIApXSEVUSEVSIElOIENPTlRS\nQUNULCBTVFJJQ1QgTElBQklMSVRZLCBPUiBUT1JUIChJTkNMVURJTkcgCk5F\nR0xJR0VOQ0UgT1IgT1RIRVJXSVNFKSBBUklTSU5HIElOIEFOWSBXQVkgT1VU\nIE9GIFRIRSBVU0UgCk9GIFRISVMgU09GVFdBUkUsIEVWRU4gSUYgQURWSVNF\nRCBPRiBUSEUgUE9TU0lCSUxJVFkgT0YgClNVQ0ggREFNQUdFLgoKW1RoaXMg\naXMgdGhlIEJTRCBsaWNlbnNlLCBzZWUKIGh0dHA6Ly9vcGVuc291cmNlLm9y\nZy9saWNlbnNlcy9CU0QtMy1DbGF1c2VdCg==\n",
"encoding": "base64",
"_links": {
"self": "https://api.github.com/repos/moq/moq4/contents/License.txt?ref=master",
"git": "https://api.github.com/repos/moq/moq4/git/blobs/c284be82c2f6c82cce6a1b182311b3b7be308dd7",
"html": "https://github.com/moq/moq4/blob/master/License.txt"
},
"license": {
"key": "other",
"name": "Other",
"spdx_id": null,
"url": null,
"node_id": "MDc6TGljZW5zZTA="
}
}

kzu:
(Closes #669.)


stakx:
@kzu: Like I commented [over in Gitter](https://gitter.im/moq/moq?at=5b900916d457a1406ca5ab09):

1. Do we need to ask for the other nominal copyright holders' agreement before changing the license?

2. I can do a global search & replace of the copyright notice in all source files after this (plus an update [in `Moq.nuspec`](https://github.com/moq/moq4/blob/9e00d7304c699b51c19b5acc0049bb2a81803ddf/src/Moq/Moq.nuspec#L9)). Would you like to keep a complete copy of the license text in each source file, or would an abbreviated version referring to the `LICENSE` file do?

3. Renaming the license file will break license links for previously published NuGet packages. (Just mentioning this in case you didn't consider this side effect but would like to avoid it.)

kzu:
Thanks @stakx. I don't think it would be a problem. @waj @edjez any issues switching to MIT?

I'd just remove the copyright header from all source files. Having it in the repo root should be enough. 

Alternatively, Roslyn uses a one-liner like:

```
// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```

Maybe the rename is pointless, however. Also, I don't want to make this more work than it deserves. Fixing up the text just to show up properly should also be enough. I'd fine either way. 

For vNext, I'm using MIT, FWIW.

edjez:
@kzu MIT works fine.

stakx:
@kzu - TL;DR: I would advise that we stay on the BSD-3-Clause license for the rest of Moq 4's remaining lifetime (which is probably quite short anyway), because changing the license causes problems with already released NuGet releases. Read on for details.

> Maybe the rename is pointless, however. Also, I don't want to make this more work than it deserves. Fixing up the text just to show up properly should also be enough. I'd fine either way.

Actually, both a rename and a content change is problematic for already released versions of Moq:

* If we rename the file to `LICENSE`, then we'll break the license URL for previously released NuGet packages. People would no longer have a way of finding out what the license for older releases was.

* If we change `License.txt` to contain the MIT license, we retroactively change the license for older versions of Moq, too. We'd be exchanging licenses behind the backs of users, not sure this is a good thing for us to do.

The only clean solution would be to have *two* license files: `License.txt` for already released NuGet versions, and `LICENSE` for new releases. But two license files in a repo is also confusing.

So, while I have absolutely no objections to the MIT license per se, I think the license change causes more trouble than necessary. I would advise that we stay on BSD-3-Clause for the rest of Moq 4's remaining lifetime (which is probably quite short anyway).

stakx:
I believe there's a lesson for us here, too: If the `<licenseUrl>` of a NuGet package references a file on GitHub (or comparable), it should reference that file *at a specific commit/tag*. For example:

```xml
<licenseUrl>https://raw.githubusercontent.com/moq/moq4/v4.9.0/License.txt</licenseUrl>
                                                       ^^^^^^
```

rather than:

```xml
<licenseUrl>https://raw.githubusercontent.com/moq/moq4/master/License.txt</licenseUrl>
                                                       ^^^^^^
```

stakx:
Awesome, thanks for confirming. I'll continue as follows:

1. Close this PR.
2. Reformat `License.txt` in order to make it GitHub/Licensee-friendly (but without changing the legal content).
3. Optionally replace the long copyright text in each source files with a one-liner or two-liner that mentions the copyright year, copyright holders, license name, but refers to the license file for most everything else.

PR #686, opened 2018-09-06
Remove or deprecate unsupported verbs after post-`Returns` callback

### LINKED ISSUE #668: `.Callback(…)` after `.Returns(…)` / `.CallBase()` causes a variety of issues
Moq has allowed a `.Callback(…)` to occur after `.Returns(…)` or `.CallBase()` for a while now. This was originally discussed in [this discussion on Google Groups](https://groups.google.com/forum/#!topic/moqdisc/cVGt2YR0sEA) and then formally requested in [Google Code issue 19](https://code.google.com/archive/p/moq/issues/19).

Unfortunately, the way after-return callbacks are currently implemented in the fluent API, they cause a variety of issues:

* [ ] #665: Fluent API allows setting the return value and throwing an exception at the same time
* [ ] #666: Fluent API allows setting the return value twice
* [ ] #667: Fluent API allows setting up a second callback that won't get involved if `.Throws()` is specified too

These issues result largely from the fluent API interface `ICallback` being reused inappropriately, instead of introducing a dedicated and more limited `ICallbackAfterReturn`.

It can be argued that the ability of configuring a callback after `.Returns(…)` or `.CallBase()` is fundamentally wrong, since in regular methods everything after a `return` statement would be unreachable. The order of the `Callback`, `CallBase`, `Raises`, `Returns`, `Throws` verbs in a Moq setup generally represents the order in which these steps will be executed at runtime, so in analogy to regular methods, one could expect that `.Returns(…)` should be the final verb.

Here are few options (there may be more) how to fix this:

1. Rename the after-return `Callback` verb to `Finally`:

   ```csharp
   mock.Setup(m => m.Message).Callback(…).Returns(…).Finally(…);
                                                     ^^^^^^^^^^
   ```

   This would be a breaking change for two reasons: (1) Different method name. (2) Different semantics: A `finally` gets invoked even in the presence of an exception. An after-return callback currently won't get invoked if an exception happens beforehand.

2. Fix the fluent API so `.Returns(…)` and `.CallBase()` return an interface that doesn't inherit `ICallback`, but a new `ICallbackAfterReturns` interface instead which has more limited "flow" options. This could fix the issues linked above (at the cost of quite a bit of additional internal code duplication), but it doesn't even address the more fundamental issue.

3. Don't fix the problems linked above. Instead, deprecate after-return `Callback` and possibly remove this facility in Moq v5 or v6.

stakx:
This closes #668 (most of it, anyway).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * `Verify` exception should report configured setups for delegate mocks (@stakx, #679)
 * `Verify` exception should include complete call expression for delegate mocks (@stakx, #680)
 
+#### Obsoleted
+
+* For non-`void` methods, the setup sequence `Returns`/`CallBase` &rarr; `Callback` &rarr; `Throws` has been deprecated because a method cannot return a value and throw an exception at the same time (@stakx, #686)
+
 
 ## 4.9.0 (2018-07-13)
 
--- a/src/Moq/Language/Flow/ICallbackAfterResult.cs
+++ b/src/Moq/Language/Flow/ICallbackAfterResult.cs
@@ -1,0 +1,66 @@
+//Copyright (c) 2007. Clarius Consulting, Manas Technology Solutions, InSTEDD
+//https://github.com/moq/moq4
+//All rights reserved.
+
+//Redistribution and use in source and binary forms,
+//with or without modification, are permitted provided
+//that the following conditions are met:
+
+//    * Redistributions of source code must retain the
+//    above copyright notice, this list of conditions and
+//    the following disclaimer.
+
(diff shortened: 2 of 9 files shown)


stakx:
Closing this, as it has an unwanted side effect:

`ICallback` interface members are still visible

Because `ICallbackAfter` inherits `ICallback` (which it must in order to avoid a binary breaking change), and we cannot just mark the members of `ICallback` as `[Obsolete]` (they're still perfectly valid *before* `Returns`/`CallBase`), we end up with Visual Studio's IntelliSense displaying these members as well when it should only display those from `ICallbackAfter`. We end up with every `Callback` method being shown twice.

Let's not do this. Let's live with the unwanted verbs for now, except for two very unlikely use cases, they'll actually work in practice. Using a Roslyn analyzer to mark them as problematic would likely be the better option here.

PR #691, opened 2018-09-08
 Make `MockFactory.VerifyMocks` do what it says it does

stakx:
The XML documentation comment for `MockFactory.VerifyMocks(Action<Mock> verifyAction)` states the purpose of that method as follows:

> Invokes verifyAction for each mock in [the repository], and accumulates the resulting verification exceptions that might be thrown from the action.

However this is not what the method has been doing up until now. It hasn't accumulated any verification errors, but only those caused by unmatched setups (i.e. those exceptions thrown by `Verify` and `VerifyAll`).

This PR makes the `VerifyMocks` method adhere to its own stated purpose by cleaning up `MockException` aggregation a little. This also allows us to reuse `VerifyMocks` in `MockFactory.VerifyNoOtherCalls`.


BlythMeister:
@stakx no need to apologize!  Just glad that my change made you look at this code 👍

PR #693, opened 2018-09-18
Support mocking of event subscription using SetupSet

idigra:
I needed a functionality of mocking the add/remove methods of an vent and I wondered if this contribution is helpful for others.
There are two possible concerns: First, now mocking an event has two APIs. One - the old Raise method, and the other - by intercepting the add/remove methods. Second, for backward compatibility I needed to leave MockBehavior.Strict to ignore non explicitly mocked events (not only BW, it seems reasonable due to the classic API of the Raise method). In my opinion it still worth it.

If deciding to pull, I think that some unit tests still missing (I added only one to demonstrate the use). I'll be happy to hear what should be covered.

### DIFF
--- a/src/Moq/Extensions.cs
+++ b/src/Moq/Extensions.cs
@@ -72,6 +72,11 @@ namespace Moq
 			return method.Name.StartsWith("remove_", StringComparison.Ordinal);
 		}
 
+		public static bool ProbablyEventAttachOrDetach(this MethodInfo method)
+		{
+			return method.IsSpecialName && (method.LooksLikeEventAttach() || method.LooksLikeEventDetach());
+		}
+
 		/// <summary>
 		/// Tests if a type is a delegate type (subclasses <see cref="Delegate" />).
 		/// </summary>
--- a/src/Moq/Interception/InterceptionAspects.cs
+++ b/src/Moq/Interception/InterceptionAspects.cs
@@ -136,7 +136,7 @@ namespace Moq
 				matchedSetup.Execute(invocation);
 				return InterceptionAction.Stop;
 			}
-			else if (mock.Behavior == MockBehavior.Strict)
+			else if (mock.Behavior == MockBehavior.Strict && !invocation.Method.ProbablyEventAttachOrDetach())
 			{
 				throw MockException.NoSetup(invocation);
 			}
@@ -178,6 +178,8 @@ namespace Moq
 			}
 
 			var methodName = invocation.Method.Name;
+			EventInfo eventInfo = null;
+			Action<string, Delegate> modifyEvent = null;
 
 			// Special case for event accessors. The following, seemingly random character checks are guards against
 			// more expensive checks (for the common case where the invoked method is *not* an event accessor).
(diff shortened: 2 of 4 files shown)


stakx:
Hi @idigra, thanks for taking the time to prepare a PR. I won't be able to review it in detail today, but I'll get back to you sometime later this week.

src/Moq/Mock.cs
```diff
@@ -431,6 +431,8 @@ internal static void VerifyNoOtherCalls(Mock mock)
 
 		#region Setup
 
+		internal bool HasEventSetup { get; private set; } = false;
```
stakx: `Mock` shouldn't have any state. This should be declared `abstract`, and then overridden in both `Mock<T>` and `AsInterface<TInterface>`.

src/Moq/Mock.cs
```diff
@@ -612,7 +614,11 @@ private static VoidSetupPhrase<T> SetupSetPexProtected<T>(Mock<T> mock, Action<T
 				}
 
 				var setter = last.Invocation.Method;
-				if (!setter.IsPropertySetter())
+				if (setter.ProbablyEventAttachOrDetach())
```
stakx: In my opinion, this doesn't belong here. It doesn't make any sense to me to have logic for an event `add`/`remove` accessor in a method that is about property  `set`-ters.

src/Moq/Interception/InterceptionAspects.cs
```diff
@@ -178,53 +178,47 @@ public override InterceptionAction Handle(Invocation invocation, Mock mock)
...
+			if (eventInfo != null)
+			{
+				// TODO: We could compare `invocation.Method` and `eventInfo.GetRemoveMethod()` (or 
+				// `eventInfo.GetAddMethod()` here.
+				// If they are equal, then `invocation.Method` is definitely an event `remove` accessor.
+				// Not sure whether this would work with F# and COM; see commit 44070a9.
+
+				if (mock.CallBase && !invocation.Method.IsAbstract)
+				{
+					if (!mock.HasEventSetup)
```
stakx: If I am understanding correctly, this means that when `mock.CallBase == true`, the base class' event accessor will no longer be called once an event (un-) subscription has been set up using `SetupSet`. This also means that an event (un-) subscription will be recorded like any other method call. Is this correct?

If so, then there's a piece missing here: You'd also need some `VerifySet` (or better, `VerifyAdd`/`VerifyRemove`) method that targets event handler (un-) subscriptions. Otherwise&mdash;unless you've marked the `SetupSet` [sic]  setup as `.Verifiable()`&mdash;you might run into situations where `VerifyNoOtherCalls` will always fail because you have unverifiable invocations to event accessors that you cannot reach using any of the other `Verify` methods.

That being said, I would prefer it even more if we did *not* have to add any more delegate-based `Setup` or `Verify` methods (such as `SetupSet` is), if possible. (The whole process behind it&mdash;"I'll try to figure out what the expression tree might have looked like by putting the mock in a dry-run state, then invoking the delegate and observing some side effects it causes"&mdash;strikes me as pretty messy and very limited, it just opens more doors for further bugs.)

src/Moq/Extensions.cs
```diff
@@ -72,6 +72,11 @@ public static bool LooksLikeEventDetach(this MethodInfo method)
 			return method.Name.StartsWith("remove_", StringComparison.Ordinal);
 		}
 
+		public static bool ProbablyEventAttachOrDetach(this MethodInfo method)
+		{
+			return method.IsSpecialName && (method.LooksLikeEventAttach() || method.LooksLikeEventDetach());
```
stakx: Not sure why we need this method at all. Why not simply use the pre-existing methods `LooksLike...`?

You might have noticed that Moq doesn't do the obvious and check `IsSpecialName` when trying to find out if a method is a property or event accessor. There's apparently some history behind this curious lack, which I tried to document: 

https://github.com/moq/moq4/blob/22cce9c3f5a7e6a66f079ad719a7e415636430bb/src/Moq/Interception/InterceptionAspects.cs#L191-L193

I have never looked deeply into the rationale behind 44070a9; perhaps it no longer exists today; I don't know.

* If you know for sure that checking `IsSpecialName` is safe, even in the face of F# and COM code, then I suggest we drop the `ProbablyEventAttachOrDetach` method you're adding, and improve the existing methods instead.

* If OTOH you're not positively sure either, then let's remove your new method and use the existing ones directly.

idigra:
> Neither API can fully take the place of the other, I think.

Technically, the new API can take the place of the other. When you can mock the event subscription, you can manually raise events. However, besides backward compatibility, it's much less natural and much complicated to do it with the new API, so I would agree that the two APIs are still required.

src/Moq/Mock.cs
idigra: OK, I can change it like you asked.

Just as a side note - notice that this property is used only to preserve the optimization documented in a comment in InterceptionAspects.cs:
> Special case for event accessors. The following, seemingly random character checks are guards against more expensive checks (for the common case where the invoked method is *not* an event accessor).

We can eliminate this property at all if you prefer.

src/Moq/Extensions.cs
idigra: I wasn't aware to the issue about `IsSpecialName`. I don't cae to revert this method and usethe original ones.

src/Moq/Mock.cs
idigra: I'm for adding `SetupAdd` / `SetupRemove`. However I'm not sure how can we avoid them from being delegate-based (but the up-side is that we don't need the other overload - the Expression based overload).

src/Moq/Interception/InterceptionAspects.cs
idigra: Yeah I missed that point. I now understand more deeply your concern. However I still see no way to bypass the C# limitation that event subscription cannot be used in expression tree because it is recognized as assignment.
Anyway, if we do go with the current proposed API, I'm not sure how to solve the particular problem you pointed on because I'm not familiar with the whole process of base classes.

src/Moq/Mock.cs
stakx: Glad to hear you're not set on `SetupSet` for events.

I might try to get some feedback from @kzu on new API methods before making a decision here. I'd like to align Moq 4 with 5 as far as is reasonably possible. If mocking event (un-) subscription isn't planned at all for 5 then I'd find it difficult to add it so late in 4's lifecycle.

We wouldn't be able to avoid delegates if we wanted to continue Moq's current event syntax (`mock => mock.SomeEvent += ...`). That wouldn't be a total no-go but there are some refactorings / code cleanups I'd like to do before adding further static `Setup...` methods.

I'll post back here as soon as I have more news!

src/Moq/Mock.cs
stakx: Hmm, then I misunderstood what the new property is for. I thought it was needed to avoid / mitigate a breaking change (see discussion in preceding review comment). Will look into this more closely soon.

Don't worry too much about that particular micro-optimization. We can always optimize again at a later time (after code changes, if needed at all).

src/Moq/Interception/InterceptionAspects.cs
stakx: You're right, delegate-based methods would likely be unavoidable due to the limitations of LINQ expression trees.

Side note: I could provide some guidance regarding the event-related code paths & the `ReturnBase` stuff (which admittedly is not very easy to understand), for now the main thing is that we understand that additional `Verify{Add,Remove}` methods would likely be needed.

src/Moq/Extensions.cs
stakx: I think that'd be best for now. I'll do some research this weekend to see if I find out more about why Moq avoids using `IsSpecialName`. I'll post back here about my findings (if any).

stakx: Here's why Moq does not test for `.IsSpecialName` for events: https://github.com/moq/moq4/pull/712.

src/Moq/Mock.cs
stakx: @idigra: The code cleanups I was referring to are now finally done. Moq now has some new machinery (`ExpressionReconstructor`) that takes a delegate and turns it back into an expression. So it should now be fairly easy to add more delegate-based methods (`SetupAdd`, `SetupRemove`, `VerifyAdd`, `VerifyRemove`).

Just wanted to ping you to check whether you'd still live to work on this at all?

idigra: Thanks for pinging. I actually found another ways to do what I needed then but maybe I'll find some time of my free time to work on this in the future.

stakx: @idigra: No worries. :) Should we close this PR here in the meantime?

idigra: No problem, closed.

Issue #414, opened 2017-07-11
The use of `Action<...>` instead of `Expression<Action<...>>` in the public API causes a variety of problems


stakx:
`Action<TMock, ...>`s are used instead of `Expression<Action<TMock, ...>>` because the compiler doesn't support e.g. assignments in expression tree lambdas. Moq solves this by then executing the setup action "expression" in some kind of dry-run mode and observing the side effects caused. The internal component used for this trickery is called `FluentMockContext`, and because it's basically an incomplete workaround around a missing language feature, it causes a variety of problems.

<strike>Many of them are related to recursive mocks setup: [...] At least in some cases, it seems probable that these issues could be caused by the (internal) use of `FluentMockContext` to execute the setup "expression" in order to install the inner mocks relationships and set up all properties on the inner mocks.</strike> (These are actually unrelated to this and are now being tracked in #643.)

<strike>In order for the setup expression to be executable, its expression tree must be compiled (which btw. in #188 is given as the main reason for Moq's bad performance on 64-bit .NET). Compiling an expression tree to a regular `Func<>` or `Action<>` also means that matchers cannot properly work.</strike> (Don't know what I was thinking, this is a non sequitur.)

The use of `FluentMockContext` also causes the following problem(s) which aren't about recursive mocks:

* [ ] #218 - SetupSet problems with It.IsAny<> and indexer properties 
* [ ] #430 - VerifySet fails on non-trivial property setter
* [ ] #432 - Use of SetupSet 'forgets' Method Setup
* [ ] #456 - Heisenberg's virtual property set in ctor with MockBehavior.Strict and VerifyAll

I'm closing the above issues in favor of this one.

It might be a good idea to replace the use of `FluentMockContext` to "dry-run" setup expressions and instead do proper analysis of the expression trees, wherever possible. The potential advantages are:

* Make matchers (`It.Is…`) work.
* Avoid setting up more properties than necessary.
* Avoid overriding existing setups.
* Improve Moq performance by avoiding expensive expression tree compilation.

stakx:
> It might be a good idea to replace the use of FluentMockContext to "dry-run" setup expressions and instead do proper analysis of the expression trees, wherever possible. The potential advantages are:
>
> * Make matchers (It.Is…) work.
> * [...]
> * Improve Moq performance by avoiding expensive expression tree compilation.

[...]

Issue #218, opened 2015-11-22
SetupSet problems with It.IsAny<> and indexer properties


misulo:
Hi all 

The topic is already discussed on Stack Overflow: http://stackoverflow.com/questions/29697406/moq-an-indexed-property-and-use-the-index-value-in-the-return-callback

I ran into it also and then looking if an issue is posted, couldn't find any. So thought I'd post one.

In a nutshell:

```csharp
// This compiles, but callback never seems to be called
mock.SetupSet(m => m[It.IsAny<int>()] = It.IsAny<int>()).Callback(DoSet);
```

As a matter of fact, forget the Callback. Turn MockBehaviour to Strict, and try to mock the indexer Setter with a `It.IsAny<>`. It will not work.

```csharp
mock.SetupSet(m => m[It.IsAny<int>()] = It.IsAny<int>());
```

So right now the only workaround is not using the Strict behavior. 

Same works without problems with SetupGet.

regards, 
Misulo.

loedeman:
I am experiencing the exact same issue. It would be nice when MockBehaviour.Strict and It.IsAny<T> would work, fortunately in most of those cases omitting the mock behaviour does the trick. Is there any progress or are there any plans on this issue?

stakx:
I was looking at this a few days ago. IIRC there was a good reason why this doesn't work ATM, indexers and expression trees aren't too good friends and there's even a disabled unit test. I'll take a closer look in the next few days.

stakx:
PS: Feel free to do the same. If you find a solution you're welcome to send a PR!

stakx:
Here's a brief analysis of *why* this doesn't work:

C# currently cannot convert a lambda involving an assignment to an expression tree. That is, the following won't compile:

    ```csharp
    Expression<Action<string>> expr = str => str = null; 
    //                                       ^^^^^^^^^^
    // CS0832: An expression tree may not contain an assignment operator
    ```

[...]

PR #694, opened 2018-09-18
Support indexer with multiple arguments

idigra:
This PR contains support in multiple arguments.

1. Fix null reference exception in case of multiple arguments indexer.
2. Proposal for new API method SetupSetMethod which allows using It.IsAny for the indexer arguments (can be used also for properties, although existing API already covers properties). This is only in proposal level, should be enhanced (documentation, overloading for any count of indexer arguments, exceptions on illegal input).

### DIFF
--- a/src/Moq/Mock.Generic.cs
+++ b/src/Moq/Mock.Generic.cs
@@ -344,12 +344,6 @@ namespace Moq
 			return Mock.SetupSet<T>(this, setterExpression, null);
 		}
 
-		/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.SetupIndexerSet"]/*'/>
-		public ISetup<T> SetupSetMethod<T1, T2, TProperty>(Expression<Action<Action<T1, T2, TProperty>>> setter)
-		{
-			return Mock.SetupSetMethod<T, TProperty>(this, new Type[] { typeof(T1), typeof(T2) }, setter);
-		}
-
 		/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.SetupProperty(property)"]/*'/>
 		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By design")]
 		[SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Property", Justification = "This sets properties, so it's appropriate.")]
--- a/src/Moq/Mock.Generic.cs
+++ b/src/Moq/Mock.Generic.cs
@@ -344,6 +344,12 @@ namespace Moq
 			return Mock.SetupSet<T>(this, setterExpression, null);
 		}
 
+		/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.SetupIndexerSet"]/*'/>
+		public ISetup<T> SetupSetMethod<T1, T2, TProperty>(Expression<Action<Action<T1, T2, TProperty>>> setter)
+		{
+			return Mock.SetupSetMethod<T, TProperty>(this, new Type[] { typeof(T1), typeof(T2) }, setter);
+		}
+
 		/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.SetupProperty(property)"]/*'/>
 		[SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "By design")]
 		[SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Property", Justification = "This sets properties, so it's appropriate.")]
--- a/src/Moq/Mock.cs
+++ b/src/Moq/Mock.cs
@@ -664,13 +664,12 @@ namespace Moq
 
 					matchers[valueIndex] = new MatchExpression(last.Match);
 
-					if (arguments.Length == 2)
+					for (int i = 0; i < arguments.Length - 1; i++)
 					{
-						// TODO: what about multi-index setters?
 						// Add the index value for the property indexer
-						values[0] = GetValueExpression(arguments[0], parameters[0].ParameterType);
+						values[i] = GetValueExpression(arguments[i], parameters[i].ParameterType);
 						// TODO: No matcher supported now for the index
-						matchers[0] = values[0];
+						matchers[i] = values[i];
 					}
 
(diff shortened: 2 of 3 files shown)


stakx:
Hi @idigra, and thanks for this one, too. Same as with the other one, I'll try to find some time for reviewing this sometime during the next few days.

I can say right now, however, that it shouldn't be necessary to introduce a `SetupSetMethod`. You'll find that the existing `Setup`, `SetupGet`, `SetupSet` etc. methods already try to recognize when they are called on indexers&mdash;indexers shouldn't have their own dedicated methods, from a user's perspective that doesn't make much sense. (`SetupGet` and `SetupSet` are probably already confusing enough.)

idigra:
I now see that the second issue I opened (https://github.com/moq/moq4/issues/696, mentioned above) is a duplicate of https://github.com/moq/moq4/issues/218 which is closed in favor of https://github.com/moq/moq4/issues/414. I couldn't find the reason why the SetupSet overload you mentioned was deprecated. Also consider the solution I proposed in my issue https://github.com/moq/moq4/issues/696.

stakx:
> I couldn't find the reason why the SetupSet overload you mentioned was deprecated.

I don't know the exact reason, either. I am guessing that this decision was made simply because being able to write `SetupSet(m => m.Something = value)` is nicer than `SetupSet(m => m.Something, () => value)`.

src/Moq/Mock.cs
```diff
@@ -664,13 +664,12 @@ private static VoidSetupPhrase<T> SetupSetPexProtected<T>(Mock<T> mock, Action<T
 
 					matchers[valueIndex] = new MatchExpression(last.Match);
 
-					if (arguments.Length == 2)
+					for (int i = 0; i < arguments.Length - 1; i++)
```
stakx: I think it would be clearer to compare `i < valueIndex` to emphasize the fact that we're excluding the `value` parameter.

tests/Moq.Tests/CallbacksFixture.cs
```diff
@@ -474,6 +474,105 @@ public void CallbackWithRefParameterCannotModifyNonRefParameter()
 			Assert.Equal("input", value);
 		}
 
+		[Fact]
+		public void CallbackWithIndexerGetter()
```
stakx: I appreciate this test, and it would be good to have one like it (if we don't have one like it yet, I didn't double-check TBH)... but could you explain how it is related to this PR?

If it isn't directly related to this PR and you simply added it for completeness' sake, then I'd appreciate if you submitted it in a separate PR.

tests/Moq.Tests/CallbacksFixture.cs
```diff
@@ -474,6 +474,105 @@ public void CallbackWithRefParameterCannotModifyNonRefParameter()
...
+		}
+
+		[Fact]
+		public void CallbackWithMultipleArgumentIndexerGetter()
+		{
+			int x = default(int);
+			int y = default(int);
+
+			var mock = new Mock<IFoo>();
+			mock.Setup(f => f[It.IsAny<int>(), It.IsAny<int>()])
```
stakx: Same as in an earlier review comment, I appreciate the test, but it doesn't seem to be related to this PR, since it uses the expression tree-based `Setup` instead of the delegate-based `SetupSet`...?

Could you state how it relates to the bugfix in your PR (and if it doesn't, could you remove it and optionally re-submit in a separate PR?)

tests/Moq.Tests/CallbacksFixture.cs
```diff
@@ -474,6 +474,105 @@ public void CallbackWithRefParameterCannotModifyNonRefParameter()
...
+				}));
+
+			mock.Object[3, 13] = 2;
+			Assert.Equal(3, x);
+			Assert.Equal(13, y);
+			Assert.Equal(2, result);
+		}
+
+		[Fact]
+		public void CallbackWithMultipleArgumentIndexerSetterWithAny()
```
stakx: Please remove along with the rest of `SetupSetMethod`-related functionality.

tests/Moq.Tests/CallbacksFixture.cs
```diff
@@ -490,7 +589,7 @@ public class Derived : Base
 		private void TraceMe(Base b)
 		{
 		}
-		
+
```
stakx: It's a tiny detail, I know, but could you please revert this whitespace-only change?

src/Moq/Mock.Generic.cs
```diff
@@ -344,6 +344,12 @@ public ISetup<T> SetupSet(Action<T> setterExpression)
 			return Mock.SetupSet<T>(this, setterExpression, null);
 		}
 
+		/// <include file='Mock.Generic.xdoc' path='docs/doc[@for="Mock{T}.SetupIndexerSet"]/*'/>
+		public ISetup<T> SetupSetMethod<T1, T2, TProperty>(Expression<Action<Action<T1, T2, TProperty>>> setter)
```
stakx: Please remove.

src/Moq/Mock.cs
```diff
@@ -683,6 +682,28 @@ private static VoidSetupPhrase<T> SetupSetPexProtected<T>(Mock<T> mock, Action<T
 			}
 		}
 
+		internal static ISetup<T> SetupSetMethod<T, TProperty>(
```
stakx: Please remove.

tests/Moq.Tests/CallbacksFixture.cs
idigra: Automatically done by VS. I reverted it manually using notepad++, hope that now it's OK.

src/Moq/Mock.cs
idigra: I tried to be consistent with the for loop that is some lines up (line 635).
Let me know if you prefer to (1) leave it as I've done, for maximal consistency between the two code lines AND minimum changes, (2) Fix both for loops, for maximal consistency AND answer your request, or (3) Fix only mine, so at least new code will answer your request (in cost of consistency).

stakx: Fair enough. Never mind for the moment, let's leave it as you put it.

stakx:
@idigra, Moq 4.10.1, which includes your bugfix, has just been published on NuGet.

idigra:
Thanks. I finally came to verify it, but then I understand that without extending the API to support Any matcher in indexer argument I can't really reach a scenario in which the bug was reproduced.
I would be happy to check it if you have a real scenario in which the bug occurs.

idigra:
https://github.com/moq/moq4/issues/696

PR #704, opened 2018-10-11
Fixed issue #314 

Damian-Pumar:
- Added .bat compilation, because vs 2017 have many problems to build sln.

- Fixed issue #314 

- Added unit test of example explained in #314


build_all_vs2017.bat
```diff
@@ -0,0 +1,30 @@
+@echo off
```
stakx: Please remove this batch file and the `vswhere.exe` binary, for two reasons: (1) I recently took pains to get rid of unnecessary build scripts, I wouldn't want to go back to a batch file. (2) Fixing the build is unrelated to the issue you're trying to solve.

If you have problems building Moq, then please raise a new issue so we can look at it.

src/Moq/Extensions.cs
```diff
@@ -91,7 +96,7 @@ public static void ThrowIfNotMockeable(this MemberExpression memberAccess)
 
 		public static bool IsMockeable(this Type typeToMock)
 		{
-			// A value type does not match any of these three 
+			// A value type does not match any of these three
```
stakx: Please revert this white space-only change.

src/Moq/Extensions.cs
```diff
@@ -293,4 +298,4 @@ private static void AddPropertiesInDepthFirstOrderTo(this Type type, List<Proper
 			}
 		}
 	}
-}
+}
```
stakx: Please revert this line separator-only change.

src/Moq/FluentMockVisitor.cs
```diff
@@ -100,10 +100,10 @@ private MethodInfo GetTargetMethod(Type objectType, Type returnType)
...
 				//.Setup(mock => mock.Solution)
 				return typeof(Mock<>)
-				       .MakeGenericType(objectType)
-				       .GetMethods("Setup")
-				       .First(m => m.IsGenericMethod)
-				       .MakeGenericMethod(returnType);
+					   .MakeGenericType(objectType)
+					   .GetMethods("Setup")
+					   .First(m => m.IsGenericMethod)
+					   .MakeGenericMethod(returnType);
```
stakx: Please revert this white space-only change.

src/Moq/FluentMockVisitor.cs
```diff
@@ -115,11 +115,11 @@ private MethodInfo GetTargetMethod(Type objectType, Type returnType)
...
-		                                   Type returnType,
-		                                   MethodInfo targetMethod,
-		                                   Expression instance,
-		                                   ParameterExpression lambdaParam,
-		                                   Expression lambdaBody)
+										   Type returnType,
+										   MethodInfo targetMethod,
+										   Expression instance,
+										   ParameterExpression lambdaParam,
+										   Expression lambdaBody)
```
stakx: Please revert this white space-only change.

src/Moq/FluentMockVisitor.cs
```diff
@@ -191,4 +193,4 @@ private MethodInfo GetTargetMethod(Type objectType, Type returnType)
 			return (Mock<TResult>)fluentMock;
 		}
 	}
-}
+}
```
stakx: Please revert this line separator-only change.

tests/Moq.Tests/Regressions/FluentMockVisitorIssues.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Xunit;
```
stakx: Right now, we have one single file for all regression tests coming from GitHub issues; `IssueReportsFixture.cs`. Usually, there's a `#region <IssueNumber>" for every issue (they are sorted by source & issue number), and in that region there's a `public class Issue<IssueNumber>` containing all tests related to that issue.

Could you please move these tests there? Thanks!

(Btw., the following hint is not relevant if you move your tests to `IssueReportsFixture.cs`, but if you create a new code file anywhere in the Moq solution, you should also add the copyright notice at the very beginning of the new file. You can usually just copy & paste it from any other code file:)

```diff
+// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
```

tests/Moq.Tests/Regressions/FluentMockVisitorIssues.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Moq.Tests.Regressions
+{
+	public class FluentMockVisitorIssues
+	{
+		[Fact]
+		public void WhenQueryingWithTwoIndexes_ThenSetsThemDirectly1()
+		{
+			var foo = Mock.Of<IFoo>(x => x[0].Value == "hello" && x[1].Value == "goodbye");
+
+			Assert.Equal("hello", foo[0].Value); // Fails here as foo[0] is the same object as foo[1] and foo[1].Value == "goodbye", Fixed!
```
stakx: The _"fails here"_ comment will no longer make sense once you've fixed the issue. Please remove it. (With all tests, the basic assumption is that they will pass, so there should be no comment saying _"this fails"_. ;-)

tests/Moq.Tests/Regressions/FluentMockVisitorIssues.cs
```diff
@@ -0,0 +1,35 @@
...
+			Assert.Equal("goodbye", foo[1].Value);
+		}
+
+		[Fact]
+		public void WhenQueryingWithTwoIndexes_ThenSetsThemDirectly2()
+		{
+			var foo = Mock.Of<IFoo>(x => x[0] == Mock.Of<IBar>(b => b.Value == "hello") && x[1] == Mock.Of<IBar>(b => b.Value == "goodbye"));
+
+			Assert.Equal("hello", foo[0].Value); // These pass no problem
```
stakx: Similar as in the above comment, there is no need to say _"these pass no problem"_ &mdash; that is what we assume for all tests, anyway.

(I know you copy & pasted these tests from the issue, and that is perfectly fine. Just pointing out that they often need to be changed a tiny bit so that they make sense in their new context.)

stakx:
Closing this (as per Gitter chat between the PR author and me). We'll reopen this once the restrictive `InnerMocks` mechanism has been replaced with something more solid.

stakx:
@damianpumar - After your Gitter ping I've quickly gone through your previous contributions (this one) to see if there's anything we could pick up again, and then I saw that this issue has since been fixed as a direct consequence of replacing the old `InnerMocks` machinery. I'm therefore removing the _unresolved_ label.

Issue #706, opened 2018-10-13
`CallBase` should not be allowed for delegate mocks


stakx:
### Reproduction code:

```csharp
var mock = new Mock<Action>();
mock.Setup(m => m()).CallBase();
mock.Object.Invoke();
```

### Actual behavior:

`mock.Object.Invoke()` throws a `NotImplementedException` with message:

```
This is a DynamicProxy2 error: The interceptor attempted to 'Proceed' for method 'Void Invoke()' which has no target. When calling method without target there is no implementation to 'proceed' to and it is the responsibility of the interceptor to mimic the implementation (set return value, out arguments etc)
```

### Expected behavior:

The call to `.CallBase()` should throw a `NotSupportedException` with a message explaining that `CallBase` cannot be used with delegate mocks.

The same should happen when setting a delegate mock's `CallBase` property to `true`.

tehmantra:
Hey, created PR #708

stakx:
@tehmantra, thank you for picking this one up! :+1: I've posted a review on your PR.

Veikedo:
It looks like something is wrong with this issue.

I have this code

    [Theory]
    [AutoMoqData]
    internal async Task GetAutomationStatistics_should_return_correct_statistics(
      [Frozen] Mock<GetAutomationsLogTotalCount> getAutomationsLogTotalCount,
      [Frozen] GetCurrentlyRunningSchedulesResult getCurrentlyRunningSchedulesResult,
      [Frozen] GetAutomationsTotalCountResult getAutomationsTotalCountResult,
      GetAutomationStatisticsQuery query,
      GetAutomationStatisticsHandler sut,
      CancellationToken ct
    )
    {
      getAutomationsLogTotalCount
        .Setup(_ => _(It.IsAny<GetAutomationsLogTotalCountQuery>(), ct))
        .ReturnsAsync((GetAutomationsLogTotalCountQuery x, CancellationToken _) =>
          new GetAutomationsLogTotalCountResult
          {
            TotalCount = (long) x.LogLevel.GetValueOrDefault()
          }
        );

      var expectedResult = new GetAutomationStatisticsResult
      {
        AutomationsRunningNow = await getCurrentlyRunningSchedulesResult.Items.LongCount(ct),
        TotalAutomations = getAutomationsTotalCountResult.TotalCount,
        ErrorsReported = (long) EapLogLevel.Error,
        WarningsReported = (long) EapLogLevel.Warn,
      };

[...]

PR #708, opened 2018-10-15
`CallBase` should not be allowed for delegate mocks

tehmantra:
Fixes: #706 

First time looking at this codebase, I've tried to follow the style as best as I could.
I added tests that capture the issue, hopefully I have understood it correctly.

Cheers


src/Moq/MethodCall.cs
```diff
@@ -123,6 +123,11 @@ public override void Execute(Invocation invocation)
 
 		public virtual void SetCallBaseResponse()
 		{
+			if (typeof(Delegate).IsAssignableFrom(this.Mock.TargetType))
```
stakx: Moq defines an [extension method `Type.IsDelegate()`](https://github.com/moq/moq4/blob/2c241fbd4599da0069668d07f69441d66a9f8687/src/Moq/Extensions.cs#L75-L81), could you please use it here?

src/Moq/Mock.Generic.cs
```diff
@@ -164,7 +166,15 @@ private void CheckParameters()
 		public override bool CallBase
 		{
 			get => this.callBase;
-			set => this.callBase = value;
+			set
+			{
+				if (typeof(Delegate).IsAssignableFrom(this.MockedType))
```
stakx: (As above:) Moq defines an [extension method `Type.IsDelegate()`](https://github.com/moq/moq4/blob/2c241fbd4599da0069668d07f69441d66a9f8687/src/Moq/Extensions.cs#L75-L81), could you please use it here?

src/Moq/Mock.Generic.cs
```diff
@@ -164,7 +166,15 @@ private void CheckParameters()
...
+				{
+					throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, Resources.CallBaseUsedOnDelegateException));
```
stakx: It is important that an exception be thrown only if `value == true`. This is because when you have a mock with `DefaultValue.Mock`, and Moq auto-mocks something (e.g. the delegate mock's return value), it'll attempt to set `CallBase`.

src/Moq/Properties/Resources.Designer.cs
```diff
@@ -87,6 +87,15 @@ internal class Resources {
 			}
 		}
 		
+		/// <summary>
+        ///   Looks up a localized string similar to CallBase cannot be used to mock Delegate types..
+        /// </summary>
```
stakx: There appears to be something wrong with indentation here, could you please take a look & fix this?

src/Moq/Properties/Resources.resx
```diff
@@ -348,4 +348,7 @@ Expected invocation on the mock once, but was {4} times: {1}</value>
 		<value>The following invocations on mock '{0}' were not verified:
 {1}</value>
 	</data>
-</root>
\ No newline at end of file
+	<data name="CallBaseUsedOnDelegateException" xml:space="preserve">
+    <value>CallBase cannot be used to mock Delegate types.</value>
```
stakx: I'd suggest a slight rephrasing here:

```diff
-CallBase cannot be used to mock Delegate types.
+CallBase cannot be used with Delegate mocks.
```

(The reason being that `CallBase` isn't the party that mocks something; the delegate mock exists already by the time `CallBase` is being set.)

src/Moq/Properties/Resources.resx
```diff
@@ -348,4 +348,7 @@ Expected invocation on the mock once, but was {4} times: {1}</value>
 		<value>The following invocations on mock '{0}' were not verified:
 {1}</value>
 	</data>
-</root>
\ No newline at end of file
+	<data name="CallBaseUsedOnDelegateException" xml:space="preserve">
```
stakx: Minor suggestion: Rename `CallBaseUsedOnDelegateException` to `CallBaseCannotBeUsedWithDelegateMocks`. (The resource names usually reflect its contents, not what it will be used for.)

tests/Moq.Tests/Regressions/IssueReportsFixture.cs
```diff
@@ -2010,6 +2011,32 @@ public interface IMyInterface
...
+			}
+
+			[Fact]
+			public void CallBase_property_should_not_be_allowed_for_delegate_mocks()
+			{
+				Mock<Action> mock = new Mock<Action>();
+
+				Exception ex = Assert.Throws<NotSupportedException>(() => mock.CallBase = true);
+				Assert.Equal(string.Format(CultureInfo.CurrentCulture, Resources.CallBaseUsedOnDelegateException), ex.Message);
+			}
```
stakx: Nice! :+1: Could you please add a third test, verifying that `mock.CallBase = false` does *not* throw? Thank you!

tehmantra:
All comments resolved. Cheers.

tehmantra:
Done. Missed that this has 2 implementations! Cheers.

stakx:
Awesome! Thank you for contributing!

stakx:
@tehmantra - Moq 4.10.1, which includes your bugfix, has just been published on NuGet.

PR #712, opened 2018-10-28
Add tests for Moq compatibility with F# events

stakx:
There are two old Google Code issues which were resolved with commit 44070a90, however no documentation or unit tests were ever added for these issues:

 * **Moq fails to mock events defined in F#:**  
   https://code.google.com/archive/p/moq/issues/238

 * **Can't raise events on mocked Interop interfaces:**  
   https://code.google.com/archive/p/moq/issues/226

This commit adds unit tests for the first of these two issues by adding an F# test project to the solution, which allows us to define F# types for use in unit tests.

Some of these tests are skipped on purpose; see https://github.com/Microsoft/visualfsharp/issues/5834.

### DIFF
--- a/tests/Moq.Tests.FSharpTypes/HasActionEvent.fs
+++ b/tests/Moq.Tests.FSharpTypes/HasActionEvent.fs
@@ -1,0 +1,12 @@
+// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
+namespace Moq.Tests.FSharpTypes
+
+open System;
+
+type HasActionEvent() =
+    let event = new DelegateEvent<Action>();
+
+    [<CLIEvent>]
+    member this.Event = event.Publish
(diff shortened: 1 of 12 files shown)


tests/Moq.Tests.FSharpTypes/HasActionEvent.fs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Moq.Tests.FSharpTypes
+
+open System;
+
+type HasActionEvent() =
+    let event = new DelegateEvent<Action>();
+
+    [<CLIEvent>]
+    member this.Event = event.Publish
```
stakx: This type and event will probably be quite useless for other Moq unit tests, since `Event` and its accessor methods won't be defined `virtual` so Moq cannot intercept them.

tests/Moq.Tests.FSharpTypes/Moq.Tests.FSharpTypes.fsproj
```diff
@@ -0,0 +1,23 @@
...
+
+	<ItemGroup>
+		<Compile Include="HasAbstractActionEvent.fs" />
+		<Compile Include="HasAbstractEventHandlerEvent.fs" />
+		<Compile Include="HasActionEvent.fs" />
+		<Compile Include="IHasActionEvent.fs" />
+		<Compile Include="IHasEventHandlerEvent.fs" />
+	</ItemGroup>
+
+	<Target Name="Test" DependsOnTargets="Build" />
```
stakx: This is necessary for the build to succeed, because this is in the `tests/` directory, but not a test project itself.

kzu:
:+1:!

PR #716, opened 2018-11-04
Make internal `Mock` methods non-generic

stakx:
This is a continuation of #663. Internal worker methods in the `Mock` class are made non-generic. The idea here is to restrict the use of generics exclusively to those places where they are useful: in the methods making up the public API. Underneath that layer, Moq should use generics as little as possible in order to avoid needless code duplication, and to make reflection easier.

### DIFF
--- a/src/Moq/Mock.cs
+++ b/src/Moq/Mock.cs
@@ -521,10 +521,7 @@ namespace Moq
 			return setup;
 		}
 
-		internal static SetterSetupPhrase<T, TProperty> SetupSet<T, TProperty>(
-			Mock<T> mock,
-			Expression<Func<T, TProperty>> expression)
-			where T : class
+		internal static MethodCall SetupSet(Mock mock, LambdaExpression expression, Expression valueExpression)
 		{
 			var prop = expression.ToPropertyInfo();
 			Guard.CanWrite(prop);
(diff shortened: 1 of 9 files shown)


src/Moq/Mock.cs
```diff
@@ -566,14 +566,11 @@ private static VoidSetupPhrase<T> SetupSetPexProtected<T>(Mock<T> mock, Action<T
...
-		private static (Mock, LambdaExpression, MethodInfo, Expression[]) SetupSetImpl<T>(
-			Mock<T> mock,
-			Action<T> setterExpression)
-			where T : class
+		private static (Mock, LambdaExpression, MethodInfo, Expression[]) SetupSetImpl(Mock mock, Delegate setterExpression)
 		{
 			using (var context = new FluentMockContext())
 			{
-				setterExpression(mock.Object);
+				setterExpression.DynamicInvoke(mock.Object);
```
stakx: There would be ways around `DynamicInvoke` but they are probably not worth the effort. (For example, we could use an `Action<object>` instead of a `Delegate`, then convert the parameter between `T` and `object` in the calling methods. However, this would also require capturing the original delegate's parameter name and passing it into `SetupSetImpl` as well; c.f. the lines of code below this one.)

src/Moq/Mock.cs
```diff
@@ -508,42 +508,17 @@ private static VoidSetupPhrase<T> SetupPexProtected<T>(Mock<T> mock, Expression<
...
 
-		private static SetterSetupPhrase<T, TProperty> SetupSetPexProtected<T, TProperty>(
-			Mock<T> mock,
-			Action<T> setterExpression,
-			Condition condition)
-			where T : class
+		private static MethodCall SetupSetPexProtected(Mock mock, Delegate setterExpression, Condition condition)
 		{
 			var (m, expr, method, value) = SetupSetImpl(mock, setterExpression);
-			Debug.Assert(value.Length == 1);
```
stakx: If we'd had a unit test calling this method for an indexer set expression, we'd have seen this assertion fail. Let's remove it.

Issue #719, opened 2018-11-06
`net45` target causes major build issues in .NET Framework 4.7 projects


bassettb:
Moq should include a build target for .Net Framework 4.7.  This target would not have a dependency on System.ValueTuple.  The current 4.5 target causes major build issues in Framework 4.7 projects.

stakx:
> The current 4.5 target causes major build issues in Framework 4.7 projects.

Interesting that noone reported any major issues up until now. Could you please explain in some more detail just what those issues are?

(P.S.: We've already had several issues asking for certain targets and it's being considered, so I will likely close this issue as a duplicate.)

bassettb:
The warning is below.  It could not be resolved with a binding redirect or by adding the latest System.ValueTuple nuget to our projects.   `warning MSB3258: The primary reference "myproject.dll" could not be resolved because it has an indirect dependency on the .NET Framework assembly "System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" which has a higher version "4.0.3.0" than the version "4.0.2.0" in the current target framework.`

stakx:
Must be something with your specific setup. Make sure you have the latest versions of everything (VS, NuGet, .NET Framework). It might also help to use `<PackageReference>` instead of `packages.config`. If you're using the old `.csproj` format, make sure you use `<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>`.

Finally, please note that this type of problem cannot be fixed for everyone on Moq's side. It has to be fixed on the .NET & tooling side. The [`netfx-compat`-tagged issues in `dotnet/standard`](https://github.com/dotnet/standard/issues?q=is%3Aissue+is%3Aopen+label%3Anetfx-compat) are a testament to how widespread assembly versioning problems have become on the .NET Framework. [This post by Immo Landwerth (MSFT)](https://github.com/dotnet/standard/issues/481) is also a nice write-up although not directly related to `net45` vs `net47`.

Btw. here's a small C# project which compiles just fine (using VS 15.8.8 or MSBuild 15.8.169.51996) , without any warnings or errors. Feel free to use it as a starting point to track down & fix your versioning issue.

#### `.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net47</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
[...]

PR #721, opened 2018-11-08
Drop `System.ValueTuple` NuGet dependency

stakx:
People are still experiencing assembly versioning problems on the .NET Framework because of the NuGet packages `System.Threading.Tasks.Extensions` and `System.ValueTuple` referenced by Moq. (See e.g. the recent #719.)

It is fairly trivial to get rid of the latter dependency... so let's get rid of it! (It means we can no longer use tuple syntax everywhere in Moq's code base, but as we haven't made excessive use of it so far, that's perhaps a small price to pay.)

### DIFF
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -141,7 +141,7 @@ namespace Moq
 		/// around either a <see cref="MethodCallExpression" /> (for a normal method call)
 		/// or a <see cref="InvocationExpression" /> (for a delegate invocation).
 		/// </summary>
-		internal static (Expression Object, MethodInfo Method, IReadOnlyList<Expression> Arguments) GetCallInfo(this LambdaExpression expression, Mock mock)
+		internal static CallInfo GetCallInfo(this LambdaExpression expression, Mock mock)
 		{
 			Guard.NotNull(expression, nameof(expression));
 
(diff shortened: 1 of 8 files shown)


src/Moq/LookupOrFallbackDefaultValueProvider.cs
```diff
@@ -32,26 +32,26 @@ namespace Moq
...
-				[typeof(ValueTuple<,,,,,,>)] = CreateValueTupleOf,
-				[typeof(ValueTuple<,,,,,,,>)] = CreateValueTupleOf,
+				["System.ValueTuple`1"] = CreateValueTupleOf,
+				["System.ValueTuple`2"] = CreateValueTupleOf,
+				["System.ValueTuple`3"] = CreateValueTupleOf,
+				["System.ValueTuple`4"] = CreateValueTupleOf,
+				["System.ValueTuple`5"] = CreateValueTupleOf,
+				["System.ValueTuple`6"] = CreateValueTupleOf,
+				["System.ValueTuple`7"] = CreateValueTupleOf,
+				["System.ValueTuple`8"] = CreateValueTupleOf,
```
stakx: What this means is that Moq now recognises `System.ValueTuple<…>` types by name instead of by strict type identity. This mimicks what the C# compiler already does.

(Unlike the C# compiler, Moq however won't validate the structural makeup of the matched types, so Moq's default value providers will not work correctly with such user types that don't have the expected constructor signature.)

src/Moq/LookupOrFallbackDefaultValueProvider.cs
```diff
@@ -122,7 +122,9 @@ protected internal sealed override object GetDefaultValue(Type type, Mock mock)
 			               : type.IsArray ? typeof(Array)
 			               : type;
 
-			return this.factories.TryGetValue(handlerKey, out Func<Type, Mock, object> factory) ? factory.Invoke(type, mock)
+			Func<Type, Mock, object> factory;
+			return this.factories.TryGetValue(handlerKey         , out factory) ? factory.Invoke(type, mock)
+			     : this.factories.TryGetValue(handlerKey.FullName, out factory) ? factory.Invoke(type, mock)
 			     : this.GetFallbackDefaultValue(type, mock);
```
stakx: The default value providers will now do two dictionary lookups instead of one in many cases. In theory, that's a slight performance degradation, but it would be surprising if this mattered much in practice. (A dictionary lookup is likely nothing when compared to all the reflection and code emitting going on elsewhere.)

src/Moq/ExpressionExtensions.cs
```diff
@@ -169,4 +169,25 @@ internal static (Expression Object, MethodInfo Method, IReadOnlyList<Expression>
 				expression.ToStringFixed()));
 		}
 	}
+
+	internal readonly struct CallInfo
```
stakx: We *could* declare a few generic tuple types instead of this and other concrete `readonly struct`s. For now, that degree of reusability isn't needed, and we spare the JITter some additional work by avoiding generics. (And, perhaps more importantly, declaring concrete, non-generic types gives us a means of giving the constituent parts names more meaningful than just `Item1`, `Item2`, etc.)

src/Moq/ExpressionExtensions.cs
```diff
@@ -169,4 +169,25 @@ internal static (Expression Object, MethodInfo Method, IReadOnlyList<Expression>
...
+		private readonly IReadOnlyList<Expression> arguments;
+
+		public CallInfo(Expression expression, MethodInfo method, IReadOnlyList<Expression> arguments)
+		{
+			this.expression = expression;
+			this.method = method;
+			this.arguments = arguments;
+		}
+
+		public void Deconstruct(out Expression expression, out MethodInfo method, out IReadOnlyList<Expression> arguments)
```
stakx: Using a `Deconstruct` instead of just making the fields `public` means that we can still use tuple syntax / deconstruction in assignments.

PR #722, opened 2018-11-11
Enable `FluentMockContext` to record >1 matcher per invocation

stakx:
Moq currently cannot correctly deal with setter setups using more one matcher, e.g. when setting up an indexer's setter. This is because the `FluentMockContext` only associates at most one matcher with observed invocations.

This PR changes how `FluentMockContext` records its observations and how it lets the outside world access them. When asked for the last observed mock invocation, `FluentMockContext` now returns *all*
matchers observed immediately before the final invocation.

The class is also renamed to `AmbientObserver` (way too many things in Moq's code base are called "fluent") and documented somewhat better using XML documentation comments and tests.

### DIFF
--- a/src/Moq/AmbientObserver.cs
+++ b/src/Moq/AmbientObserver.cs
@@ -3,35 +3,46 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Moq
 {
 	/// <summary>
-	/// Tracks the current mock and interception context.
+	///   A per-thread observer that records invocations to mocks and matchers for later inspection.
 	/// </summary>
-	internal class FluentMockContext : IDisposable
+	/// <remarks>
+	///   <para>
+	///     This component requires the active cooperation of the respective subsystems.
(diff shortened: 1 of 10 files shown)


src/Moq/AmbientObserver.cs
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		/// <summary>
+		///   Checks whether the last observed thing was a mock invocation.
+		///   If <see langword="true"/>, details about that mock invocation are provided via the <see langword="out"/> parameters.
+		/// </summary>
+		/// <param name="mock">The <see cref="Mock"/> on which an invocation was observed.</param>
+		/// <param name="invocation">The observed <see cref="Invocation"/>.</param>
+		/// <param name="matches">The <see cref="Match"/>es that were observed just before the invocation.</param>
+		public bool LastObservationWasMockInvocation(out Mock mock, out Invocation invocation, out Matches matches)
```
stakx: Name mismatch, `Invocation` in `OnInvocation` vs `MockInvocation` in `LastObservationWasMockInvocation`. It would be good to decide on one naming scheme.

Issue #727, opened 2018-12-01
error messages don't display the parent class name of an inner class object


powerdude:
If i have classes like this:

```
public class Foo
{
   public class Bar
  {
 }
}

public class Foo1
{
   public class Bar{}
}
```

and try to verify a mock call:
```
mock.Verify( x=>x.Send(It.IsAny<Foo.Bar>());
```
and call the service `service.Send(new Foo1.Bar())`, the messages displayed fail to indicate the parent class name.
```
Expected invocation on the mock at least once, but was never performed: x=>x.Send(Bar)
```

Ideally, it should say:
```
Expected invocation on the mock at least once, but was never performed: x=>x.Send(Foo+Bar)
```

stakx:
@powerdude, thanks for reporting this. Would you like to submit a PR for this?

This could be quite easily done by changing Moq's [`StringBuilderExtensions.AppendNameOf(this StringBuilder, Type)` method](https://github.com/moq/moq4/blob/c5ffae8c2486b8244a81acb3a6abe2e1a7d555fe/src/Moq/StringBuilderExtensions.cs#L37-L68).

It just so happens that I wrote a small library called TypeNameFormatter not too long ago; see its [repo on GitHub](https://github.com/stakx/TypeNameFormatter), or the [source code-only package on NuGet.org](https://www.nuget.org/packages/TypeNameFormatter.Sources). The reason for a separate library is that formatting type names the way they look in e.g. C# turns out to be non-trivial due to a few gotchas (one of which you just described), so I figured let's encapsulate the logic for that in a library so we don't have to think too hard about it anymore.

I've simply not got around to applying that library to Moq yet, but `AppendNameOf` would become as simple as this:

```diff
+// <PackageReference Include="TypeNameFormatter.Sources" Version="1.0.0" PrivateAssets="All" />
+// using TypeNameFormatter;
+
 public static StringBuilder AppendNameOf(this StringBuilder stringBuilder, Type type)
 {
 	Debug.Assert(type != null);
 
+	return stringBuilder.AppendFormattedName(type);
-	var name = type.Name;
-	var backtickIndex = name.IndexOf('`');
-	if (backtickIndex >= 0)
-	{
-		stringBuilder.Append(name, 0, backtickIndex);
-	}
-	else
-	{
-		stringBuilder.Append(name);
-	}
-
-	if (type.GetTypeInfo().IsGenericType)
-	{
-		var genericArguments = type.GetGenericArguments();
[...]

PR #728, opened 2018-12-03
address #727

powerdude:
added TypeNameFormatter so that richer class names could be displayed in error messages.

### DIFF
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -39,6 +39,10 @@
 		<PackageReference Include="Castle.Core" Version="4.3.1" />
 		<PackageReference Include="IFluentInterface" Version="2.1.0" />
 		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.3.0" />
+		<PackageReference Include="TypeNameFormatter.Sources" Version="1.0.0">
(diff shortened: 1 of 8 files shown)


src/Moq/Moq.csproj
```diff
@@ -39,6 +39,10 @@
 		<PackageReference Include="Castle.Core" Version="4.3.1" />
 		<PackageReference Include="IFluentInterface" Version="2.1.0" />
 		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.3.0" />
+		<PackageReference Include="TypeNameFormatter.Sources" Version="1.0.0">
+		  <PrivateAssets>all</PrivateAssets>
+		  <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+		</PackageReference>
```
stakx: I think you could simplify this to just `<PackageReference Include="TypeNameFormatter.Sources" Version="1.0.0" PrivateAssets="All" />` (but I might be wrong).

powerdude: i didn't add this manually.  It was added by VS.

stakx:
@powerdude - Merged. Thank you for contributing! :+1:

Issue #749, opened 2019-02-14
Verify gets confused between the same generic and non-generic signature


lepijohnny:
Hi,

A few days ago I came across this [SO question](https://stackoverflow.com/questions/54637948/moq-verify-generic-method-tested-with-base-type-when-inherited-type-is-used). Although the question itself states something which is not correct I started investigating a bit and found one kinda strange thing which I want to share. Imagine this situation:

```
public interface IBazParam { }

public interface IBaz
{
	void Subscribe<T>() where T : IBazParam;
	void Subscribe();
}

//this test will actually pass
[Fact]
public void This_is_weird_test()
{
       //Arrange
       var mock = new Mock<IBaz>();

       //Act
       mock.Object.Subscribe<IBazParam>();

       //Assert
       mock.Verify(foo => foo.Subscribe(), Times.Once);
}
```

My actual question is why this test is passing, the method `Subcribe<T>()` is not the same as `Subscribe()`?
 
I also looked at the code and found potential bug within `InvocationShape.cs` class. During the matching of the invocation clearly `MethodInfo` of these methods are not equal so the `IsOverride` has to be called. `IsOverride` method will check if the verifying method is generic, in this case not as we are verifying `Subscribe` and it will fallback to parameter check which is in this case equal as `Subcribe<T>` and `Subcribe` do not have any parameters(in general it will be true if both method have the same set of parameters). IMO the check should be changed to _if any of these methods `IsGenericMethod`_, something like:

```
if (method.IsGenericMethod || invocationMethod.IsGenericMethod)
{
    ...
}
```

stakx:
@lepijohnny - Thank you for your report & analysis. This indeed looks like a bug, and your analysis of `InvocationShape.IsOverride` seems accurate. I think we should change this, if it doesn't break existing behaviour (tests). Would you like to submit a fix for this (including additional tests such as the above)?

lepijohnny:
**@stakx** - I already played around a bit and it seems the fix won't break any of the existing tests. I would be honored to contribute to this magnificent library so I would like to send PR with the fix (and tests of course).

stakx:
@lepijohnny - Please go ahead! Contributions are welcome! :+1:

stakx:
Fixed in #750.

PR #750, opened 2019-02-15
Verify gets confused between the same generic and non-generic signature

lepijohnny:
The fix for #749. I also added a few tests to verify the derived generic parameters verification. It is already existing functionality in moq but the tests were missing. IMO it cannot be harmful anyway to add more tests.


stakx:
All good. Thank you for contributing! :+1:

lepijohnny:
@stakx np, I will try to contribute more. You are doing great job guys 👍.

PR #755, opened 2019-02-21
Import Cleanup and Grouping for the test project

Shereef:


Shereef:
@stakx: I don't think this needs a changelog entry, does it ?

stakx:
Thank you for taking care of this cleanup. It is great to separate such reformattings into a separate PR to keep the actual PRs clean and focused. You've done exactly the right thing. :+1:

> I don't think this needs a changelog entry, does it ?

Right, no changelog entry is needed for this. Moq users typically won't care what the library's source code looks like. 😁

Issue #699, opened 2018-09-22
`Verify` can create setups that causes a subsequent `VerifyAll` to fail


stakx:
### Reproduction code:

```csharp
public interface IX
{
    IX X { get; }
}

var mock = new Mock<IX>();
mock.Verify(m => m.X.X, Times.Never);
mock.VerifyAll();
```

### Actual behavior:

`VerifyAll` throws a `MockException` with message:

```
The following setups on mock 'Mock<IX:00000001>' were not matched:
IX mock => mock.X
```

### Expected behavior:

All verification steps should pass because there are no invocations nor any setups.

lepijohnny:
There are more cases affected by this one as well. I would like to share them as they are, IMO, all related with the way how `FluentMockVisitor` unfold fluent expression into inner mocks with setup even though we actually do `Verify` but not `Setup`. It seems that `Verify` using fluent expression add (unexpected) side effects.

## 1. MochBehaviour.Strict has been avoided

### Reproduction code:

```
public interface IX
{
    IX X { get; }
}

var mock = new Mock<IX>(MockBehavior.Strict);
mock.Verify(m => m.X.X, Times.Never);
_ = mock.Object.X;
```

### Actual behaviour

No exception

### Expected behavior

`MockException` has been thrown as there is no setup for `m.X`

## 2. It is possible to invoke method\property without proper setup

### Reproduction code:

```
public interface IX
{
    IX X { get; }
}

var mock = new Mock<IX>();
mock.Verify(m => m.X.X, Times.Never);
_ = mock.Object.X.X;
```

### Actual behaviour

No exception

### Expected behavior

`NullReferenceException` has been thrown because we didn't setup return value for `IX` and it should be null.

lepijohnny:
@stakx - `FluentMockVisitor` translate the fluent expression the same way, it doesn't matter if it is `Setup` or `Verify`. It will always, if inner mock is not there, create one, `SetupAllProperties` and add the `Setup`. This could lead to pollution of the mock and this vague behavior. What would be drawback of introducing additional info and passing it to `FluentMockVisitor` in order to avoid setup at all if verify is invoked?

stakx:
@lepijohnny - I am in the process of re-architecting the way how such setups are performed internally. Once I'm done with that, `FluentMockVisitor` will likely no longer be needed at all. ;-)

(I should probably mark this and some other issues as "assigned to myself" so others know that a bugfix is already being prepared.)

lepijohnny:
[...]

PR #759, opened 2019-02-26
WIP: Replace `InnerMocks` dictionary with `IReturnValueSetup`

### LINKED ISSUE #703: Setup gets included in `Verify` despite being "unreachable"
### Reproduction code:

```csharp
public interface IX
{
    IX X { get; }
}

var mock = new Mock<IX>();
mock.Setup(m => m.X.X).Verifiable("1");
mock.Setup(m => m.X).Verifiable("2");
_ = mock.Object.X;

mock.Verify();
```

### Actual behavior:

`Verify` throws a `MockException` with message:

```
The following setups on mock 'Mock<IX:00000002>' were not matched:
1: IX m => m.X.X
```

### Expected behavior:

No exception should be thrown, since setup 2 overrides the "root" of setup 1 (`m.X`) and there's no longer any way of getting at 2's `m.X.X` through `mock` (on which verification is triggered).

stakx:
A good portion of the currently documented Moq bugs are related in some way to multi-dot / recursive setup expressions and a mock's internal `InnerMocks` collection used to track its so-called "inner mocks" that get included in verification, thereby turning verification into a "recursive" operation.

Over the past months, I have repeatedly attempted to refactor the Moq codebase such that those are handled in a more logical and robust way, and failed every time because of a fundamental misunderstanding: The inclusion of "inner mocks" in recursive verification should not just happen in the case of multi-dot expressions, but *whenever* Moq can figure out a path from the outer mock to an inner mock. It doesn't matter at all whether an inner mock `y` of an outer mock `x` was set up in a multi-dot setup (`x => x.Y.Z`), or any other way (e.g. by `DefaultValue.Mock`).

This revised understanding of Moq's inner working cannot be implemented 100 % reliably, and may lead to the occasional counterintuitive situation, but in practice it is very handy.

The commits in this PR are mostly a bunch of refactorings to prepare the Moq codebase for further bug fixing with regard to recursive verification (adhering to my revised understanding of its working principle).

The main change is the substitution of `IReturnValueSetup` types for the `Mock.InnerMocks` dictionary. Using setups instead has three main advantages:

1. First, there are no longer both setups and `InnerMocks` that can go out of sync. This makes understanding Moq's mode of operation a lot easier.

2. One example of setups and `InnerMocks` going out of sync is when a setup leading to an inner mock got overridden. However, a corresponding entry in `InnerMocks` might just have gone on staying active, which meant that "unreachable" mocks still got included in verification when they no longer should have.

3. `InnerMocks` is keyed by method exclusively, i.e. argument information (such as matchers) are lost. Using regular setups to track inner mocks means that we'll now be able to properly handle argument matchers in multi-dot setup / verification expressions.

This PR fixes (1) and (2), but only provides the base for (3). We'll (hopefully) fix that in future PRs.

Fixes #703 (and the test posted in https://github.com/moq/moq4/issues/699#issue-362869476).


src/Moq/Mock.cs
```diff
@@ -687,9 +687,10 @@ private static void SetupAllPropertiesPexProtected(Mock mock, DefaultValueProvid
 							initialValue = mock.GetDefaultValue(getter, useAlternateProvider: DefaultValueProvider.Empty);
 						}
 
-						if (initialValue is IMocked mocked)
+						if (Mock.TryGetFromReturnValue(initialValue, out var innerMock))
```
stakx: Up until now, `SetupAllProperties` didn't recursively set up inner mocks yielded by properties of type `[Value]Task<TResult>`. Granted, such properties may be seen as an API design error, but this is still a functional change that should perhaps be documented in a unit test.

src/Moq/Mock.cs
```diff
@@ -403,15 +405,18 @@ internal bool HasAnyInnerMocks()
 
 		internal IEnumerable<Mock> GetInnerMocks()
 		{
-			return this.InnerMocks.Values.Select(inner => inner.Mock);
+			foreach (var returnValue in this.InnerMocks.Values)
+			{
+				Mock.TryGetFromReturnValue(returnValue, out var innerMock);
```
stakx: This is essentially the "unwrap" operation that `MockWithWrappedMockObject` previously prevented by caching the unwrapped mock next to the wrapped return value. I should check how performance is affected by having this caching removed.

src/Moq/Mock.cs
```diff
@@ -395,36 +393,53 @@ internal void AddInnerMock(MethodInfo method, object returnValue)
...
 			Debug.Assert(Mock.TryGetFromReturnValue(returnValue, out _));
 
-			this.InnerMocks.TryAdd(method, returnValue);
+			var arguments = GetArguments(method);
+			var obj = Expression.Parameter(this.MockedType, "mock");
+			var expression = Expression.Lambda(Expression.Call(obj, method, arguments), obj);
+
+			this.Setups.Add(new FixedReturnValueSetup(method, arguments, expression, returnValue));
+
+			Expression[] GetArguments(MethodInfo m)
```
stakx: This is were the difference between setups and `InnerMocks` entries becomes apparent: The former carry argument information, while the latter don't. As we refactor from `InnerMocks` towards setups, this lack of argument information becomes explicit, and we need to fill in the blanks with `It.IsAny<TArg>`. These are temporary placeholders; later we'll be able to build smarter setups.

src/Moq/Mock.cs
```diff
@@ -395,36 +393,53 @@ internal void AddInnerMock(MethodInfo method, object returnValue)
...
+			{
+				var itIsAnyMethod = typeof(It).GetMethod(nameof(It.IsAny), BindingFlags.Public | BindingFlags.Static);
+				var parameterTypes = m.GetParameterTypes();
+				var result = new Expression[parameterTypes.Count];
+				for (int i = 0, n = parameterTypes.Count; i < n; ++i)
+				{
+					result[i] = Expression.Call(itIsAnyMethod.MakeGenericMethod(parameterTypes[i]));
```
stakx: Hmm. I wonder if there are any edge cases &mdash; nullable parameter types, generic type parameter, etc. &mdash; that could make this fail. This might require some further testing.

src/Moq/Mock.cs
```diff
@@ -395,36 +393,53 @@ internal void AddInnerMock(MethodInfo method, object returnValue)
...
+				}
+				return result;
+			}
 		}
 
 		internal bool HasAnyInnerMocks()
 		{
-			return this.InnerMocks.Any();
+			return this.Setups.ToArrayLive(s => s is FixedReturnValueSetup).Any();
```
stakx: Probably just rewrite this as `this.GetInnerMocks().Any()` so we don't DRY.

src/Moq/Mock.cs
```diff
@@ -395,36 +393,53 @@ internal void AddInnerMock(MethodInfo method, object returnValue)
...
 			{
-				Mock.TryGetFromReturnValue(returnValue, out innerMock);
-				return true;
-			}
-			else
-			{
-				innerMock = default;
-				returnValue = default;
-				return false;
+				if (setup.Method == method && Mock.TryGetFromReturnValue(setup.ReturnValue, out innerMock))
```
stakx: The check `setup.Method == method` could be moved into the `ToArrayLive` predicate.

src/Moq/ISetup.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
+using System.Reflection;
+
+namespace Moq
+{
+	internal interface ISetup
```
stakx: I really dislike having this interface. We don't really want two parallel type hierarchies (`Setup` and `ISetup`). It might be better, after all, to just let `AutoImplementedPropertyGetterSetup` sub-class `FixedReturnValueSetup`, which would have to be renamed to e.g. `DeterministicReturnValueSetup`. This would incur a minute additional overhead in `AutoImplementedPropertyGetterSetup`; but we could get rid of these parallel interfaces.

src/Moq/FluentMockVisitor.cs
```diff
@@ -173,11 +178,10 @@ private MethodInfo GetTargetMethod(Type objectType, Type returnType)
 				if (Mock.TryGetFromReturnValue(returnValue, out innerMock))
 				{
 					Mock.SetupAllProperties(innerMock);
-					mock.AddFixedReturnValueSetup(info, returnValue);
+					mock.AddFixedReturnValueSetup(info, arguments, setup, returnValue);
 				}
 			}
 
-			mock.Setup(setup).Returns((TResult)returnValue);
```
stakx: Originally, this setup partially made up for missing argument information in the `InnerMocks` record. It might be an error to move this into a conditional code block when previously, the setup was always recorded. Perhaps we need a unit test to see what difference this could possibly make.

src/Moq/Mock.cs
```diff
@@ -389,6 +389,24 @@ internal static void VerifyNoOtherCalls(Mock mock)
...
+			if (expression.IsProperty())
+			{
+				var property = expression.ToPropertyInfo();
+				Guard.CanRead(property);
+
+				Debug.Assert(method == property.GetGetMethod(true));
+			}
+
+			ThrowIfSetupExpressionInvolvesUnsupportedMember(expression, method);
+			ThrowIfSetupMethodNotVisibleToProxyFactory(method);
```
stakx: Strange that we need validation only in this new overload, but not in the preexisting one...?

src/Moq/Mock.cs
```diff
@@ -407,24 +407,25 @@ internal void AddFixedReturnValueSetup(MethodInfo method, IReadOnlyList<Expressi
 			this.Setups.Add(new FixedReturnValueSetup(method, arguments, expression, returnValue));
 		}
 
-		internal void AddFixedReturnValueSetup(MethodInfo method, object returnValue)
+		internal void AddFixedReturnValueSetup(MethodInfo method, object[] arguments, object returnValue)
```
stakx: (Hmm. Which is better, having a setup where an original `It.IsAny<TArg>` (turned `default(TArg)` in the invocation's `arguments` array) is treated as `default(TArg)`; or having a setup where any original arguments like `1` end up being misrepresented as `It.IsAny<int>`?)

stakx:
Superseded by #760. There's too much happening in this PR for me to be comfortable with it. I'll re-submit the same work as a series of (somewhat) smaller bites.

PR #764, opened 2019-03-02
Make `System.Object` members interceptable by *any* interceptor

stakx:
`System.Object` methods are currently not interceptable if the proxied type is an interface type, and the interceptor used is not an instance of the `Mock` class. These methods should be interceptable by all interceptors, regardless of their type.

As it happens, this requires the introduction of an `IProxy` interface which is to `IInterceptor` what `IMocked` is to `Mock`. It works in exactly the same fashion, too.

This PR might seem a bit pointless, since the only implementation of `IInterceptor` *is* the `Mock` class, so at the surface `IMocked` should suffice.

However, this lays the groundwork for my next PR, which will introduce some new machinery for converting `Action<T>` delegates back to `Expression<Action<T>>` (which is important to fix some long-standing bugs with e.g. `SetupSet`). That PR will require a new interceptor type... so we'll need `IProxy` anyway.

### DIFF
--- a/src/Moq/ProxyFactories/CastleProxyFactory.cs
+++ b/src/Moq/ProxyFactories/CastleProxyFactory.cs
@@ -59,16 +59,21 @@ namespace Moq
 		/// <inheritdoc />
 		public override object CreateProxy(Type mockType, Moq.IInterceptor interceptor, Type[] interfaces, object[] arguments)
 		{
+			// All generated proxies need to implement `IProxy`:
+			var additionalInterfaces = new Type[1 + interfaces.Length];
+			additionalInterfaces[0] = typeof(IProxy);
+			Array.Copy(interfaces, 0, additionalInterfaces, 1, interfaces.Length);
+
 			if (mockType.GetTypeInfo().IsInterface)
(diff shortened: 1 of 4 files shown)


src/Moq/ProxyFactories/CastleProxyFactory.cs
```diff
@@ -59,16 +59,21 @@ public CastleProxyFactory()
 		/// <inheritdoc />
 		public override object CreateProxy(Type mockType, Moq.IInterceptor interceptor, Type[] interfaces, object[] arguments)
 		{
+			// All generated proxies need to implement `IProxy`:
+			var additionalInterfaces = new Type[1 + interfaces.Length];
+			additionalInterfaces[0] = typeof(IProxy);
+			Array.Copy(interfaces, 0, additionalInterfaces, 1, interfaces.Length);
```
stakx: It's a pity that this kind of thing happens twice now; once to add `IMocked<T>` for `Mock`, and once here, to add `IProxy` for `IInterceptor`. It would be nice to eventually optimize this a little.

PR #765, opened 2019-03-03
Remove expression execution from `Setup` and `Verify` & make them recursive

### LINKED ISSUE #699: `Verify` can create setups that causes a subsequent `VerifyAll` to fail

stakx:
When `Setup` and `Verify` are given "recursive" / "fluent" expressions, e.g.

```csharp
mock.Setup(m => m.Foo.Bar.Baz.DoSomething())...;
```

... then up until now, Moq simply compiled that expression & executed it with `DefaultValue.Mock` temporarily enabled to set up the mock object graph leading up to `baz.DoSomething`. As elsewhere, expression compilation and execution has a variety of side effects (e.g. #699), so it would be good to avoid it.

We can actually do something fairly simple: We break apart the given expression into several setup expressions:

* `m => m.Foo`
* `(foo) => (foo).Bar`
* `(bar) => (bar).Baz`
* `(baz) => (baz).DoSomething()`

... then set these sub-expressions recursively: The original `mock` sets up the first expression to yield an inner mock of type `IBar`, who will set up the next expression to return an inner mock of type `IBaz`, and so on.

Recursive verification by `Verify` works similarly, except that it never sets up inner mocks that aren't already there. Instead, it has one additional rule: The absence of an inner mock is treated as a verification error if the expected call count on an inner mock's method is anything other than `Times.Never`.

We can only do this change for `Setup` and `Verify` methods that are already expression-based; i.e. this excludes `SetupSet` and `VerifySet`. We'll deal with those in a later PR.

Fixes #699.

### DIFF
--- a/src/Moq/Mock.cs
+++ b/src/Moq/Mock.cs
@@ -259,42 +259,72 @@ namespace Moq
 			}
 		}
 
-		internal static void VerifyVoid(Mock mock, LambdaExpression expression, Times times, string failMessage)
+		internal static void Verify(Mock mock, LambdaExpression expression, Times times, string failMessage)
 		{
 			Guard.NotNull(times, nameof(times));
 
-			var (obj, method, args) = expression.GetCallInfo(mock);
-			ThrowIfVerifyExpressionInvolvesUnsupportedMember(expression, method);
-
-			var expectation = new InvocationShape(method, args);
-			VerifyCalls(GetTargetMock(obj, mock), expectation, expression, times, failMessage);
-		}
-
-		internal static void VerifyNonVoid(Mock mock, LambdaExpression expression, Times times, string failMessage)
-		{
-			Guard.NotNull(times, nameof(times));
-
-			if (expression.IsProperty())
+			var parts = expression.Split();
+			Mock.VerifyRecursive(mock, expression, parts, times, failMessage, verifyLast: (part, targetMock) =>
 			{
-				VerifyGet(mock, expression, times, failMessage);
-			}
-			else
-			{
-				var (obj, method, args) = expression.GetCallInfo(mock);
-				ThrowIfVerifyExpressionInvolvesUnsupportedMember(expression, method);
-
-				var expectation = new InvocationShape(method, args);
-				VerifyCalls(GetTargetMock(obj, mock), expectation, expression, times, failMessage);
-			}
+				var expectation = new InvocationShape(part.Method, part.Arguments);
+				VerifyCalls(targetMock, expectation, expression, times, failMessage);
+			});
 		}
 
 		internal static void VerifyGet(Mock mock, LambdaExpression expression, Times times, string failMessage)
 		{
-			var method = expression.ToPropertyInfo().GetGetMethod(true);
-			ThrowIfVerifyExpressionInvolvesUnsupportedMember(expression, method);
+			Guard.NotNull(times, nameof(times));
 
-			var expectation = new InvocationShape(method, new IMatcher[0]);
-			VerifyCalls(GetTargetMock(((MemberExpression)expression.Body).Expression, mock), expectation, expression, times, failMessage);
+			if (!expression.IsPropertyIndexer())  // guard because `.ToPropertyInfo()` doesn't (yet) work for indexers
+			{
+				var property = expression.ToPropertyInfo();
+				Guard.CanRead(property);
+			}
+
+			Mock.Verify(mock, expression, times, failMessage);
+		}
+
+		private static void VerifyRecursive(Mock mock, LambdaExpression expression, Stack<LambdaExpressionPart> parts, Times times, string failMessage, Action<LambdaExpressionPart, Mock> verifyLast)
+		{
+			var (expr, method, arguments) = parts.Pop();
+
+			if (mock.IsDelegateMock)
+			{
+				// The expression we have is for a call on the delegate, not our
+				// delegate interface proxy, so we need to map instead to the
+				// method on that interface.
+				_ = ProxyFactory.Instance.GetDelegateProxyInterface(mock.TargetType, out method);
+			}
+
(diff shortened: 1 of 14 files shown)


src/Moq/Mock.cs
```diff
@@ -259,42 +259,77 @@ private MockException TryVerifySetups(Func<Setup, MockException> verifySetup)
...
 
-			var (obj, method, args) = expression.GetCallInfo(mock);
-			ThrowIfVerifyExpressionInvolvesUnsupportedMember(expression, method);
-
-			var expectation = new InvocationShape(method, args);
-			VerifyCalls(GetTargetMock(obj, mock), expectation, expression, times, failMessage);
+			var parts = expression.Split();
+			Mock.VerifyRecursive(mock, expression, parts, times, failMessage, verifyLast: (part, targetMock) =>
+			{
+				var expectation = new InvocationShape(part.Method, part.Arguments);
```
stakx: In the future, we might not need `InvocationShape` anymore. Or rather, we might want to fuse it with `LambdaExpressionPart` into something that represents a single-dot setup expression.

src/Moq/Mock.cs
```diff
@@ -259,42 +259,77 @@ private MockException TryVerifySetups(Func<Setup, MockException> verifySetup)
...
 
-				var expectation = new InvocationShape(method, args);
-				VerifyCalls(GetTargetMock(obj, mock), expectation, expression, times, failMessage);
-			}
+			var parts = expression.Split();
+			Mock.VerifyRecursive(mock, expression, parts, times, failMessage, verifyLast: (part, targetMock) =>
+			{
+				var expectation = new InvocationShape(part.Method, part.Arguments);
+				VerifyCalls(targetMock, expectation, expression, times, failMessage);
+			});
```
stakx: There's some code duplication here. We might just as well delegate to `Mock.Verify`.

src/Moq/Mock.cs
```diff
@@ -259,42 +259,77 @@ private MockException TryVerifySetups(Func<Setup, MockException> verifySetup)
...
+					// some number of invocations were expected on a inner mock that doesn't exist.
+					throw MockException.NoMatchingCalls(
+						failMessage,
+						mock.Setups.ToArrayLive(s => AreSameMethod(s.Expression, expr)),
+						mock.MutableInvocations.ToArray(),
+						expression,
+						times, 0);
+
+					bool AreSameMethod(LambdaExpression l, LambdaExpression r) =>
+						l.Body is MethodCallExpression lc && r.Body is MethodCallExpression rc && lc.Method == rc.Method;
```
stakx: There's some small amount of code duplication here. (The same local function exists in `VerifyCalls`.)

src/Moq/Mock.cs
```diff
@@ -390,78 +390,46 @@ internal static void VerifyNoOtherCalls(Mock mock)
...
-			var targetMock = GetTargetMock(obj, mock);
-			targetMock.Setups.Add(setup);
-
-			return setup;
+			return Mock.SetupRecursive(mock, expression, setupLast: (part, targetMock) =>
+			{
+				var setup = new MethodCall(targetMock, condition, part.Expression, part.Method, part.Arguments);
+				targetMock.Setups.Add(setup);
+				return setup;
+			});
```
stakx: There's some code duplication here, we might just as well delegate to `Setup`, making it clear that the only distinction between `Setup` and `SetupGet` is that the latter performs some additional checks on the setup expression.

src/Moq/Mock.cs
```diff
@@ -598,29 +566,59 @@ private static Expression GetValueExpression(object value, Type type)
...
+		[DebuggerStepThrough]
+		private static TSetup SetupRecursive<TSetup>(Mock mock, LambdaExpression expression, Func<LambdaExpressionPart, Mock, TSetup> setupLast)
 		{
-			if (expression.IsProperty())
+			Debug.Assert(mock != null);
+			Debug.Assert(expression != null);
+			Debug.Assert(setupLast != null);
+
+			var parts = expression.Split();
+			return PexProtector.Invoke(SetupRecursivePexProtected, mock, expression, parts, setupLast);
```
stakx: It's not so clear that we still need `PexProtector` in this place. After all, there's no longer any expression compilation going on, and we're not protecting expression tree operations elsewhere. As it's really difficult to know where to use `PexProtector` without having Pex-based tooling available (i.e. Visual Studio Ultimate with its IntelliTest feature), it may be best to remove `PexProtector` altogether once we've converted `SetupSet` and `VerifySet` to the new recursive / expression-rewriting approach.

src/Moq/Mock.cs
```diff
@@ -598,29 +566,59 @@ private static Expression GetValueExpression(object value, Type type)
...
-				targetMock.Setups.Add(setup);
-				return setup;
+				Mock innerMock;
+				if (!(mock.GetInnerMockSetups().TryFind(method, out var setup) && setup.ReturnsInnerMock(out innerMock)))
+				{
+					var returnValue = mock.GetDefaultValue(method, out innerMock, useAlternateProvider: DefaultValueProvider.Mock);
+					if (innerMock == null)
+					{
+						// TODO:
+						throw new ArgumentException(
```
stakx: Oops! This needs some more attention.

src/Moq/Mock.cs
stakx: We'll soon need to improve `SetupCollection` so that it can identify setups matching a call expression. Best to take care of this duplication when that happens.

PR #766, opened 2019-03-03
Fix `IProxy` interface visibility

stakx:
Otherwise DynamicProxy will not be able to implement it for the same reason it couldn't use `InterfaceProxy` as a proxy base class; so basically , the same explanation from 75d9d8ba applies.

### DIFF
--- a/src/Moq/ProxyFactories/IProxy.cs
+++ b/src/Moq/ProxyFactories/IProxy.cs
@@ -1,15 +1,17 @@
 // Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
-namespace Moq
+using System.ComponentModel;
+
+namespace Moq.Internals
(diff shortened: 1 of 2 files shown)


src/Moq/ProxyFactories/IProxy.cs
```diff
@@ -1,15 +1,17 @@
...
-	internal interface IProxy
+	/// <summary>Do not use. (Moq requires this interface so that <see langword="object"/> methods can be set up on interface mocks.)</summary>
+	// NOTE: This type is actually specific to our DynamicProxy implementation of `ProxyFactory`.
+	// It is directly related to `InterfaceProxy`, see the comment there.
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public interface IProxy
 	{
-		IInterceptor Interceptor { get; }
+		/// <summary/>
+		object Interceptor { get; }
```
stakx: Changing this property's type to `object` is unfortunate, but required unless we want to make `IInterceptor` public, too.

PR #767, opened 2019-03-04
Reconstruct `SetupSet` and `VerifySet` expressions from delegates

stakx:
This converts the current `SetupSet` machinery with a more accurate one based on a new service component called the `ExpressionReconstructor`. It takes a delegate and returns a mostly equivalent expression.

The differences between the new component and the old code are these:

 * The old code executed the setup/verification expression delegate on the actual mock; the new code uses a completely separate and more light-weight recording proxy. (This has become possible thanks to #764.)

* The new code has much better logic for reconstructing argument matchers, no matter where on a fluent / recursive call they happened. This is done mostly through "timestamping" all invocations, combined with a back-tracking algorithm for spreading them across suitable parameters left-to-right.

This PR also does some additional work to ensure custom matchers will work with `SetupSet` and `VerifySet`.


src/Moq/ActionObserver.cs
```diff
@@ -28,70 +32,75 @@ private ActionObserver()
...
+				}
+
+				// Now we've either got no error and a completely reconstructed expression, or
+				// we have an error and a partially reconstructed expression which we can use for
+				// diagnostic purposes:
+				if (error == null)
 				{
-					body = Expression.Call(body, invocation.Method, GetArgumentExpressions(invocation));
+					return (Expression<Action<T>>)Expression.Lambda<Action<T>>(Prettifier.Instance.Visit(body), rootExpression)
+					                                        ;//.PartialMatcherAwareEval();
```
stakx: Delete this.

src/Moq/ActionObserver.cs
```diff
@@ -255,6 +255,7 @@ public void Intercept(Invocation invocation)
 
 		// Post-processes a reconstructed expression to...
 		//  * convert property accessor method calls into property member accesses (`.get_X()` -> `.X`)
+		//  * evaluate captured variables in labmdas to their value
```
stakx: `labmdas` &rarr; `lambdas`

stakx:
Roughly around the time of these commits, I've started noticing that test runs in Visual Studio's _Test Explorer_ no longer work reliably: There's the odd `NullReferenceException` that happens on full test runs, but not when a failing test is debugged. This appears to be caused by tests in `ActionObserverFixture`; whatever tests get executed along with those will fail with it, if executed in a separate batch, they'll succeed.

This needs some further looking into; could actually be a bug, and not just a _Test Explorer_ malfunction.

stakx:
Note to self: Some code paths in the main interception pipeline are still geared to the old "execute and observe" approach that is largely being deprecated here... however `Mock.Of` still relies on it.

It would be good to refactor `Mock.Of` to use the new expression `Split`-ting based, recursive setup machinery prior to this PR getting merged.

Once that's happened, we can remove obsolete code paths from the interception pipeline, plus trim down `AmbientObserver` to a mere `MatcherObserver` (as noone will be registering invocations there anymore).

src/Moq/ActionObserver.cs
```diff
@@ -168,6 +256,18 @@ private Prettifier()
 			{
 			}
 
+			protected override Expression VisitMember(MemberExpression node)
+			{
+				if (node.Member is FieldInfo field && node.Expression is ConstantExpression constant)
+				{
+					return Expression.Constant(field.GetValue(constant.Value));
```
stakx: Rebasing this PR on top of #768 has led to this being duplicated code: The same logic now already exists in `CapturedVariablesEvaluator`. Might want to check whether we can merge the two somehow.

src/Moq/ActionObserver.cs
```diff
@@ -101,8 +101,9 @@ Expression[] GetArgumentExpressions(Invocation invocation, Match[] matches)
 				{
 					expressions[i] = Expression.Constant(invocation.Arguments[i], parameterTypes[i]);
 
-					// Add a `Convert` node (like the C# compiler does) in places where type coercion happens:
-					if (invocation.Arguments[i] != null && invocation.Arguments[i].GetType() != parameterTypes[i])
+					// Add a `Convert` node (like the C# compiler does) in places where nullable type coercion happens:
```
stakx: It might be a good idea to add a unit test fixture that documents the C# compiler's behavior with regard to `Convert` nodes, so that we're really sure what should happen here.

src/Moq/ActionObserver.cs
```diff
@@ -139,7 +132,7 @@ Expression[] GetArgumentExpressions(Invocation invocation, Match[] matches)
 
 							// The remaining matchers can be distributed over the remaining parameters,
 							// so we can use up this matcher:
-							expressions[argumentIndex] = matches[matchIndex].RenderExpression;
+							expressions[argumentIndex] = new MatchExpression(matches[matchIndex]);
```
stakx: FYI that's the whole rationale behind #768.

src/Moq/ActionObserver.cs
```diff
@@ -165,12 +158,23 @@ bool CanDistribute(int msi, int asi)
 					}
 				}
 
-				// Finally box all values that need to be boxed:
+				// Finally, add explicit type casts (aka `Convert` nodes) where necessary:
```
stakx: As above, it might be good to document the exact rules of how the C# compiler handles this as tests.

src/Moq/Language/Flow/WhenPhrase.cs
```diff
@@ -38,13 +38,19 @@ public ISetup<T> Setup(Expression<Action<T>> expression)
 
 		public ISetupSetter<T, TProperty> SetupSet<TProperty>(Action<T> setterExpression)
 		{
-			var setup = Mock.SetupSet(mock, setterExpression, this.condition);
+			Guard.NotNull(setterExpression, nameof(setterExpression));
+			var expression = ExpressionReconstructor.Instance.ReconstructExpression(setterExpression);
+
+			var setup = Mock.SetupSet(mock, expression, this.condition);
```
stakx: That kind of code is duplicated a lot in other overloads. If `ReconstructExpression` were not generic, this code pattern could be pushed down into the `Mock` class, and this fluent API method (as well as others) would turn into one-liners.

src/Moq/Obsolete/Mock.Legacy.cs
```diff
@@ -1,24 +0,0 @@
...
-
-using Moq.Matchers;
-
-namespace Moq
-{
-	// Keeps legacy implementations.
-	public partial class Mock
-	{
-		[Obsolete]
-		internal static void VerifySet(Mock mock, LambdaExpression expression, Times times, string failMessage)
```
stakx: Please check once more whether it is really safe to delete this. Which other methods used this one? What are they calling now instead?

src/Moq/Protected/ProtectedMock.cs
```diff
@@ -219,9 +221,10 @@ public void VerifySet<TProperty>(string propertyName, Times times, object value)
 			ThrowIfPublicSetter(property, typeof(T).Name);
 			Guard.CanWrite(property);
 
+			var expression = GetSetterExpression(property, ItExpr.IsAny<TProperty>());
 			// TODO should consider property indexers
 			// TODO should receive the parameter here
```
stakx: Check whether these TODOs. I am suspecting that indexers work now (let's add tests to find out for sure!), and I cannot make sense of the second TODO.

tests/Moq.Tests/ActionObserverFixture.cs
```diff
@@ -44,8 +44,8 @@ public void Void_method_call_with_coerced_arg()
 			public void Void_method_call_with_coerced_nullable_arg()
 			{
 				AssertReconstructable(
-					x => x.VoidWithNullableInt(42),
-					x => x.VoidWithNullableInt(42));
+					"x => x.VoidWithNullableInt(42)",
+					 x => x.VoidWithNullableInt(42));
```
stakx: Need to find out why this change was necessary.

stakx:
> Roughly around the time of these commits, I've started noticing that test runs in Visual Studio's Test Explorer no longer work reliably: There's the odd `NullReferenceException` that happens on full test runs, but not when a failing test is debugged. This appears to be caused by tests in `ActionObserverFixture`; whatever tests get executed along with those will fail with it, if executed in a separate batch, they'll succeed.

This turned out to be an actual code issue caused by static member initialization order between base and derived class; I juggled around the `instance` and `Instance` bits of `ActionObserver` and `ExpressionReconstructor`, should be fine now. The same change might have to be done for `DelegateCompiler` et. al.

src/Moq/Protected/ProtectedMock.cs
stakx: The TODOs are still valid:

* The first one is relevant because validation before the `GetSetterExpression` call might fail for indexers.

* The second one could mean that it should be possible for `VerifySet` to receive the right-hand side of the assignment invocation to be verified, instead of just assuming `It.IsAny<T>()`.

tests/Moq.Tests/ActionObserverFixture.cs
stakx: Non-null literal for nullable parameters causes a `Convert`. We don't care about that, so a string comparison is fine.

src/Moq/ActionObserver.cs
stakx: Done. Two expression tree rewriters have been moved into a dedicated namespace, I am anticipating that there will eventually be a few more of those.

src/Moq/ActionObserver.cs
stakx: Done, but I'm still not feeling super-confident about this. The rules for when `Convert` nodes are required are somewhat arcane.

stakx:
To be perfectly honest, I do expect this to cause a few problems in the beginning (when people start updating from <4.11.0 to >=4.11.0), and thus some patching will likely be required. However, for now I think I've done my best.

PR #768, opened 2019-03-04
Make handling of `MatchExpression` more robust

stakx:
This is extracted from, and required by, #767. Moq should be able to process expressions containing a `MatchExpression` more or less everywhere except during compilation (because it doesn't carry enough information to properly allow that scenario).

### DIFF
--- a/src/Moq/Evaluator.cs
+++ b/src/Moq/Evaluator.cs
@@ -33,7 +33,7 @@ namespace Moq
 		/// <returns>A new tree with sub-trees evaluated and replaced.</returns>
 		public static Expression PartialEval(Expression expression)
 		{
-			return PartialEval(expression, e => e.NodeType != ExpressionType.Parameter);
+			return PartialEval(expression, e => e.NodeType != ExpressionType.Parameter && !(e is MatchExpression));
 		}
 
 		/// <summary>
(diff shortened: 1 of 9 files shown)


src/Moq/Evaluator.cs
```diff
@@ -33,7 +33,7 @@ public static Expression PartialEval(Expression expression, Func<Expression, boo
 		/// <returns>A new tree with sub-trees evaluated and replaced.</returns>
 		public static Expression PartialEval(Expression expression)
 		{
-			return PartialEval(expression, e => e.NodeType != ExpressionType.Parameter);
+			return PartialEval(expression, e => e.NodeType != ExpressionType.Parameter && !(e is MatchExpression));
```
stakx: Instead of this change, perhaps it would be better to add a check right inside `Evaluator` that recognizes irreducible `Extension` nodes.

src/Moq/ExpressionExtensions.cs
```diff
@@ -334,6 +334,9 @@ private static bool PartialMatcherAwareEval_ShouldEvaluate(Expression expression
 				case ExpressionType.Parameter:
 					return false;
 
+				case ExpressionType.Extension:
+					return !(expression is MatchExpression);
+
```
stakx: Instead of this change, perhaps it would be better to add a check right inside `Evaluator` that recognizes irreducible `Extension` nodes.

PR #774, opened 2019-03-07
Allow raising events on inner mocks

### LINKED ISSUE #166: Mock<T>.Raise only raises events on root object

stakx:
Fixes #166, i.e. when `mock.Raise` is given an event `+= null` expression for a sub-mock's event, then Moq will now raise the event on that sub-mock instead of on the root mock. This is achieved by the new machinery we recently added in #767 (improved expression reconstruction from delegates) and #765 (expression splitting & recursive processing).


src/Moq/ExpressionExtensions.cs
```diff
@@ -153,9 +157,25 @@ void Split(Expression e, out Expression r /* remainder */, out LambdaExpressionP
 							arguments[ai] = lhs.Arguments[ai];
 						}
 						arguments[arguments.Length - 1] = assignmentExpression.Right;
+
+						Func<Expression, Expression, BinaryExpression> assign;
+						switch (e.NodeType)
```
stakx: This isn't particularly nice. Instead of `Expression.Assign`, use `Expression.MakeBinary` and specify `e.NodeType`.

src/Moq/MethodCall.cs
```diff
@@ -191,29 +191,25 @@ public void SetCallbackResponse(Delegate callback)
...
-			}
-			else
-			{
-				this.raiseEventResponse = null;
-			}
+			Guard.NotNull(eventExpression, nameof(eventExpression));
+
+			var expression = ExpressionReconstructor.Instance.ReconstructExpression(eventExpression);
+
+			// TODO: validate that expression is for event subscription or unsubscription
```
stakx: An obvious TODO left to do.

src/Moq/MethodCall.cs
```diff
@@ -191,29 +191,25 @@ public void SetCallbackResponse(Delegate callback)
...
```
stakx: As above, to this TODO.

src/Moq/MethodCall.cs
```diff
@@ -469,40 +465,54 @@ public override void RespondTo(Invocation invocation)
...
+						this.expression,
+						this.expression.Split(),
+						args);
 				}
+
+				Mock.RaiseEvent(
+					this.mock,
+					this.expression,
+					this.expression.Split(),
+					args);
```
stakx: Oops. Some obvious repetition here. Remove the inner call, and put the second one on a single line for brevity.

PR #777, opened 2019-03-08
`InvocationShape` should be equality-comparable

stakx:
Another refactoring that makes `InvocationShape` that will be needed shortly. It also permits a clearer phrasing in `SetupCollection` when a setup overrides another.

### DIFF
--- a/src/Moq/InvocationShape.cs
+++ b/src/Moq/InvocationShape.cs
@@ -18,9 +18,9 @@ namespace Moq
 	///     for faster access.
 	///   </para>
 	/// </summary>
-	internal sealed class InvocationShape
+	internal sealed class InvocationShape : IEquatable<InvocationShape>
 	{
-		private static readonly IReadOnlyList<Expression> noArguments = new Expression[0];
+		private static readonly Expression[] noArguments = new Expression[0];
 		private static readonly IMatcher[] noArgumentMatchers = new IMatcher[0];
 
 		public readonly LambdaExpression Expression;
@@ -28,6 +28,7 @@ namespace Moq
 		public readonly IReadOnlyList<Expression> Arguments;
 
 		private readonly IMatcher[] argumentMatchers;
+		private Expression[] partiallyEvaluatedArguments;
 
 		public InvocationShape(LambdaExpression expression, MethodInfo method, IReadOnlyList<Expression> arguments = null)
 		{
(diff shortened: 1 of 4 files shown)


src/Moq/InvocationShape.cs
```diff
@@ -109,5 +110,71 @@ private bool IsOverride(MethodInfo invocationMethod)
...
 		}
+
+		public bool Equals(InvocationShape other)
+		{
+			if (this.Method != other.Method)
+			{
+				return false;
+			}
+
+			if (this.Arguments.Count != other.Arguments.Count)
```
stakx: This could probably be just an assertion. The argument count should always be equal if the methods are already the same.

src/Moq/InvocationShape.cs
```diff
@@ -109,5 +110,71 @@ private bool IsOverride(MethodInfo invocationMethod)
...
+			{
+				return false;
+			}
+
+			if (this.partiallyEvaluatedArguments == null)
+			{
+				this.partiallyEvaluatedArguments = PartiallyEvaluateArguments(this.Arguments);
```
stakx: It might be good to someday think about whether this needs additional precautions for thread-safety. For now I'll pretend to be naïve and think that this operation would yield the same result even when called several times by several threads, so it's fine to just let the last evaluation win.

src/Moq/InvocationShape.cs
```diff
@@ -109,5 +110,71 @@ private bool IsOverride(MethodInfo invocationMethod)
...
+			}
+
+			if (other.partiallyEvaluatedArguments == null)
+			{
+				other.partiallyEvaluatedArguments = PartiallyEvaluateArguments(other.Arguments);
```
stakx: As above.

src/Moq/Setup.cs
```diff
@@ -21,6 +21,8 @@ protected Setup(InvocationShape expectation)
 
 		public virtual Condition Condition => null;
 
+		public InvocationShape Expectation => this.expectation;
+
 		public LambdaExpression Expression => this.expectation.Expression;
```
stakx: Should check whether `Expression`, `Method` still need to be exposed at all now that they'd be accessible through `Expectation`. It might be convenient to keep them around.

src/Moq/SetupCollection.cs
```diff
@@ -86,7 +86,7 @@ public Setup[] ToArrayLive(Func<Setup, bool> predicate)
 			// The following verification logic will remember each processed setup so that duplicate setups
 			// (that is, setups overridden by later setups with an equivalent expression) can be detected.
 			// To speed up duplicate detection, they are partitioned according to the method they target.
```
stakx: That partitioning might no longer be as crucial as before, since the first equality check `Invocation.Equals` performs is for the `MethodInfo`s to be equal.

Issue #428, opened 2017-07-18
Issue with Moq.pdb in version 4.7.99: "Symbol indexes could not be retrieved"


epot:
I am getting an error when trying to publish the symbols:
    
    Symbol indexes could not be retrieved.

stakx:
Can you please describe in more detail what you're attempting to do? (For example, as I understand it, you should consume the symbols, not publish them. What location are you publishing to?)

epot:
Thanks for the quick answer! To be honest, I am not sure what is going on under the hood. I am using a TFS build that has a step "Index sources & publish symbols". That step was working with my previous moq dependency (4.7.1) and not anymore. I will try to find out more about what it is trying to do.

stakx:
I'm wondering why TFS would even attempt to source-index a third-party dependency's PDBs.

The change here (as mentioned in the [changelog](https://github.com/moq/moq4/blob/master/CHANGELOG.md#4799-2017-07-17)) is that Moq 4.7.99 now includes PDBs again (for the first time since version 4.5.7), and they are **already** source-indexed.

I suppose you might want to simply exclude that PDB from the build step you mention, if that is an option?

epot:
I will check that, I wonder indeed why it does not only work on my component pdb.

stakx:
@epot: Possibly because your component uses the classic PDB format (`full`, `pdb-only`)? Moq uses the relatively new `portable` PDB format. The people at Microsoft clearly say that there are still many tools that don't understand this new format, but it's the future and that's where they invest their efforts.

Magrangs:
I am also getting the same issue with a hosted buid in team services online

stakx:
As suggested in [this comment over at sendgrid/sendgrid-csharp#494](https://github.com/sendgrid/sendgrid-csharp/issues/494#issuecomment-317065010) I might also change the .NET 4.5 target's PDB back to `<DebugType>full</DebugType>` so you can upgrade again (if you want :-). I've subscribed to that other issue and will wait to see how it resolves.

gregveres:
+1 I hit this tonight. I backed off to the previous version and the build is working just fine for now. I will keep an eye out for a new version that works with VSTS and .Net 4.5.

Thanks

stakx:
[...]

PR #789, opened 2019-03-10
Use portable PDB in separate NuGet symbol package

### LINKED ISSUE #447: Consider switching to portable PDB symbol files
This is a reminder (not likely to be relevant before mid-2018) to check whether we can make the switch back to portable PDBs. We reverted (in #443) to classic Windows PDBs because of #428.

As mentioned in that issue, the main showstopper appears to be TFS. Microsoft will supposedly have fixed the issue in the 2018 version of TFS (as mentioned [here](https://developercommunity.visualstudio.com/content/problem/33776/indexed-source-information-could-not-be-retrieved-1.html)).

stakx:
We want to continue publishing SourceLink-ed debug symbols, but using classic Windows PDBs generates a huge package. Using the much smaller portable PDB format would be much better, but some people may still experience problems with those. (See e.g. GitHub issue #428.)

Let's follow the official guidance (see links below) and publish debug symbols in a separate `.snupkg` package to NuGet's symbol server. This means that we can (must, in fact) use portable PDBs as it will be the consumers' choice whether to use the separate symbols or not.

Closes #447.

References:
 * https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg
 * https://github.com/dotnet/SourceLink#alternative-pdb-distribution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 ## Unreleased
 
+#### Changed
+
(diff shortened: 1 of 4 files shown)


stakx:
Changed this PR's status to "work in progress" because AppVeyor apparently does not yet support pushing `.snupkg` to NuGet, see https://github.com/appveyor/ci/issues/2753.

stakx:
AppVeyor still doesn't support `.snupkg` packages, I am going to close this PR until that changes.

stakx:
This looks good as far as I can see, but I'll have to run a final test and actually publish to NuGet (as another pre-release).

PR #790, opened 2019-03-29
add It.Ignore for AnyMatcher

helloserve:
While experimenting with .NET Core 3.0 preview I ran into a problem where they were no longer boxing an internal class to `object` to the `ILogger.Log<TState>` abstraction. This means that I can no longer `Setup` or `Verify` against that abstraction, since `It.IsAny<>()` has to exactly match the invocation types. See https://github.com/aspnet/Extensions/issues/1319 for the details on this. And really, as we discover more changes in 3.0 Preview I expect to see more of these decisions impacting us.

The only solution I can think of is to have a way to completely short-circuit on specific parameters (who's types you cannot see) by explicitly choosing to do so via a new `It.Ignore<>()` method.

I don't expect you to merge this in to 4.0 at all, but rather to see it as a feature request for 5.0 via implementation. However if 3.0 is finalized and lands before Moq 5.0 ships I expect a few projects will be broken when they upgrade. At least then you'll have this to decide to include or not.

### DIFF
--- a/src/Moq/It.cs
+++ b/src/Moq/It.cs
@@ -51,6 +51,12 @@ namespace Moq
 			return Expression.Call(It.isAnyMethod.MakeGenericMethod(genericArgument));
 		}
 
+		/// <include file='It.xdoc' path='docs/doc[@for="It.Ignore"]/*'/>
+		public static TValue Ignore<TValue>()
+		{
+			return Match<TValue>.Create<TValue>(null, () => It.Ignore<TValue>());
+		}
+
 		/// <include file='It.xdoc' path='docs/doc[@for="It.IsNotNull"]/*'/>
 		public static TValue IsNotNull<TValue>()
 		{
--- a/src/Moq/It.xdoc
+++ b/src/Moq/It.xdoc
@@ -28,6 +28,22 @@
 		</example>
 		<typeparam name="TValue">Type of the value.</typeparam>
 	</doc>
+	<doc for="It.Ignore">
+		<summary>
+			Ignores any type or value.
+		</summary>
+		<remarks>
+			Typically used when the actual argument type for the method is 
+			internal or private.
+		</remarks>
+		<example>
+			<code>
+				// Verifies that a call to the logger was made for severity Warning, regardless of the actual log formatter.
+				mock.Verify(x => x.Log(LogLevel.Warning, It.IsAny&lt;EventId&gt;(), It.IsAny&lt;object&gt;(), It.IsAny&lt;Exception&gt;(), It.Ignore&lt;Func&lt;object, Exception, string&gt;&gt;()));
+			</code>
+		</example>
+		<typeparam name="TValue">Type of the value.</typeparam>
+	</doc>
 	<doc for="It.IsNotNull">
 		<summary>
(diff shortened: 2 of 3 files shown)


stakx:
Hi @helloserve, thanks for raising this issue (well, PR actually). I'll review it soon and look at .NET Core 3 a little more closely.

In the meantime, remember that you don't need to wait for Moq to have a suitable matcher, you can simply create a custom matcher yourself, using `Match.Create`:

```csharp
static T Ignore<T>()
{
    return Match.Create(x => true, () => Ignore<T>());
}
```

helloserve:
Ah right. Now I feel kinda dumb. I did not know about Custom matchers. Thanks!

stakx:
@helloserve, sorry for the long silence, I've actually been looking into this several weeks ago and seen that custom matchers don't help in your case, after all. I can see how this PR would help solve some situations that Moq currently doesn't have any other solution for.

On the other hand, I'd really prefer that we *didn't* have to add this `It.Ignore` matcher, mainly for the reason that it might be rather confusing to have it alongside `It.IsAny`, as the difference between these two might be too subtle for most users to grasp easily.

I've been thinking that we might add it but keep it from showing up in IntelliSense using `[EditorBrowsable(EditorBrowsableState.Advanced)]` trickery, but that feels like adding something just because we haven't solved another, larger problem in Moq in a clean and thorough manner. 

Do you see any other solutions for what you want to achieve?

helloserve:
I completely feel the same way about it. Having to ignore a parameter seems counter intuitive to setting up a mock, and matching invocations specifically is crucial to prevent false-positive results. Or at least, I can totally see how false positives can occur when using `It.Ignore()` wantonly.  This is why the decision to internalize types used on the public side of (the ILogger) abstractions is pretty disappointing to me.

I think the only really clean solution here is if C# can introduce the `bottom` keyword - see [here](https://github.com/dotnet/csharplang/issues/538). Using it as `It.IsAny<IReadOnlyList<bottom>>()` for example would match the type in this particular instance for the `ILogger` invocation. I do however find that this is a pretty advanced concept. I had to reread that proposal discussion probably 5 times...

So while we wait for that, `It.Ignore()` is really just a temporary workaround tbh. I made the PR in a moment of frustration and used the word _feature_, but really it's all it is. I'm not sure how to decorate that effectively, or how long we'll wait. Having it as an undocumented feature doesn't feel right though. Perhaps just mark it as deprecated straight out of the gate? Or perhaps we can change the name to be `It.IgnoreType()` to be slightly more specific?

Personally I don't believe that Moq itself has a larger problem to solve here. The implementation policies of the Core framework is what is changing and causing this, after all.

stakx:
>> It looks like Moq should support bottom-nothing-never like type in its generic type parameter DSL dotnet/csharplang#538 to allow match any generic (and they can do that with no language level support)
>
> How to do that without having bottom, or know all the (internal or not) types is beyond me thus far. I can help you investigate that if you want?

I believe that #343 is about just that.

> `It.Ignore()` is really just a temporary workaround tbh.

Agreed. And as such, we shouldn't add it to Moq because once added, it will no longer be able to stay "temporary"&mdash;removing it again will be a breaking change.

I suggest that we don't merge this PR for the above reasons and instead keep at #343.

Issue #791, opened 2019-04-01
Stackoverflow when setting up and calling object.Equals(object) on T : IEquatable<T>


BrunoJuchli:
### Version
Moq 4.10.1
.Net Framework 4.7.2

### Description
performing `mocked.Equals(new object()` throws a `StackOverflowException`. Conditions:

- mocked imlements `IEquatable<T>` interface
- setup for `object.Equals(object)` done which references the mock

### Repro

Given the following class `Foo` which implements the `IEquatable<Foo>` interface:

    public class Foo : IEquatable<Foo>
    {
        public virtual bool Equals(Foo other)
        {
            throw new NotImplementedException();
        }

        public override bool Equals(object obj)
        {
            throw new NotImplementedException();
        }

        public override int GetHashCode()
        {
            throw new NotImplementedException();
        }
    }

performing the following throws a `StackOverflowException`:

    var foo = new Mock<Foo>();

    foo.Setup(x => x.Equals((object)foo.Object)).Returns(true);

    foo.Object.Equals(new object());

### Workaround

setup `object.Equals(object)` like this instead:

    var foo = new Mock<Foo>();
    foo.Setup(x => x.Equals(It.Is<object>(o => ReferenceEquals(o, foo.Object)))).Returns(true);

stakx:
Hi @BrunoJuchli, thanks for reporting. I'll take a closer look at this shortly to find out where Moq starts running in circles.

In theory, though, you shouldn't need such a setup at all:

> ```csharp
> foo.Setup(x => x.Equals((object)foo.Object)).Returns(true);
> ```

Because Moq should provide exactly that behavior out of the box:

https://github.com/moq/moq4/blob/efaeafd3860880c17b3ca16a8f5f5e2a882c84e6/src/Moq/Interception/InterceptionAspects.cs#L37-L43

BrunoJuchli:
@stakx 
Thanks. Note: using [windbg you can catch `StackoverflowException`](https://stackoverflow.com/questions/4733878/how-to-debug-a-stackoverflowexception-in-net/49882734#49882734).

I'll look into the default behavior. I was under the impression it wasn't actually working like that, but I' was probably mistaken - I'll check tomorrow! :-)

stakx:
@BrunoJuchli - you can reduce the repro code to the following three lines:

```csharp
var mock = new Mock<object>();
mock.Setup(m => m.Equals(mock.Object)).Returns(true);
mock.Object.Equals(new object());
```

[...]

PR #792, opened 2019-04-01
Make `ConstantMatcher` more (but not fully) `StackOverflowException`-proof

stakx:
This is in response to #791.


stakx:
@BrunoJuchli, are you OK with leaving it at this, or would you like to go further?

PR #793, opened 2019-04-01
Keep `InvocationShape` argument matchers & expressions in sync

### LINKED ISSUE #711: `VerifyAll` throws System.InvalidOperationException: 'Nullable object must have a value.'
I'm running a unit test that passed using Moq 4.2.1409.1722 but now fails using 4.10.0 with the exception above. I loaded the symbols for Moq to see if I could figure out what was failing. This is what I saw:

image

Not sure if related, but note the error when trying to view the `expression` parameter: "Cannot obtain value of the local variable or argument because it is not available at this instruction pointer, possibly because it has been optimized away."

I went up the call stack to see if `expression` was being passed in. I can see that expression had a value:

image

I might be going down the wrong path, but to me it appears like something is causing that expression to disappear inside the call to `this.fnCanBeEvaluated(expression)`  which I am unable to see inside of.

So it might be a bug in Moq or, if it's a problem with my code, then Moq should probably check for the condition that causes the exception and throw an outer exception with a better description.

stakx:
Fixes #711. See explanation in https://github.com/moq/moq4/issues/711#issuecomment-433386423 as well as in commit messages for details.

### DIFF
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -17,6 +17,38 @@ namespace Moq
 {
 	internal static partial class ExpressionExtensions
 	{
+		/// <summary>
+		///   Wraps this <paramref name="expression"/> in a <see cref="ExpressionType.Convert"/> node if needed.
+		/// </summary>
+		/// <param name="expression">The <see cref="Expression"/> which should be wrapped.</param>
+		/// <param name="type">The <see cref="Type"/> with which to make the <paramref name="expression"/> compatible.</param>
+		/// <remarks>
+		///   LINQ expression trees generally enforce type compatibility rules that are stricter than
+		///   the assignment-compatibility used by e.g. <see cref="Type.IsAssignableFrom(Type)"/>. For
+		///   example, while <see langword="int"/> is assignable-to <see langword="object"/>, you
+		///   will need a conversion in a LINQ expression tree to model the value-type boxing operation.
+		/// </remarks>
+		internal static Expression ConvertIfNeeded(this Expression expression, Type type)
+		{
+			if (expression.Type == type)
+			{
+				// if source and target types match exactly,
+				// no conversion is necessary:
+				return expression;
+			}
+			if (!expression.Type.IsValueType && !type.IsValueType && type.IsAssignableFrom(expression.Type))
+			{
+				// if source and target types are reference types and assignment-compatible,
+				// no conversion is necessary:
+				return expression;
+			}
+			else
+			{
+				// everything else requires a conversion:
+				return Expression.Convert(expression, type);
+			}
+		}
+
 		internal static Delegate CompileUsingExpressionCompiler(this LambdaExpression expression)
 		{
 			// Expression trees are not compiled directly.
--- a/src/Moq/MatcherFactory.cs
+++ b/src/Moq/MatcherFactory.cs
@@ -15,29 +15,30 @@ namespace Moq
 {
 	internal static class MatcherFactory
 	{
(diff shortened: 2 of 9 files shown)


src/Moq/MatcherFactory.cs
```diff
@@ -67,20 +68,20 @@ public static IMatcher CreateMatcher(Expression argument, ParameterInfo paramete
...
 					}
 
-					return new RefMatcher(constant.Value);
+					return new Pair<IMatcher, Expression>(new RefMatcher(constant.Value), constant);
 				}
 			}
 			else if (parameter.IsDefined(typeof(ParamArrayAttribute), true) && (argument.NodeType == ExpressionType.NewArrayInit || !argument.Type.IsArray))
 			{
-				return new ParamArrayMatcher((NewArrayExpression)argument);
+				return new Pair<IMatcher, Expression>(new ParamArrayMatcher((NewArrayExpression)argument), argument);
```
stakx: Using `argument` as the associated expression here isn't actually good enough, since `ParamArrayMatcher` will re-invoke `MatcherFactory` for each of its sub-parameter expressions. A new `NewArrayExpression` should be constructed from that.

src/Moq/Matchers/ParamArrayMatcher.cs
```diff
@@ -16,7 +16,7 @@ public ParamArrayMatcher(NewArrayExpression expression)
 			if (expression != null)
 			{
 				this.matchers = expression.Expressions
-					.Select(e => MatcherFactory.CreateMatcher(e)).ToArray();
+					.Select(e => MatcherFactory.CreateMatcher(e).Item1).ToArray();
```
stakx: As above. We're losing `.Item2` here (i.e. the expression from which the matcher instance is derived).

src/Moq/MatcherFactory.cs
```diff
@@ -71,9 +71,23 @@ internal static class MatcherFactory
 					return new Pair<IMatcher, Expression>(new RefMatcher(constant.Value), constant);
 				}
 			}
-			else if (parameter.IsDefined(typeof(ParamArrayAttribute), true) && (argument.NodeType == ExpressionType.NewArrayInit || !argument.Type.IsArray))
```
stakx: This last condition just plain didn't make any sense. Why would we want to enter into the `ParamArrayMatcher` clause for `params` parameters whose argument is *not* an array?! This may have been an attempt to allow `null` values in place of an empty array, but at least the C# compiler will not produce such a value.

src/Moq/ExpressionExtensions.cs
```diff
@@ -17,6 +17,38 @@ namespace Moq
...
+		/// </summary>
+		/// <param name="expression">The <see cref="Expression"/> which should be wrapped.</param>
+		/// <param name="type">The <see cref="Type"/> with which to make the <paramref name="expression"/> compatible.</param>
+		/// <remarks>
+		///   LINQ expression trees generally enforce type compatibility rules that are stricter than
+		///   the assignment-compatibility used by e.g. <see cref="Type.IsAssignableFrom(Type)"/>. For
+		///   example, while <see langword="int"/> is assignable-to <see langword="object"/>, you
+		///   will need a conversion in a LINQ expression tree to model the value-type boxing operation.
+		/// </remarks>
+		internal static Expression ConvertIfNeeded(this Expression expression, Type type)
```
stakx: * Check whether this method can be reused in `ActionObserver`, which currently has [its own (but slightly different) logic](https://github.com/moq/moq4/blob/0aa95749a97a9504e0e3adc12b398c02301b0ec5/src/Moq/ActionObserver.cs#L186-L209) for doing the same thing.

 * Also, add a set of tests that test our assumptions (stated here informally inside `<remarks>`) about when `Convert` nodes are necessary.

PR #796, opened 2019-04-03
Match indexers used as arguments in setup expression eagerly

### LINKED ISSUE #714: ArgumentOutOfRangeException when setup expression contains indexer access
I'm getting an ArgumentOutOfRangeException when my Setup expression contains accessing an object in an array at an index.

```
_Helper.Setup(s => s.Parse(containers[i]));
```

This can be remedied by first accessing the object at the index.

```
var currentContainer = containers[i];
_Helper.Setup(s => s.Parse(currentContainer));
```

If this is intended behavior, could Moq be enhanced to detect this type of expression and throw an exception with a helpful message? This would add value in troubleshooting these expressions.

Below is some code to reproduce the exception and the result stack trace.

```
public class Container
{
    public int MyProperty { get; set; }
}
public interface IHelper
{
    void Parse(Container container);
}
public class Parser
{
    IHelper _Helper;
    public Parser(IHelper helper)
    {
        _Helper = helper;
    }
    public void Parse(List<Container> containers)
    {
        foreach (var container in containers)
            _Helper.Parse(container);
    }
}
[TestFixture]
public class HelperTests
{
    Mock<IHelper> _Helper;
    Parser _Parser;
    [SetUp]
    public void SetUp()
    {
        _Helper = new Mock<IHelper>();
        _Parser = new Parser(_Helper.Object);
    }
    [Test]
    public void Parse()
    {
        var containers = new List<Container> { new Container() };
        for (var i = 0; i < containers.Count; i++)
            _Helper.Setup(s => s.Parse(containers[i]));
        _Parser.Parse(containers);
    }
}
```

```
Result StackTrace:	
at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at Moq.Evaluator.SubtreeEvaluator.Evaluate(Expression e) in C:\projects\moq4\src\Moq\Evaluator.cs:line 73
   at Moq.Matchers.LazyEvalMatcher.Matches(Object value) in C:\projects\moq4\src\Moq\Matchers\LazyEvalMatcher.cs:line 20
   at Moq.InvocationShape.IsMatch(Invocation invocation) in C:\projects\moq4\src\Moq\InvocationShape.cs:line 52
   at Moq.SetupCollection.FindMatchFor(Invocation invocation) in C:\projects\moq4\src\Moq\SetupCollection.cs:line 63
   at Moq.FindAndExecuteMatchingSetup.Handle(Invocation invocation, Mock mock) in C:\projects\moq4\src\Moq\Interception\InterceptionAspects.cs:line 117
   at Moq.Mock.Moq.IInterceptor.Intercept(Invocation invocation) in C:\projects\moq4\src\Moq\Interception\Mock.cs:line 21
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at UnitTests.Parser.Parse(List`1 containers) in REDACTED
   at UnitTests.HelperTests.Parse() in REDACTED
--ArgumentOutOfRangeException
   at lambda_method(Closure )
Result Message:	
System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
  ----> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
```

stakx:
This resolves #714.

It doesn't make sense for indexers to be matched lazily when other properties are being matched eagerly, so this will make Moq a little more consistent.

(While technically a small breaking change, making it anyway seems permissible because Moq has neglected indexers far too long and essentially left its behavior regarding them semi-unspecified.)

### DIFF
--- a/src/Moq/MatcherFactory.cs
+++ b/src/Moq/MatcherFactory.cs
@@ -132,9 +132,13 @@ namespace Moq
 				}
 #pragma warning restore 618
 
-				return new Pair<IMatcher, Expression>(new LazyEvalMatcher(originalExpression), originalExpression);
+				var method = call.Method;
+				if (!method.IsPropertyGetter() && !method.IsPropertyIndexerGetter())
+				{
+					return new Pair<IMatcher, Expression>(new LazyEvalMatcher(originalExpression), originalExpression);
+				}
 			}
-			else if (expression is MemberExpression memberAccess)
+			else if (expression is MemberExpression || expression is IndexExpression)
 			{
 				if (expression.IsMatch(out var match))
 				{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This release contains several **minor breaking changes**. Please review your cod
 * `mock.Verify` no longer creates setups, nor will it override existing setups, as a side-effect of using a recursive expression. (@stakx, #765)
 * More accurate detection of argument matchers with `SetupSet` and `VerifySet`, especially when used in fluent setup expressions or with indexers (@stakx, #767)
 * `mock.Verify(expression)` error messages now contain a full listing of all invocations that occurred across all involved mocks. Setups are no longer listed, since they are completely irrelevant in the context of call verification. (@stakx, #779, #780)
+* Indexers used as arguments in setup expressions are now eagerly evaluated, like all other properties already are (except when they refer to matchers) (@stakx, #794)
 
 #### Added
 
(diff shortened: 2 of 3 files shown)


src/Moq/MatcherFactory.cs
```diff
@@ -132,9 +132,13 @@ internal static class MatcherFactory
 				}
 #pragma warning restore 618
 
-				return new Pair<IMatcher, Expression>(new LazyEvalMatcher(originalExpression), originalExpression);
+				var method = call.Method;
+				if (!method.IsPropertyGetter() && !method.IsPropertyIndexerGetter())
```
stakx: This actually does a little more than target indexers. It also makes sure that if somehow a property or index expression has made it into the setup expression in the form of a `get_*` method call, we recognize them anyway.

PR #803, opened 2019-04-09
Add ReturnsAsync method for sequence setups

### LINKED ISSUE #795: Setup sequence doesn't support ReturnsAsync with Func<>
In SetupSequence there is an overload for Returns can that take Func<> as an argument, but there isn't such overload for ReturnsAsync. 

I think that adding this overload to ReturnsAsync would be a nice enhancement.

VladFasie:
Closes #795.

### DIFF
--- a/src/Moq/Language/Flow/SetupSequencePhrase.cs
+++ b/src/Moq/Language/Flow/SetupSequencePhrase.cs
@@ -4,7 +4,7 @@
 using System;
 using System.ComponentModel;
 using System.Reflection;
-
+using System.Threading.Tasks;
 using Moq.Properties;
 
 namespace Moq.Language.Flow
@@ -71,9 +71,31 @@ namespace Moq.Language.Flow
 			if (valueFunction is TResult)
 			{
 				return this.Returns((TResult)(object)valueFunction);
-			}
+			}	
 
 			this.setup.AddReturns(() => valueFunction());
+			return this;
+		}
+
+		public ISetupSequentialResult<TResult> ReturnsAsync(Func<TResult> valueFunction)
+		{
+			Guard.NotNull(valueFunction, nameof(valueFunction));
+
+			// If `valueFunction` is `TResult`, that is someone is setting up the return value of a method
+			// that returns a `TResult`, then we have arrived here because C# picked the wrong overload:
+			// We don't want to invoke the passed delegate to get a return value; the passed delegate
+			// already is the return value.
+			if (valueFunction is TResult)
+			{
+				return this.Returns((TResult)(object)valueFunction);
+			}
+
+			this.setup.AddReturns(() => {
+				return Task.Run(() =>
+				{
+					return valueFunction();
+				});
+			});
 			return this;
 		}
 


src/Moq/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using System.ComponentModel;
 using System.Reflection;
-
+using System.Threading.Tasks;
```
stakx: Please add a blank line after this so that `using`s are grouped by their main namespace (i.e. one group for `System.*`, one for `Moq.*`).

src/Moq/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -71,12 +71,34 @@ public ISetupSequentialResult<TResult> Returns(Func<TResult> valueFunction)
 			if (valueFunction is TResult)
 			{
 				return this.Returns((TResult)(object)valueFunction);
-			}
+			}
```
stakx: Whitespace-only change, please revert.

src/Moq/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -71,12 +71,34 @@ public ISetupSequentialResult<TResult> Returns(Func<TResult> valueFunction)
...
 			{
 				return this.Returns((TResult)(object)valueFunction);
-			}
+			}	
 
 			this.setup.AddReturns(() => valueFunction());
 			return this;
 		}
 
+		public ISetupSequentialResult<TResult> ReturnsAsync(Func<TResult> valueFunction)
```
stakx: This doesn't seem right for two reasons:

* You are defining this method on a type that is defined `internal`, without having any `public` counterpart. I would assume that it's actually not possible to call this added method. (Have you tried?)

* By putting this method here, you cannot distinguish between a setup on a method returning `Task`, `ValueTask`, or something else. If you want to make that distinction, you need to implement things as an extension method. See e.g. [the `ThrowsAsync` extension methods on `ISetupSequentialResult<[Value]Task<TResult>>` in `AsyncExtensions.cs`](https://github.com/moq/moq4/blob/bf19ad64540b9e3e71b8a869b2b1ab204ff1dbcb/src/Moq/AsyncExtensions.cs#L275-L299). I believe this is where this new method should be implemented.

src/Moq/Language/Flow/SetupSequencePhrase.cs
```diff
@@ -71,12 +71,34 @@ public ISetupSequentialResult<TResult> Returns(Func<TResult> valueFunction)
...
+			Guard.NotNull(valueFunction, nameof(valueFunction));
+
+			// If `valueFunction` is `TResult`, that is someone is setting up the return value of a method
+			// that returns a `TResult`, then we have arrived here because C# picked the wrong overload:
+			// We don't want to invoke the passed delegate to get a return value; the passed delegate
+			// already is the return value.
+			if (valueFunction is TResult)
+			{
+				return this.Returns((TResult)(object)valueFunction);
+			}
```
stakx: Have you verified (before copy-and-pasting) whether this check still makes sense for your new method?

stakx:
@VladFasie - It's been a while, would you like to keep working on this at all? (If not, please close this PR so we can look for a different solution in #795.)

Issue #355, opened 2017-06-03
Implement a generic extension point in `Times`


stakx:
I'm opening this issue as a place to track a possible feature mentioned by @kzu in a comment in #307:

> I'd rather have a `Times.Matching(Func<int, bool> evaluator, string failMessage)` instead of adding more very specific calculations.

stakx:
I'd already prepared a proposal for this feature but started to have doubts whether it is wise to make `Times` much more flexible.

The problem is that `Times` implements `IEquatable<Times>`, and if we want `Times` to act and feel like a value type, then by-value comparison is going to become really difficult if `Times` is a glorified `Func<int, bool>`. It is like a set type that only has a `Contains` method... how do you compare two such objects for equality by value?!

Options:

1. Don't extend `Times`, keep it closed as it is now.
2. Sacrifice perfect by-value comparability in favour of the flexibility offered by `Times.Matching`. (In that case, the PR mentioned above would be good to go.)
3. Mark `Times`'s implementing of `IEquatable<Times>` as obsolete. Then implement `Times.Matching` in a future version.

stakx:
Regarding the above, it turns out that there would be no direct harmful side effects (judging from a test run) when removing `Equals`, `operator ==` , etc. from `Times`.

Nevertheless, this issue should probably wait until after #21 is resolved, as making `Times` more flexible might make life harder there.

stakx:
(#211 needs to be fixed first, as well.)

stakx:
Closing this as it doesn't have priority at the moment and noone else seems to agree this is needed.

PR #806, opened 2019-04-12
Fix implementation of `Times.Equals`

### LINKED ISSUE #805: `Times.Equals` is implemented incorrectly
### Reproduction code:

```csharp
// see whether the kind of `Times` objects being compared is taken into account:
Assert.Equal(Times.Once(), Times.Exactly(1));

// see whether it is only the effective lower and upper bounds that get compared:
Assert.Equal(Times.Between(2, 3, Range.Inclusive), Times.Between(1, 4, Range.Exclusive));
```

### Actual behavior:

The first assertion passes, but the second one fails.

### Expected behavior:

Either both assertions should fail (meaning that the kind of `Times` being compared matters), or both should pass (meaning that only the effective lower and upper bounds matter).

`Times` is currently implemented in a way that leans towards the latter.

stakx:
Closes #805 (and applies some minor cleanup refactoring to `Times` while we're at it).

### DIFF
--- a/src/Moq/Times.cs
+++ b/src/Moq/Times.cs
@@ -8,7 +8,7 @@ using Moq.Properties;
 namespace Moq
 {
 	/// <include file='Times.xdoc' path='docs/doc[@for="Times"]/*'/>
-	public struct Times
+	public struct Times : IEquatable<Times>
 	{
 		private int from;
 		private int to;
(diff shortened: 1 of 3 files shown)


src/Moq/Times.cs
```diff
@@ -10,14 +10,14 @@ namespace Moq
 	/// <include file='Times.xdoc' path='docs/doc[@for="Times"]/*'/>
 	public struct Times
 	{
-		private Func<int, bool> evaluator;
```
stakx: This makes a future generic extension point such as #355 impossible. That's however not an issue at all, as `Times` today implements equality comparison, and there is `IOccurrence`, both of which wouldn't work well together with a `Times.Matching(Func<int>)`.

PR #812, opened 2019-04-17
Add ToString() for fluent setup API objects

### LINKED ISSUE #810: ToString() on SetupPhrase does not return expression
We have a unit test that verifies the message we receive when a sequenced setup is not run, but with version 4.10+ the message is no longer displaying the expression. Instead, it displays the full name of the type and method, which is the default `ToString()` output. I originally thought this may be an issue with Moq.Sequences, however it seems to be originating inside Moq.

To reproduce the issue here is a simple unit test:
```C#
[Fact]
public void SetupPhraseConvertsExpressionToDescriptiveString()
{
	var setupPhrase = new Mock<IFoo>().Setup(x => x.DoThings(null));

	Assert.Equal("x => x.DoThings(null)", setupPhrase.ToString());
}
```

Expected:
```C#
x => x.DoThings(null)
```
Actual:
```C#
Moq.Language.Flow.VoidSetupPhrase`1[Moq.Tests.Regressions.IssueReportsFixture+IFoo]
```

I can submit a PR to fix this (simply overridding the `ToString()` on `SetupPhrase`), but wanted to see if that would be welcome first. Thanks!

jacob-ewald:
When a `Setup` fails verification, it's helpful to have the exact expression that failed, rather than the default `ToString()` representation of the `Setup` class. To illustrate:

What I'd like to see in the output message:
```C#
x => x.DoThings(null)
```
What I actually get:
```C#
Moq.Language.Flow.VoidSetupPhrase`1[Moq.Tests.Regressions.IssueReportsFixture+IFoo]
```

This PR fixes that by overriding the `ToString()` methods and calling `ToStringFixed()` on the expression that is being used. This allows for a nicely formatted string representation of the expression to be returned.

Closes #810.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ This release contains several **minor breaking changes**. Please review your cod
 * `Verify[All]` fails because of lazy (instead of eager) setup argument expression evaluation (@aeslinger, #711)
 * `ArgumentOutOfRangeException` when setup expression contains indexer access (@mosentok, #714)
 * Incorrect implementation of `Times.Equals` (@stakx, #805)
-
+* Calling `ToString()` on `SetupPhrase` does not return expression (@jacob-ewald, #810)
 
 ## 4.10.1 (2018-12-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ This release contains several **minor breaking changes**. Please review your cod
 * `Verify[All]` fails because of lazy (instead of eager) setup argument expression evaluation (@aeslinger, #711)
 * `ArgumentOutOfRangeException` when setup expression contains indexer access (@mosentok, #714)
 * Incorrect implementation of `Times.Equals` (@stakx, #805)
-* Calling `ToString()` on `SetupPhrase` does not return expression (@jacob-ewald, #810)
+
 
 ## 4.10.1 (2018-12-03)
 
--- a/src/Moq/Language/Flow/SetupPhrase.cs
+++ b/src/Moq/Language/Flow/SetupPhrase.cs
@@ -162,5 +162,10 @@ namespace Moq.Language.Flow
 		{
 			this.setup.Verifiable(failMessage);
 		}
+
+		public override string ToString()
+		{
+			return setup.Expression.ToStringFixed();
+		}
 	}
 }
--- a/tests/Moq.Tests/Regressions/IssueReportsFixture.cs
+++ b/tests/Moq.Tests/Regressions/IssueReportsFixture.cs
@@ -2475,35 +2475,35 @@ namespace Moq.Tests.Regressions
 			[Fact]
 			public void VoidSetupPhraseConvertsExpressionToDescriptiveString()
 			{
-				var voidSetupPhrase = new Mock<IFoo>().Setup(x => x.DoThings(null));
+				var voidSetup = new Mock<IFoo>().Setup(x => x.DoThings(null));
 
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -48,7 +48,7 @@ This release contains several **minor breaking changes**. Please review your cod
 * `Verify[All]` fails because of lazy (instead of eager) setup argument expression evaluation (@aeslinger, #711)
 * `ArgumentOutOfRangeException` when setup expression contains indexer access (@mosentok, #714)
 * Incorrect implementation of `Times.Equals` (@stakx, #805)
-
+* Calling `ToString()` on `SetupPhrase` does not return expression (@jacob-ewald, #810)
```
stakx: This is somewhat misplaced as it is an enhancement (*Added*), not a bug fix (*Fixed*): The ability to inspect a setup while it's being built was never specified.

What's even more important though is that the audience (users of the Moq library) typically won't have a clue what `SetupPhrase` is, nor should they care: It's an implementation detail that's never seen outside the library.

I'd prefer it&mdash;but only if it doesn't make you feel cheated out of your contribution being formally acknowledged&mdash; if we didn't list this change in the changelog at all. It probably doesn't clear the bar for things that users typically need to know about, as the use case is somewhat unusual. (In typical usage, you wouldn't log your setups to the console. You shouldn't even need to store them in a variable! Typically, verification error messages should mention any setups relevant to finding out what's wrong, so the use case we're dealing with here is somewhat unusual to begin with.)

If you'd really like to see this in the changelog despite these counterarguments, then please (a) move it to the *Added* section; (b) make it sound like an addition, not a bug fix, and (c) rephrase it such that it doesn't mention some internal implementation classes.

src/Moq/Language/Flow/SetupPhrase.cs
```diff
@@ -162,5 +162,10 @@ public void Verifiable(string failMessage)
 		{
 			this.setup.Verifiable(failMessage);
 		}
+
+		public override string ToString()
+		{
+			return setup.Expression.ToStringFixed();
```
stakx: Ideally, this would just read `return this.setup.ToString()` in order to avoid duplication of logic that belongs in the setup. But I realise this isn't currently possible, because e.g. `MethodCall` includes a failure message in its `ToString()` formatting.

This should be refactored at some later time, but how you've done it is fine for now. :+1:

CHANGELOG.md
jacob-ewald: I'm not concerned about my name showing up in the changelog, I was just following the contribution guidelines. This change is [reverted](https://github.com/moq/moq4/pull/812/commits/b78daf95a1e673651e96bcde6878f622c45c23fc).

tests/Moq.Tests/Regressions/IssueReportsFixture.cs
```diff
@@ -2466,6 +2466,47 @@ public void InvokeApplyRule(object obj)
...
+		#endregion
+		
+		#region #810
+
+		public class _810
+		{
+			[Fact]
+			public void VoidSetupPhraseConvertsExpressionToDescriptiveString()
+			{
+				var voidSetupPhrase = new Mock<IFoo>().Setup(x => x.DoThings(null));
```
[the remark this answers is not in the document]
jacob-ewald: I [removed](https://github.com/moq/moq4/pull/812/commits/86d47e6c360c53622fc46db0dca5cc061e66406c) the `Phrase` from the variables names.

jacob-ewald:
I sent you some messages on gitter, but wanted to mention them here in case you don't get notifications from gitter.

PR #816, opened 2019-04-23
Fix issue 657 Explicit Interface Implementation

damianpumar:
Fix issue 657, using example reported


damianpumar:
> This solution is tempting because it's so simple! +1
> 
> Help me think this through, please: Will this solve only the very specific problem described in #657, or will it solve the wider class of similar problems? Is this safe, or could this produce any collateral damage? (For example: Someone is explicitly implementing an interface, then requesting to re-implement the same interface using `mock.As<>`, and enabling `CallBase`.) Will it work with `Verify(expr)` as well?

Sorry for merge commit, my version was old!

stakx:
No worries. (You can always rebase, then force-push to your branch to clean up your PR's commits, btw.)

src/Moq/InvocationShape.cs
```diff
@@ -77,6 +77,11 @@ private bool IsOverride(MethodInfo invocationMethod)
 				return false;
 			}
 
+			if (!invocationMethod.IsPublic && method.DeclaringType.GetTypeInfo().IsInterface)
```
lepijohnny: Hi. This might not be a good solution because I think it doesn't solve the problem in general but just one specific case. For example, this scenario will fail:

```
public interface IExplicit
{
	int Bar();
}

public class Explicit : IExplicit
{
	int IExplicit.Bar() => 1;

	public virtual int Bar() => 2;
}

[Fact]
public void ExplicitTest()
{
	var mock = new Mock<Explicit>();
	mock.As<IExplicit>().Setup(m => m.Bar()).CallBase();

	var a = ((IExplicit)mock.Object).Bar();
	var b = mock.Object.Bar();

	Assert.True(a == 1, "a");
	Assert.True(b == 0, "b");
}
```

Assert b will fail.

damianpumar: Yes, I got it.
Do you know how I detect this case, when we have two method with equal name, argument and return type?

Cheers!

lepijohnny: I proposed a different solution at #657. Take a look, I actually used your tests to verify it :bowtie:

lepijohnny: > Yes, I got it.
> Do you know how I detect this case, when we have two method with equal name, argument and return type?
> 
> Cheers!

In this case, you have two methods one is explicitly implemented and the other is a class method. So whenever the interface method has been invoked you should look if that method is explicitly implemented and if so match with that one instead of matching with the class method. In the case method has not been explicitly implemented you could match with the class method...

However, finding if the method has been explicitly implemented is, IMO, a bit awkward...

damianpumar: Yes, is very difficult detect this case, but I was thinking about this case and I think I can detect the explicit implementation.

Cheers!

stakx:
Sorry for the long silence, I still haven't had the time to think this through properly. It seems that your PR would have to also include a solution for the test case added above by @lepijohnny?

I'll take a closer look at that as soon as I can.

src/Moq/InvocationShape.cs
```diff
@@ -91,6 +91,11 @@ private bool IsOverride(MethodInfo invocationMethod)
 				return false;
 			}
 
+			if (!invocationMethod.DeclaringType.IsAssignableFrom(method.DeclaringType))
+			{
+				return false;
+			}
```
stakx: Unfortunately, this is not a workable solution.

Notice how, after this addition, we now have two subsequent checks: 

* `!method.DeclaringType.IsAssignableFrom(invocationMethod.DeclaringType)`
* `!invocationMethod.DeclaringType.IsAssignableFrom(method.DeclaringType)`

`IsAssignableFrom` checks for a subtype-or-same-type relationship, so these checks combined are equivalent to:

* `method.DeclaringType != invocationMethod.DeclaringType`

And this obviously is not the correct thing to do when checking for method overrides, since an overriding method *can* be declared in a type other than that of the overridden method (specifically, in a subtype).

For example, the following test case will start to fail after this change:

```csharp
public interface IFoo
{
	void Method();
}

public class Foo : IFoo
{
	public virtual void Method() { }
}

var mock = new Mock<Foo>();
mock.As<IFoo>().Setup(m => m.Method());  // sets up IFoo.Method
mock.Object.Method();  // calls Foo.Method, which implements IFoo.Method
mock.VerifyAll();  // should therefore pass
```

What your PR has brought into plain sight is the fact that Moq's unit test suite is apparently lacking some very basic tests about method overrides, and how they map to setups. I think that we need to add a rigorous and complete set of unit tests to cover that area before we proceed with fixing this issue.

stakx:
I am going to close this PR for now. Right now, I don't believe we'll be able to find an easy bugfix in the vein of this PR (i.e. a simple addition in `InvocationShape.IsOverride`)... but I'd be happy to be proven wrong.

Unless and until that happens, I'll continue thinking that we should find a bugfix that does its work when a setup is created, not during invocations (as originally stated [here](https://github.com/moq/moq4/issues/657#issuecomment-414128048)).

PR #817, opened 2019-04-26
Explicit interface implementation

lepijohnny:
Hi,

I have created a potential fix for #657.

Nikola


lepijohnny:
Even though this will solve the problem don't believe this naive solution is actually the right one because the state has been introduced into `InvocationShape`. I will close it. Sorry for the inconvenience.

PR #836, opened 2019-06-01
Handle write-only properties in SetupAllProperties for strict mocks (fix #835)

### LINKED ISSUE #835: SetupAllProperties does not setup write-only properties
Opening a new issue as per https://github.com/moq/moq4/pull/826#discussion_r288929791.

**Steps to reproduce**
```csharp
public abstract class WriteOnlyProperty
{
   public abstract string Test { set; }
}

[Fact]
public void Write_only_property_should_be_ignored_by_SetupAllProperties()
{
   var mock = new Mock<WriteOnlyProperty>(MockBehavior.Strict);
   mock.SetupAllProperties();
	
   mock.Object.Test = "test";
}
```

**Expected behavior** 
Test is passed, writing to Test property is ignored.

**Actual behavior** 
Exception is thrown: 
```
Moq.MockException : WriteOnlyProperty.Test = "test" invocation failed with mock behavior Strict.
All invocations on the mock must have a corresponding setup.
```

vanashimko:
Fix #835.

Was not sure, if such a minor change should be included in a changelog, but included just in case.

### DIFF
--- a/tests/Moq.Tests/StubExtensionsFixture.cs
+++ b/tests/Moq.Tests/StubExtensionsFixture.cs
@@ -185,10 +185,12 @@ namespace Moq.Tests
 			public abstract string Test { set; }
 		}
 
-		[Fact]
-		public void Write_only_property_should_be_ignored_by_SetupAllProperties()
+		[Theory]
+		[InlineData(MockBehavior.Strict)]
+		[InlineData(MockBehavior.Loose)]
+		public void SetupAllProperties_should_setup_write_only_properties(MockBehavior mockBehavior)
 		{
-			var mock = new Mock<WriteOnlyProperty>();
+			var mock = new Mock<WriteOnlyProperty>(mockBehavior);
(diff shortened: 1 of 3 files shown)


tests/Moq.Tests/StubExtensionsFixture.cs
```diff
@@ -185,10 +185,12 @@ public abstract class WriteOnlyProperty
 			public abstract string Test { set; }
 		}
 
-		[Fact]
-		public void Write_only_property_should_be_ignored_by_SetupAllProperties()
+		[Theory]
+		[InlineData(MockBehavior.Strict)]
+		[InlineData(MockBehavior.Loose)]
+		public void Write_only_property_should_be_ignored_by_SetupAllProperties(MockBehavior mockBehavior)
```
stakx: This test's name no longer seems right. (If `SetupAllProperties` truly ignored write-only properties, then setting such a property *would* fail for strict mocks.)

PR #844, opened 2019-06-05
Proposal: Prevent capturing arguments on match failure

ocoanet:
Hi.

I would like to improve the capture helper to support the following scenario:
```cs
[Fact]
public void ShouldOnlyCaptureParameterForSpecificArgumentAfterCollection()
{
	var items = new List<string>();
	var mock = new Mock<IFoo>();
	mock.Setup(x => x.DoSomething(Capture.In(items), 1));

	mock.Object.DoSomething("Hello!", 1);
	mock.Object.DoSomething("World", 2);

	var expectedValues = new List<string> { "Hello!" };
	Assert.Equal(expectedValues, items);
}
```

This test currently fails because the capture match is evaluated on every method call, even when other matches do not pass. The goal of this PR is to perform the capture only after all parameter match were successfully evaluated.

The current implementation is a proposal and I am clearly open to suggestions or improvements ideas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * `mock.SetupAllProperties()` now setups write-only properties for strict mocks, so that accessing such properties will not throw anymore. (@vanashimko, #836)
 * Regression: `mock.SetupAllProperties()` and `Mock.Of<T>` fail due to inaccessible property accessors (@Mexe13, #845)
 * Regression: `VerifyNoOtherCalls` causes stack overflow when mock setup returns the mocked object (@bash, #846)
+* `Capture.In()` no longer captures arguments when other setup arguments do not match (@ocoanet, #844).
+* `CaptureMatch` no longer invokes the capture callback when other setup arguments do not match (@ocoanet, #844).
 
 
 ## 4.11.0 (2019-05-28)
--- a/src/Moq/CaptureMatch.cs
+++ b/src/Moq/CaptureMatch.cs
@@ -19,7 +19,7 @@ namespace Moq
 		/// </summary>
 		/// <param name="captureCallback">An action to run on captured value</param>
 		public CaptureMatch(Action<T> captureCallback)
-		: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
+			: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
 		{
 		}
 
@@ -29,7 +29,7 @@ namespace Moq
 		/// <param name="captureCallback">An action to run on captured value</param>
 		/// <param name="predicate">A predicate used to filter captured parameters</param>
 		public CaptureMatch(Action<T> captureCallback, Expression<Func<T, bool>> predicate)
-		: base(BuildCondition(predicate), () => It.Is(predicate), captureCallback)
+			: base(BuildCondition(predicate), () => It.Is(predicate), captureCallback)
 		{
 		}
 
--- a/src/Moq/CaptureMatch.cs
+++ b/src/Moq/CaptureMatch.cs
@@ -12,27 +12,14 @@ namespace Moq
 	/// <typeparam name="T"></typeparam>
 	public class CaptureMatch<T> : Match<T>
 	{
-		private readonly Matcher matcher;
+		private static readonly Predicate<T> matchAllPredicate = _ => true;
 
 		/// <summary>
 		/// Initializes an instance of the capture match.
 		/// </summary>
 		/// <param name="captureCallback">An action to run on captured value</param>
 		public CaptureMatch(Action<T> captureCallback)
-			: this(captureCallback, false)
-		{
-		}
-		
-		/// <summary>
-		/// Initializes an instance of the capture match.
-		/// </summary>
-		/// <param name="captureCallback">An action to run on captured value</param>
-		/// <param name="captureOnSuccessOnly">
-		/// Indicates whether <paramref name="captureCallback"/> should be invoked on parameter evaluation
-		/// or only after all parameters were successfully evaluated.
-		/// </param>
-		public CaptureMatch(Action<T> captureCallback, bool captureOnSuccessOnly)
-			: this(new Matcher(captureCallback, null, captureOnSuccessOnly))
+		: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
 		{
 		}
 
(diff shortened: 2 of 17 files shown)


stakx:
@ocoanet, thanks for submitting this proposal&mdash;it makes total sense! I'll try to review it in detail as soon as I can... might take a few days, though. Please be patient.

Right now, I'm suspecting that Moq's current behavior might simply be buggy: An unmatched setup should in all likelihood stay completely inert and not cause any observable side effects whatsoever. So we might eventually treat your PR as a bug fix, instead of as an enhancement. (The difference of course being that the latter doesn't add e.g. new ctors in `CaptureMatch`.)

I'll get back to you soon.

ocoanet:
Yes, I added overloaded constructors to `CaptureMatch` because I wanted to avoid breaking existing code. But I suppose that `CaptureMatch` is not often used directly, and the current behavior can be considered to be a bug.

Tell me if you want to treat this change as a bug fix, in this case I will remove the new constructors and simplify the code.

src/Moq/CaptureMatch.cs
```diff
@@ -27,30 +42,80 @@ public CaptureMatch(Action<T> captureCallback)
...
+				this.captureCallback = captureCallback;
+
+				if (predicate != null)
+				{
+					this.predicate = predicate.CompileUsingExpressionCompiler();
+					this.RenderExpression = () => It.Is(predicate);
+				}
+				else
+				{
+					this.predicate = _ => true;
```
stakx: I'd prefer if we could avoid unnecessary delegate creation. Why not allow `this.predicate` to be `null`, then use null-conditional `this.predicate?.Invoke(...) == true` later?

src/Moq/CaptureMatch.cs
```diff
@@ -12,12 +12,27 @@ namespace Moq
...
+		/// <summary>
+		/// Initializes an instance of the capture match.
+		/// </summary>
+		/// <param name="captureCallback">An action to run on captured value</param>
+		/// <param name="captureOnSuccessOnly">
+		/// Indicates whether <paramref name="captureCallback"/> should be invoked on parameter evaluation
+		/// or only after all parameters were successfully evaluated.
+		/// </param>
+		public CaptureMatch(Action<T> captureCallback, bool captureOnSuccessOnly)
+			: this(new Matcher(captureCallback, null, captureOnSuccessOnly))
```
stakx: As mentioned earlier, I still think we should treat this as  bug fix; i.e. instead of adding new ctors, let's just change the behavior of the existing ones. (As you said, `Capture.*` are probably a little-used feature so we can probably afford a small breaking change.)

src/Moq/CaptureMatch.cs
```diff
@@ -27,30 +42,80 @@ public CaptureMatch(Action<T> captureCallback)
...
 
-		private static Predicate<T> CreatePredicate(Action<T> captureCallback, Expression<Func<T, bool>> predicate)
+		internal override void OnSuccess()
 		{
-			var predicateDelegate = predicate.CompileUsingExpressionCompiler();
-			return value =>
+			this.matcher.OnSuccess();
+		}
+
+		private class Matcher
```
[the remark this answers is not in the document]
ocoanet: I added this type because `CaptureMatch` inherits from `Match` and the current `Match` constructor needs the condition and the renderExpression. I see two other options:
- Add a parameterless constructor to `Match`.
- Invoke the base constructor using `base(null, null)` and then set the properties.
I will add the parameterless constructor.

src/Moq/CaptureMatch.cs
ocoanet: This kind of "static", non-capturing delegate gets cached by the compiler. See https://sharplab.io/#v2:EYLgHgbALAPgAgJgIwFgBQcDMACR2DCAPACoB82A3utjdtbVtgAoBOApgCYCWAxgIYAXNiXL4A9gDtuArpMrYA5mwEBubAGdlagL70ae3DjhRsAWQAUASkoHa2AQAsu6gHTipXGXIC82APrY3uQCLACubCoGumjaQA==

ocoanet:
I updated the PR:
- `CaptureMatch` is now immutable.
- `CaptureMatch` only captures the argument on success.

ocoanet:
Should I update `Match<T>.Equals` to compare the `SuccessCallback`?

src/Moq/Match.cs
```diff
@@ -33,8 +33,14 @@ internal static TValue Matcher<TValue>()
 
 		internal abstract bool Matches(object value);
 
+		internal virtual void OnSuccess(object value)
+		{
+		}
```
stakx: You could declare this method `abstract` (like the `Matches` method above it), this empty implementation will never actually get used anywhere as `Match<T>` is (and will most likely remain) the sole subclass of `Match`.

src/Moq/CaptureMatch.cs
```diff
@@ -27,30 +29,14 @@ public CaptureMatch(Action<T> captureCallback)
 		/// <param name="captureCallback">An action to run on captured value</param>
 		/// <param name="predicate">A predicate used to filter captured parameters</param>
 		public CaptureMatch(Action<T> captureCallback, Expression<Func<T, bool>> predicate)
-			: base(CreatePredicate(captureCallback, predicate), () => It.Is(predicate))
+		: base(BuildCondition(predicate), () => It.Is(predicate), captureCallback)
```
stakx: Code style: As above, could you please insert this line one more tab?

src/Moq/CaptureMatch.cs
```diff
@@ -12,12 +12,14 @@ namespace Moq
...
 	{
+		private static readonly Predicate<T> matchAllPredicate = _ => true;
+
 		/// <summary>
 		/// Initializes an instance of the capture match.
 		/// </summary>
 		/// <param name="captureCallback">An action to run on captured value</param>
 		public CaptureMatch(Action<T> captureCallback)
-			: base(CreatePredicate(captureCallback), () => It.IsAny<T>())
+		: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
```
stakx: Code style: Could you please insert this line one more tab?

```diff
 		public CaptureMatch(Action<T> captureCallback)
-		: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
+			: base(matchAllPredicate, () => It.IsAny<T>(), captureCallback)
```

src/Moq/Match.cs
ocoanet: I like the idea to leave this method abstract, but it would be a breaking change because `Match` is public.
Are you sure about that?

ocoanet:
I applied your advises and I added the missing unit test. I also rename `OnSuccess` to `SetupEvaluatedSuccessfully` in order to avoid introducing a new term and to make `Condition` and `Match` more similar.

I did not add the abstract method in `Match`, I will do it if you confirm that the breaking change is acceptable.

ocoanet:
By the way, now that `Match<T>` has a success callback, I feel that `CaptureMatch` is almost unnecessary. It could be turned obsolete and replaced by a static factory method `Match<T>.Capture`.

src/Moq/Match.cs
stakx: The type may be `public`, but it is impossible to subclass it outside Moq's own assembly due to the presence of existing `internal abstract` members. So I don't see any harm in adding another abstract member.

(Am I missing something else?)

ocoanet:
Argh, I forgot the changelog again!

stakx:
It appears you've edited an out-of-date version of `CHANGELOG.md` (since there have been a few changes since you opened your PR), can you rebase on top of current `moq/moq4#master`?

CHANGELOG.md
```diff
@@ -21,8 +21,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Adding `Callback` to a mock breaks async tests (@marcin-chwedczuk-meow, #702)
 * `mock.SetupAllProperties()` now setups write-only properties for strict mocks, so that accessing such properties will not throw anymore. (@vanashimko, #836)
-* Regression: `mock.SetupAllProperties()` and `Mock.Of<T>` fail due to inaccessible property accessors (@Mexe13, #845)
-* Regression: `VerifyNoOtherCalls` causes stack overflow when mock setup returns the mocked object (@bash, #846)
+* `Capture.In()` no longer captures arguments when other setup arguments do not match (@ocoanet, #844).
+* `CaptureMatch` no longer invokes the capture callback when other setup arguments do not match (@ocoanet, #844).
```
stakx: Please don't remove existing changelog entries. :wink: (I guess this happened by accident because of the merge conflict...?)

ocoanet: 😅

ocoanet: Done.

stakx:
:rocket: Thank you for your contribution!

PR #854, opened 2019-06-30
Let `mock.Invocations.Clear()` remove traces of earlier invocations more thoroughly

### LINKED ISSUE #733: Invocations.Clear() does not cause Verify to fail.
If an expected function on a Mock is called and then Invocations.Clear() is called on the Mock object, it does not erase the record of the call.  Consequently, a call to a Verify function after the call to Invocations.Clear succeeds.  

I would expect the code to throw an error on the call to an error.
For example:

    try
    {
        Mock<ITest> mockTest = new Mock<ITest>(MockBehavior.Strict);
        mockTest.Setup(t => t.CallMe());
        mockTest.Object.CallMe();
        mockTest.VerifyAll(); // Succeeds as expected.

        mockTest.Invocations.Clear();
        mockTest.VerifyAll(); // Still succeeds

        Assert.Fail("Code should not have reached this point.");
    }
    catch (MockException e)
    {
        Console.WriteLine("Expected exception thrown.");
    }
    catch (AssertFailedException ex)
    {
        Console.WriteLine("Code got to the last line of the try.");
    }

The above code generates the output:

    Code got to the last line of the try.

stakx:
Resolves #733.

### DIFF
--- a/src/Moq/InvocationCollection.cs
+++ b/src/Moq/InvocationCollection.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
+using System.Diagnostics;
 
 namespace Moq
 {
@@ -16,6 +16,14 @@ namespace Moq
 		private int count = 0;
 
 		private readonly object invocationsLock = new object();
+		private readonly Mock owner;
+
+		public InvocationCollection(Mock owner)
+		{
+			Debug.Assert(owner != null);
+
+			this.owner = owner;
+		}
 
 		public int Count
 		{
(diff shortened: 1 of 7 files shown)


src/Moq/InvocationCollection.cs
```diff
@@ -68,6 +76,9 @@ public void Clear()
 				this.invocations = null;
 				this.count = 0;
 				this.capacity = 0;
+
+				this.owner.Setups.UninvokeAll();
```
stakx: Drilling into another property of the owning mock isn't particularly nice. It would be nicer to invoke something on the owning mock itself.

src/Moq/InvocationCollection.cs
```diff
@@ -68,6 +76,9 @@ public void Clear()
...
+				// ^ TODO: Currently this could cause a deadlock as another lock will be taken inside this one!
```
stakx: This raises the question of whether Moq should generally perform more coarse-grained locking.

This would probably mean having just one lock object on the `Mock` instance that would be used both during setups and invocations.

Client code should probably perform all setups first, and only then start invoking the mocked object (possibly from several threads). Code that is structured that way already shouldn't be affected by a move to more coarse-grained locking.

Issue #852, opened 2019-06-27
Improve "Member x does not exist" failure message


thomasfdm:
Currently an `ArgumentException` is thrown when a `.Protected().Verify(methodName, times, arguments)` call is made when `arguments` doesn't match the arguments in methodName's signature which just states `System.ArgumentException : Member Class.Method does not exist.`.

For example, if mocking a HttpMessageHandler you would use the following code to check if SendAsync was called with any (correct type) arguments:

```c#
mockHttpMessageHandler.Protected().Verify(
           "SendAsync",
           Times.Exactly(1),
           ItExpr.IsAny<HttpRequestMessage>(),
           ItExpr.IsAny<CancellationToken>()
);
```

If I was to accidentally use the wrong generic type for one of the arguments (for example `string` instead of `CancellationToken`) my code would instead look like this:
```c#
mockHttpMessageHandler.Protected().Verify(
           "SendAsync",
           Times.Exactly(1),
           ItExpr.IsAny<HttpRequestMessage>(),
           ItExpr.IsAny<string>()
);
```
and, then the test is run would give the following error:
```
System.ArgumentException : Member HttpMessageHandler.SendAsync does not exist.
```

This gives the impression that there is no such method as SendAsync at all in the class, whereas it's only the case that there isn't a SendAsync signature which takes `(HttpRequestMessage, string)`. Might it be worth updating this error message to include the erroneous signature i.e. something like: 
```
System.ArgumentException: Member HttpMessageHandler.SendAsync(HttpRequestMessage, string) does not exist.
```
to make it easier to debug why tests are failing?

stakx:
@thomasfdm, not sure how often people run into this problem, but having good diagnostic messages can't hurt. Would you like to submit a PR?

thomasfdm:
Happy to give it a go, but no definite ETA on it 😄

thomasfdm:
I'll try and get it done today

stakx:
@thomasfdm - One thing to keep in mind (IIRC): It might be wrong to assume that the method parameter list has to match exactly with the provided arguments. (For example, on some `.Protected()....` method overloads, you might request `exactParameterMatch: false`.) So it might be more correct to generate an error message saying, _"No protected method X was found whose parameters are compatible with those provided: (T1, T2, T3, ...)"_, instead of, _"No protected method X(T1, T2, T3, ...) found."_ The latter might be just as misleading (or worse) as what we have now.

thomasfdm:
[...]

PR #862, opened 2019-07-14
Improve Error message given if protected method is setup with bad arguments

thomasfdm:
Fixes issue #852


tests/Moq.Tests/ProtectedMockFixture.cs
```diff
@@ -426,13 +450,37 @@ public void ThrowsIfVerifyEmptyResultMethodName()
...
+		[Fact]
+		public void ThrowsIfVerifyIncompatibleArgumentSupplied()
+		{
+			var actual = Record.Exception(() =>
+			{
+				new Mock<FooBase>().Protected().Verify<string>("StringArg", Times.Once(), ItExpr.IsAny<int>());
+			});
+
+			Assert.IsType<ArgumentException>(actual);
+			Assert.Equal("No Protected Method FooBase.StringArg found whose parameters are compatible with provided types : (System.Int32).", actual.Message);
```
stakx: Once the implementation uses `TypeNameFormatter`, you'll need to change `System.Int32` to `int`. :wink:

src/Moq/Protected/ProtectedMock.cs
```diff
@@ -288,6 +288,33 @@ private static void ThrowIfMemberMissing(string memberName, MemberInfo member)
...
 		}
 
+		private static void ThrowIfMethodMissing(string methodName, MethodInfo method, object[] args)
+		{
+			if (method == null)
+			{
+				List<string> extractedTypeNames = new List<string>();
+				foreach (object o in args)
+				{
+					if (o is System.Linq.Expressions.MethodCallExpression)
```
stakx: While I haven't thought about this in great detail, it seems wrong to have such a very specific special case for `o is MethodCallExpression`. The existing method `ProtectedMock.ToArgTypes(object[] args)` suggests that any value in an `args` array might be one of various kinds of `Expression` (when matchers are used), or some constant value (otherwise). Therefore, it would seem right to change this line to `if (o is Expression expr)`.

thomasfdm: I did wonder about this, cheers for the pointer. You're right, it's probably better to use `Expression`, I hadn't known about it's parent classes and that they'd have a `Type` property too, I've switched it now :)

stakx:
Great work, thanks for your contribution! :rocket:

PR #880, opened 2019-08-08
Fix `ExpressionStringBuilder` formatting of indexer assignment

stakx:
In some cases, `ExpressionStringBuilder` will format an indexer assignment expression `x => x[i1, i2] = v` incorrectly as `x => x[i1, i2, v] = v`.

### DIFF
--- a/src/Moq/ExpressionStringBuilder.cs
+++ b/src/Moq/ExpressionStringBuilder.cs
@@ -325,7 +325,7 @@ namespace Moq
 				else if (node.Method.IsPropertyIndexerSetter())
 				{
 					this.builder.Append('[');
-					AsCommaSeparatedValues(node.Arguments.Skip(paramFrom), ToString);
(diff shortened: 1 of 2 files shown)


src/Moq/ExpressionStringBuilder.cs
```diff
@@ -325,7 +325,7 @@ private void ToStringMethodCall(MethodCallExpression node)
 				else if (node.Method.IsPropertyIndexerSetter())
 				{
 					this.builder.Append('[');
-					AsCommaSeparatedValues(node.Arguments.Skip(paramFrom), ToString);
+					AsCommaSeparatedValues(node.Arguments.Skip(paramFrom).Take(node.Arguments.Count - paramFrom - 1), ToString);
```
stakx: It might be nice to have some kind of `.Slice(...)` extension method, but let's first see whether there would be other use cases in the Moq code base for it. Or, perhaps wait for C# 8 ranges.

Issue #884, opened 2019-08-09
Support expression-based mock creation for constructor calls that takes parameters


frblondin:
Existing mock creation overloads for constructors that take parameters has some limitations:
* Compiler can't check whether the constructor actually exists. Issues/regressions are detected during runtime rather than leveraging code editor capacity
* Editor features such as _Finding references..._ to a constructor is missing calls to the constructor made by these mocks
* When refactoring code, we end up missing those references... and introduce errors (back to first bullet point)

I think it would be relatively simple to add new overloads to `Mock` and `MockRepository` that take an `Expression<Func<T>>` as a parameter. Something like:
```
var mock1 = new Mock<SomeType>(() => new SomeType("foo"));
var mock2 = repository.Create(() => new SomeType("foo"));

```
If you think it's a good idea I could submit a PR. What do you think?

stakx:
Hi @frblondin, that idea would be in line with one of Moq's declared goals (promoting correct code through static typing), so I don't have any objection to adding the new overloads that you're suggesting. Feel free to submit a PR.

> add new overloads to `Mock` and `MockRepository` that take an `Expression<Func<T>>` as a parameter

* The overload should go in `Mock<T>`, not in `Mock` (as I'm sure you realise).

* There has to be a possibility to optionally specify a `MockBehavior`; either via an optional parameter, or an additional overload. Ideally, the parameter order will be based on that of existing constructors / methods.

stakx:
(@frblondin, I have assigned this issue to you and a future version milestone, just to help me track this issue more easily. But don't let that put any pressure on you.)

frblondin:
Just submitted a PR: #888!

stakx:
Implemented in #888.

PR #888, opened 2019-08-11
Added support for lambda expressions while creating a mock

frblondin:
Implements #884.

The following methods have been added:
```
Mock<T>.ctor(Expression<Func<T>> constructorExpression, MockBehavior behavior = MockBehavior.Strict)
MockRepository.Create<T>(Expression<Func<T>> constructorExpression, MockBehavior behavior = MockBehavior.Strict)
```

> Ideally, the parameter order will be based on that of existing constructors / methods.

As per your remark in the issue, I favored the use of an optional parameter for the behavior. Using an optional parameter doesn't allow us to keep the same order because the behavior is necessary the last parameter (the one being optional...).

The implementation works as follows:
* The lambda expression is being analyzed using a new visitor
* I wanted these new methods to successfully get the argument values whatever they use (captured local variables, method calls, constructors, new array inits...). Once the `NewExpression` is visited, an internal lambda will be created & invoked to produce the arguments as an instance of `object[]`

I was about to suggest to mark the existing mock creation method/constructors (taking `params object[]`) as deprecated... but in any case we still need them when creating a mock of an abstract class!

I added unit tests & updated the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
 
-* Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b")` and `repository.Create<SomeType>("a", "b")` (@frblondin, #884)
+* Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b"))` and `repository.Create<SomeType>(() => new SomeType("a", "b"))`. This makes the process of mocking a class without a parameterless constructor simpler (compiler syntax checker...). (@frblondin, #884)
 
 #### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Moq will throw when it detects that an argument matcher will never match anything due to the presence of an implicit conversion operator. (@michelcedric, #897, #898)
 
+* New algorithm for matching invoked methods against methods specified in setup/verification expressions. (@stakx, #904)
+
 #### Added
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
@@ -26,6 +28,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b"))` and `repository.Create<SomeType>(() => new SomeType("a", "b"))`. This makes the process of mocking a class without a parameterless constructor simpler (compiler syntax checker...). (@frblondin, #884)
 
 #### Fixed
+
+* Moq does not mock explicit interface implementation and `protected virtual` correctly. (@oddbear, #657)
 
 * `Invocations.Clear()` does not cause `Verify` to fail (@jchessir, #733)
 
@@ -36,6 +40,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * `Verify` throws `TargetInvocationException` instead of `MockException` when one of the recorded invocations was to an async method that threw. (@Cufeadir, #883)
 
 * Regression in 4.12.0: `SetupAllProperties` removes indexer setups. (@stakx, #901)
+
+* Parameter types are ignored when matching an invoked generic method against setups. (@stakx, #903)
 
 
 ## 4.12.0 (2019-06-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
 
+* Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b")` and `repository.Create<SomeType>("a", "b")` (@frblondin, #884)
+
 #### Fixed
 
 * `Invocations.Clear()` does not cause `Verify` to fail (@jchessir, #733)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Subscription to mocked events used to be handled less strictly than subscription to regular CLI events. As with the latter, subscribing to mocked events now also requires all handlers to have the same delegate type. (@stakx, #891)
 
+* Moq will throw when it detects that an argument matcher will never match anything due to the presence of an implicit conversion operator. (@michelcedric, #897, #898)
+
 #### Added
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
@@ -32,6 +34,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * Regression: `MockDefaultValueProvider` will no longer attempt to set `CallBase` to true for mocks generated for delegates. (@dammejed, #874)
 
 * `Verify` throws `TargetInvocationException` instead of `MockException` when one of the recorded invocations was to an async method that threw. (@Cufeadir, #883)
+
+* Regression in 4.12.0: `SetupAllProperties` removes indexer setups. (@stakx, #901)
 
 
 ## 4.12.0 (2019-06-20)
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -9,7 +9,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
-
+using Moq.Linq;
 using Moq.Properties;
 using Moq.Protected;
 
@@ -420,6 +420,11 @@ namespace Moq
 		public static string ToStringFixed(this Expression expression)
 		{
 			return new StringBuilder().AppendExpression(expression).ToString();
+		}
+
+		public static object[] ExtractConstructorCallArguments<T>(this Expression<Func<T>> constructorExpression)
+		{
+			return ConstructorCallVisitor.ExtractArgumentValues(constructorExpression);
 		}
 	}
 }
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -9,7 +9,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
-using Moq.Linq;
+
 using Moq.Properties;
 using Moq.Protected;
 
@@ -420,11 +420,6 @@ namespace Moq
 		public static string ToStringFixed(this Expression expression)
 		{
 			return new StringBuilder().AppendExpression(expression).ToString();
-		}
-
-		public static object[] ExtractConstructorCallArguments<T>(this Expression<Func<T>> constructorExpression)
-		{
-			return ConstructorCallVisitor.ExtractArgumentValues(constructorExpression);
 		}
 	}
 }
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -339,7 +339,7 @@ namespace Moq
 		/// </summary>
 		public static bool IsProperty(this LambdaExpression expression)
 		{
-			Guard.NotNull(expression, nameof(expression));
+			Debug.Assert(expression != null);
 
 			return expression.Body is MemberExpression memberExpression && memberExpression.Member is PropertyInfo;
 		}
@@ -349,7 +349,7 @@ namespace Moq
 		/// </summary>
 		public static bool IsPropertyIndexer(this LambdaExpression expression)
 		{
-			Guard.NotNull(expression, nameof(expression));
+			Debug.Assert(expression != null);
 
 			return expression.Body is IndexExpression
 				|| (expression.Body is MethodCallExpression methodCallExpression && methodCallExpression.Method.IsSpecialName);
--- a/src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
+++ b/src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
@@ -2,6 +2,7 @@
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
 using System;
+using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -27,28 +28,31 @@ namespace Moq.Expressions.Visitors
 			var visitor = new ConstructorCallVisitor();
 			visitor.Visit(newExpression);
 
-			if (visitor._constructor == null)
+			if (visitor.constructor == null)
 			{
 				throw new NotSupportedException(Resources.NoConstructorCallFound);
 			}
 
-			return visitor._arguments;
+			return visitor.arguments;
 		}
 
-		private ConstructorInfo _constructor;
-		private object[] _arguments;
+		private ConstructorInfo constructor;
+		private object[] arguments;
 
 		public override Expression Visit(Expression node)
 		{
-			switch (node)
+			switch (node.NodeType)
 			{
-				case LambdaExpression _:
-				case NewExpression _:
-				case UnaryExpression unary when unary.NodeType == ExpressionType.Quote:
+				case ExpressionType.Lambda:
+				case ExpressionType.New:
+				case ExpressionType.Quote:
 					return base.Visit(node);
 				default:
 					throw new NotSupportedException(
-						string.Format(Resources.UnsupportedExpressionInConstructorCall, node.NodeType.ToString()));
+						string.Format(
+							CultureInfo.CurrentCulture,
+							Resources.UnsupportedExpression,
+							node.ToStringFixed()));
 			}
 		}
 
@@ -56,7 +60,7 @@ namespace Moq.Expressions.Visitors
 		{
 			if (node != null)
 			{
-				_constructor = node.Constructor;
+				constructor = node.Constructor;
 
 				// Creates a lambda which uses the same argument expressions as the
 				// arguments contained in the NewExpression
@@ -64,7 +68,7 @@ namespace Moq.Expressions.Visitors
 					Expression.NewArrayInit(
 						typeof(object),
 						node.Arguments.Select(a => Expression.Convert(a, typeof(object)))));
-				_arguments = argumentExtractor.Compile().Invoke();
+				arguments = ExpressionCompiler.Instance.Compile(argumentExtractor).Invoke();
 			}
 			return node;
 		}
--- a/src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
+++ b/src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
@@ -1,25 +1,31 @@
-﻿using Moq.Properties;
+﻿// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 
-namespace Moq.Linq
+using Moq.Properties;
+
+namespace Moq.Expressions.Visitors
 {
 	internal class ConstructorCallVisitor : ExpressionVisitor
 	{
 		/// <summary>
 		/// Extracts the arguments from a lambda expression that calls a constructor.
 		/// </summary>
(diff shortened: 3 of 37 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
 
+* Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b")` and `repository.Create<SomeType>("a", "b")` (@frblondin, #884)
```
stakx: * I think it would be great if you mentioned the reason why the new syntax got added, i.e. why it is beneficial when mocking classes without a default parameterless constructor.
* Missing closing parenthesis in `new Mock<SomeType>(() => new SomeType("a", "b")`
* `repository.Create<SomeType>("a", "b")` is a strange example... shouldn't there be a lambda expression somewhere?

src/Moq/Linq/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Moq.Linq
+{
+	internal class ConstructorCallVisitor : ExpressionVisitor
```
stakx: This new class & file do not belong here, `Moq.Linq` is for stuff related to `Mock.Of<>`. Please move it either to `Moq` or (perhaps more accurately) in `Moq.Expressions.Visitors`.

src/Moq/Mock.Generic.cs
```diff
@@ -147,6 +147,21 @@ public Mock(MockBehavior behavior, params object[] args)
...
+		///   Initializes an instance of the mock using the given constructor call including its
+		///   argument values and with a specific <see cref="MockBehavior"/> behavior.
+		/// </summary>
+		/// <param name="constructorExpression">Lambda expression that creates an instance of <typeparamref name="T"/> T.</param>
+		/// <param name="behavior">Behavior of the mock.</param>
+		/// <example>
+		/// <code>var mock = new Mock&lt;MyProvider&gt;(() => new MyProvider(someArgument, 25));</code>
+		/// </example>
+		[SuppressMessage("Microsoft.Usage", "CA2214:DoNotCallOverridableMethodsInConstructors")]
+		public Mock(Expression<Func<T>> constructorExpression, MockBehavior behavior = MockBehavior.Strict)
```
stakx: Please put the `MockBehavior` parameter in first position so that the signature is closer to the existing `public Mock(MockBehavior behavior, params object[] args)`. (Since it's an optional parameter, that means you'll need to add another constructor overload.)

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -5,7 +5,7 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
-
+using System.Linq.Expressions;
 using Moq.Properties;
```
stakx: Please insert a blank line above this line, so that imports are grouped by the main namespace (`System.*`, `Moq.*`).

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -293,6 +293,30 @@ public Mock<T> Create<T>(MockBehavior behavior, params object[] args)
...
+		/// <code>
+		/// var factory = new MockFactory(MockBehavior.Default);
+		/// 
+		/// var mock = factory.Create&lt;MyClass&gt;(() => new MyClass("Foo", 25, true));
+		/// // use mock on tests
+		/// 
+		/// factory.Verify();
+		/// </code>
+		/// </example>
+		public Mock<T> Create<T>(Expression<Func<T>> constructorExpression, MockBehavior behavior = MockBehavior.Strict)
```
stakx: Same as above, please put the `MockBehavior` parameter in first position so this new overload is closer to the already existing `public Mock<T> Create<T>(MockBehavior behavior, params object[] args)`.

src/Moq/Properties/Resources.Designer.cs
```diff
@@ -9,763 +9,781 @@
 //------------------------------------------------------------------------------
 
 namespace Moq.Properties {
-	using System;
```
stakx: I know it's annoying, but could you please manually restore indentation to use tabs instead of spaces, so that your PR's diff only includes what actually changed. Thanks!

(I'll eventually get around to fixing this annoying issue.)

tests/Moq.Tests/MockRepositoryFixture.cs
```diff
@@ -303,5 +313,12 @@ public virtual void BaseMethod()
 				BaseCalled = true;
 			}
 		}
+
+		public class ConcreteClass : BaseClass
```
stakx: Since this class will potentially be shared with / reused by other tests in the future, could you please rename it to something more specific, like e.g. `ClassWithoutParameterlessConstructor`? Thanks!

src/Moq/ExpressionExtensions.cs
```diff
@@ -9,7 +9,7 @@
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
-
+using Moq.Linq;
```
stakx: Once you've moved `ConstructorCallVisitor` out of the `Moq.Linq` namespace, please adjust this import. Also, as mentioned in the other review comment, please insert a blank line above, so that namespace imports are grouped by the main namespaces (`System.*`, `Moq.*`).

src/Moq/Linq/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,66 @@
+﻿using Moq.Properties;
```
stakx: * Please add the copyright notice at the beginning of the file. (See any other source code file for guidance.)
* Please put `System.*` imports before any `Moq.*` imports (and separate the two import groups by a blank line).

src/Moq/ExpressionExtensions.cs
```diff
@@ -421,5 +421,10 @@ public static string ToStringFixed(this Expression expression)
 		{
 			return new StringBuilder().AppendExpression(expression).ToString();
 		}
+
+		public static object[] ExtractConstructorCallArguments<T>(this Expression<Func<T>> constructorExpression)
+		{
+			return ConstructorCallVisitor.ExtractArgumentValues(constructorExpression);
+		}
```
stakx: If this method remains a one-liner, I think you could just as well inline it everywhere.

However, it might be better to keep it, but add some basic validation:

* Throw `ArgumentNullException` for `constructorExpression` if it is `null`.
* Throw `ArgumentException` when `constructorExpression`'s `.Body` does not actually represent a `new` expression.

Optionally document that these exceptions may get thrown in the new `new Mock<T>` / `MockRepository.Create` overloads.

src/Moq/ExpressionExtensions.cs
```diff
@@ -421,5 +421,10 @@ public static string ToStringFixed(this Expression expression)
 		{
 			return new StringBuilder().AppendExpression(expression).ToString();
 		}
+
+		public static object[] ExtractConstructorCallArguments<T>(this Expression<Func<T>> constructorExpression)
```
stakx: Consider renaming the parameter from `constructorExpression` &rarr; `newExpression` (here, and in the added public methods), as the latter would be more accurate: The expressions don't merely refer to a constructor; they should be `new` expressions (which in turn refer to a constructor).

src/Moq/Properties/Resources.resx
```diff
@@ -1,366 +1,372 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-	<!--
```
stakx: Same here, please manually change indentation back from spaces to tabs so that your PR's diff doesn't include the whole file but only what you changed/added.

I know it's annoying, I'll eventually get around to fixing this problem from occurring for good.

src/Moq/Mock.Generic.cs
frblondin: Did the change locally, there is now an ambiguity when using `new Mock<Foo>(null)` between `Mock<T>.Mock(params object[])` and `Mock<T>.Mock(Expression<Func<T>>)`.

I have no strong preference between what I did and what you suggested... but given that it introduces an ambiguity I see no other alternative than keeping what I did.

What do you think?

src/Moq/ExpressionExtensions.cs
frblondin: Finally inlined it... first exception was missing, second exception when `.Body` doesn't represent a `new` expression was already managed.

src/Moq/Mock.Generic.cs
frblondin: Pushed the change in a separate commit so you can see the ambiguity issue...

frblondin: @stakx I'm afraid you were waiting until the PR status gets green again. It's red to show the issue when reverting the order of parameters as suggested. If you have some time please have a look... and sorry if you were off!

stakx: @frblondin, sorry for not having replied yet, I thought that's why the build status was currently marked as failed.

> Did the change locally, there is now an ambiguity when using `new Mock<Foo>(null)` between `Mock<T>.Mock(params object[])` and `Mock<T>.Mock(Expression<Func<T>>)`.

If the C# compiler didn't refuse to compile ambiguous calls i.e. if it just chose one of the two possible methods, we could have made sure ourselves that a call with `null` gets redirected to the `params object[]` overload (preexisting behavior).

But since it completely refuses to compile ambiguous calls, I see no other alternative to what you're suggesting: revert the parameter order back to how you initially had it.

My apologies for not realizing earlier that we'd run into this!

frblondin: > I see no other alternative to what you're suggesting: revert the parameter order back to how you initially had it.

Reverted the last commit.

> My apologies for not realizing earlier that we'd run into this!

No worries, it was really tricky to forecast this type of compiler ambiguity 🔎 .

frblondin:
Looks all good to me. Let me know!

src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+			if (visitor._constructor == null)
+			{
+				throw new NotSupportedException(Resources.NoConstructorCallFound);
+			}
+
+			return visitor._arguments;
+		}
+
+		private ConstructorInfo _constructor;
+		private object[] _arguments;
```
stakx: For consistency with the rest of the codebase, please remove the leading underscores from these fields' names.

src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+
+		public override Expression Visit(Expression node)
+		{
+			switch (node)
+			{
+				case LambdaExpression _:
+				case NewExpression _:
+				case UnaryExpression unary when unary.NodeType == ExpressionType.Quote:
```
stakx: Please rewrite the `switch` to match `node.NodeType` instead of `node`. That'll quite possibly be more efficient and it'll actually simplify the third `case` quite a bit.

src/Moq/Mock.Generic.cs
```diff
@@ -147,6 +147,21 @@ public Mock(MockBehavior behavior, params object[] args)
...
+		///   Initializes an instance of the mock using the given constructor call including its
+		///   argument values and with a specific <see cref="MockBehavior"/> behavior.
+		/// </summary>
+		/// <param name="newExpression">Lambda expression that creates an instance of <typeparamref name="T"/> T.</param>
+		/// <param name="behavior">Behavior of the mock.</param>
+		/// <example>
+		/// <code>var mock = new Mock&lt;MyProvider&gt;(() => new MyProvider(someArgument, 25));</code>
+		/// </example>
+		[SuppressMessage("Microsoft.Usage", "CA2214:DoNotCallOverridableMethodsInConstructors")]
+		public Mock(Expression<Func<T>> newExpression, MockBehavior behavior = MockBehavior.Strict)
```
stakx: * `behavior` must default to `MockBehavior.Default`. It'll just be confusing if Moq sometimes creates strict mocks by default, but loose mocks in other places.

* Moq's public API in general does not use optional parameters, but instead sets default values through different overloads and constructor / method chaining. It would be great if you could follow that pattern (for consistency).

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -293,6 +293,30 @@ public Mock<T> Create<T>(MockBehavior behavior, params object[] args)
...
+		/// <code>
+		/// var factory = new MockFactory(MockBehavior.Default);
+		/// 
+		/// var mock = factory.Create&lt;MyClass&gt;(() => new MyClass("Foo", 25, true));
+		/// // use mock on tests
+		/// 
+		/// factory.Verify();
+		/// </code>
+		/// </example>
+		public Mock<T> Create<T>(Expression<Func<T>> newExpression, MockBehavior behavior = MockBehavior.Strict)
```
stakx: * `behavior` must default to `MockBehavior.Default` (reason given in a previous review comment).

* Moq's public API in general does not use optional parameters, but instead sets default values through different overloads and constructor / method chaining. It would be great if you could follow that pattern (for consistency).

tests/Moq.Tests/MockFixture.cs
```diff
@@ -467,6 +468,61 @@ public void ConstructsClassWithNoDefaultConstructorAndNullValue()
...
+
+			Assert.Equal("Hello", mock.Object.StringValue);
+			Assert.Equal(26, mock.Object.IntValue);
+		}
+
+		[Fact]
+		public void ConstructsObjectsWithCtorLambda_NoParameter()
+		{
+			var mock = new Mock<FooOverrideEquals>(
+				() => new FooOverrideEquals(),
```
stakx: It's definitely good to have a test for this use case... but it makes me wonder whether Moq should throw an `ArgumentException` in this case, since the lambda is completely unnecessary and the code could be written more simply. (An analyzer would be better suited for this kind of thing but we still haven't got one.)

That's not a request, just a thought. What's your opinion?

src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+					return base.Visit(node);
+				default:
+					throw new NotSupportedException(
+						string.Format(Resources.UnsupportedExpressionInConstructorCall, node.NodeType.ToString()));
```
stakx: * In cases where a string resource contains placeholders (`{0}` etc.), the Moq codebase usually explicitly passes `CultureInfo.CurrentCulture` to `string.Format`. (I didn't come up with that one, but:) Please follow that convention for consistency.

* Instead of reporting just the node type, it would be better to report the whole sub-expression that wasn't understood (`node.ToStringFixed()`). You could even just reuse the preexisting resource `Resources.UnsupportedExpression`.

src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
```diff
@@ -0,0 +1,72 @@
...
+			{
+				_constructor = node.Constructor;
+
+				// Creates a lambda which uses the same argument expressions as the
+				// arguments contained in the NewExpression
+				var argumentExtractor = Expression.Lambda<Func<object[]>>(
+					Expression.NewArrayInit(
+						typeof(object),
+						node.Arguments.Select(a => Expression.Convert(a, typeof(object)))));
+				_arguments = argumentExtractor.Compile().Invoke();
```
stakx: Please compile the expression via Moq's `ExpressionCompiler.Instance.Compile(...)` extension point (again, for consistency with the rest of the codebase).

src/Moq/Mock.Generic.cs
```diff
@@ -147,6 +147,21 @@ public Mock(MockBehavior behavior, params object[] args)
 			this.CheckParameters();
 		}
 
+		/// <summary>
+		///   Initializes an instance of the mock using the given constructor call including its
+		///   argument values and with a specific <see cref="MockBehavior"/> behavior.
+		/// </summary>
+		/// <param name="newExpression">Lambda expression that creates an instance of <typeparamref name="T"/> T.</param>
```
stakx:
```diff
-an instance of <typeparamref name="T"/> T
+an instance of <typeparamref name="T"/>
```

src/Moq/Obsolete/MockFactory.cs
```diff
@@ -293,6 +293,30 @@ public Mock<T> Create<T>(MockBehavior behavior, params object[] args)
 			return CreateMock<T>(behavior, args);
 		}
 
+		/// <summary>
+		/// Creates an instance of the mock using the given constructor call including its
+		/// argument values and with a specific <see cref="MockBehavior"/> behavior.
+		/// </summary>
+		/// <typeparam name="T">Type to mock.</typeparam>
+		/// <param name="newExpression">Lambda expression that creates an instance of <typeparamref name="T"/> T.</param>
```
stakx: Same as in the other review comment, there's a superfluous `T` after the `<typeparamref/>`.

src/Moq/Expressions/Visitors/ConstructorCallVisitor.cs
stakx: (P.S., speaking of resources: You'll notice that you have a merge conflict in `Resources.resx`. I've removed a few superfluous string resources since you opened this commit. Please make sure you don't reintroduce those when rebasing & resolving the conflict.)

tests/Moq.Tests/MockFixture.cs
frblondin: Don't necessarily aligned with you on this: if I were to write test on a type that likely could receive new parameters in the future (new DI parameters or simple parameters), I could decide to use the lambda-based syntax as an efficient way to get compiler errors if my tests have to be updated.

I believe it's a call of the test writer to use this syntax or the previous one.

src/Moq/Mock.Generic.cs
frblondin: Tried your second suggestion - unfortunately this creates once again a ambiguity with Mock<T>.Mock(params object[])...

frblondin:
I integrated most your changes: except the new overloads to avoid use of optional parameters (creates ambiguities) and your question about use of lambda syntax for parameter-less constructors.

Hope this version will be the good one ;-)

tests/Moq.Tests/MockRepositoryFixture.cs
stakx: You resolved this review comment, but didn't actually change the class' name. Instead, you changed the name of the *other* `ConcreteClass` (in `IssueReportsFixture.cs`) to `ClassWithoutParameterlessConstructor` even though it has a default constructor...

(P.S. on a side note, in future PRs, I'd appreciate it if you didn't resolve review comments yourself; let the reviewers perform that step. It should be up to the original author of a review comment to decide whether their comment / request has been sufficiently addressed. Letting reviewers resolve comments makes it easier for them to review your PRs because it will take less time to keep an overview of what remains to be done. :wink:)

tests/Moq.Tests/Regressions/IssueReportsFixture.cs
```diff
@@ -320,7 +320,7 @@ public interface ISomeInterface
 				void Method();
 			}
 
-			public class ConcreteClass : ISomeInterface
+			public class ClassWithoutParameterlessConstructor : ISomeInterface
```
stakx: Incorrect class name, see other review comment above.

src/Moq/Properties/Resources.resx
```diff
@@ -1,348 +1,351 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema
```
stakx: I am sorry to have to bother you again about whitespace... but I'm afraid I need to ask you again to revert the whitespace from tabs back to spaces so we see the actual changes to this file. 🙈

(This is my bad&mdash;after the first review of this PR, I adjusted this project's `.editorconfig` file such that it matches what VS autogenerates, so that this annoying problem would go away in the future. I should have waited with that until after this PR gets merged, then fixed up the whitespace myself. I'd like to apologise for that.)

tests/Moq.Tests/Regressions/IssueReportsFixture.cs
frblondin: > (P.S. on a side note, in future PRs, I'd appreciate it if you didn't resolve review comments yourself; let the reviewers perform that step. It should be up to the original author of a review comment to decide whether their comment / request has been sufficiently addressed. Letting reviewers resolve comments makes it easier for them to review your PRs because it will take less time to keep an overview of what remains to be done. 😉)

Sorry... I am never sure as of how to resolve conversations. There should be a two-step validation (_Taken into account_ from PR author, _Validated_ from reviewer).

frblondin: Changed in upcoming commit.

src/Moq/Properties/Resources.resx
frblondin: My git diff was ignoring whitespace/tab changes. Didn't notice there was such a problem... sorry too!

frblondin:
New round of updates, I believe your remaining change requests have been taken into account - you to confirm & resolve!

Thanks for you time.

src/Moq/Properties/Resources.resx
```diff
@@ -120,6 +123,35 @@
...
+
+
+
+
+
+
+
+
+
+
```
stakx: There shouldn't be any blank lines (65-67 and 126-154).

Perhaps I should explain why I am being so peculiar about whitespace. This is a file generated & updated by tooling. Upon the next run, these blank lines will very likely be removed by the resources editor, causing whitespace changes yet again. So whoever introduces whitespace changes here will cause them also for the next committer.

PR #890, opened 2019-08-14
New Reset Extensions

Rayffer:
The aim of this pull request is to provide new and isolated ways of resetting properties of the mock selectively, the parts that can be resetted selectively now are the following:

- ConfiguredDefaultValues
- Setups
- EventHandlers
- Invocations

This last one is the same as ResetCalls, which is obsolete so it could be ignored in the pull request.

A set of unit test have been added to demonstrate the behaviour of each of the isolated resets.

One thing to take into consideration is that Reset has not been thoroughly tested, specifically, the part about invocations and eventhandlers has not been tested

### DIFF
--- a/src/Moq/MockExtensions.cs
+++ b/src/Moq/MockExtensions.cs
@@ -23,7 +23,6 @@ namespace Moq
 			mock.Invocations.Clear();
 		}
 
-
 		/// <summary>
 		/// Reset the mock's configured default values
 		/// </summary>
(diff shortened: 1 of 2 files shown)


stakx:
Hi @Rayffer, and thank you for contributing. I'd prefer to not merge this PR in its present form, nor at the present time. That's basicslly for the same reason why `mock.ResetCalls` has been deprecated: I'd rather expose the `Clear` methods on the various collections directly, which however requires exposing the collections first (as was already done for `Invocations`). If we add these methods now, we'll only have to deprecate them again.

Exposing a collection for, say, `Setups`, includes exposing a type for individual setups. I'm working on this and hopefully we can expose setups in ca. Moq version  4.14.0.

Rayffer:
Good to know you're working on it and thanks for your time, I tried to expose them but didn't have the time to get to know your stack to do something that works.

PR #891, opened 2019-08-14
Make mock event subscription equally strict as regular event subscription

stakx:


src/Moq/EventHandlerCollection.cs
```diff
@@ -3,33 +3,23 @@
...
 			{
-				List<Delegate> handlers;
-				if (!this.eventHandlers.TryGetValue(eventName, out handlers))
-				{
-					handlers = new List<Delegate>();
-					this.eventHandlers.Add(eventName, handlers);
-				}
-
-				handlers.Add(eventHandler);
+				this.eventHandlers[eventName] = Delegate.Combine(this.TryGet(eventName), eventHandler);
```
stakx: Calling the public `TryGet` method will take the same lock once more. Try to prevent this. Same in `Remove`.

Issue #898, opened 2019-08-17
Setup method DateTimeOffset vs DateTime


michelcedric:
During a setup if a method take a parameter of type DateTimeOffset but defined
It.IsAny<DateTime>() and not It.IsAny<DateTimeOffset>()
After that the mtehod return null.
I can understand it's not the strict same type. But it compile and there no exception with a correct message but just return null.
Could you block the compilation or raise a good exception with a specific message. 

https://github.com/moq/moq4/issues/897

Best regards

Cédric Michel

stakx:
Please don't post duplicate issues.

stakx:
P.S.:

> Could you block the compilation or raise a good exception with a specific message.

No. `DateTimeOffset` declares an implicit conversion operator to `DateTime`. Know your APIs. There's nothing we can do about that on Moq's side.

michelcedric:
`var serviceMock = new Mock<IMeteringDataServiceAgent>();
            serviceMock.Setup(s => s.GetDataList(It.IsAny<DateTime>())).Returns(_data);
            var result3 = serviceMock.Object.GetDataList(new DateTime(2019,1,1));

            //Expected the list but received null
            //Because during the setup I put It.IsAny<DateTime>() and not It.IsAny<DateTimeOffset>()
            Assert.AreEqual(_data, result3);`

michelcedric:
It's not a duplicate in my repository I have 2 test case with 2 different issue
1 about return null for IEnumerable but not for a List and I think it's a strange behavior
2 Setup and invocation with sub type DateTime or in the Mocked interface is DateTimeOffset and he return null.

stakx:
See your other issue, to which you linked from this one, for an answer regarding all of these tests.

PR #900, opened 2019-08-17
Fail when implicit conversion operator renders argument matcher unmatchable

### LINKED ISSUE #897: Method without setup not return null
Hi,
I make a MockTestCase project : https://github.com/michelcedric/MockTestCase/blob/master/MockTest/UnitTest1.cs

When a method without setup is defined normaly he returns null.
But If the return it's a IEnumerable, it's return an empty list.

I think it's a strange behavior and I don't understand why not return null?

Best regards

Cédric Michel

stakx:
Closes #897, #898.

### DIFF
--- a/src/Moq/MatcherFactory.cs
+++ b/src/Moq/MatcherFactory.cs
@@ -11,6 +11,8 @@ using System.Reflection;
 using Moq.Matchers;
 using Moq.Properties;
 
+using TypeNameFormatter;
+
 namespace Moq
 {
 	internal static class MatcherFactory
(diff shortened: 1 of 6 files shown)


stakx:
@michelcedric: With this proposed change, your `DateTimeNotWorking` test from #897 would fail here:

> ```csharp
> serviceMock.Setup(s => s.GetDataList(It.IsAny<DateTime>())).Returns(_data);
> ```

with the following message:

> **System.ArgumentException**: Matcher `It.IsAny<DateTime>()` is unmatchable: An implicit conversion operator will convert arguments of type `DateTime` to the parameter's type `DateTimeOffset`, which is assignment-incompatible.

Do you think this will be a clear enough indication of why the setup is written incorrectly?

michelcedric:
Yes thanks

src/Moq/MatcherFactory.cs
```diff
@@ -33,6 +35,23 @@ internal static class MatcherFactory
 
 		public static Pair<IMatcher, Expression> CreateMatcher(Expression argument, ParameterInfo parameter)
 		{
+			if (argument.NodeType == ExpressionType.Convert)
+			{
+				var convertExpression = (UnaryExpression)argument;
+				if (convertExpression.Method?.Name == "op_Implicit")
```
stakx: Perhaps we could take `op_Implicit` handling even further and evaluate it if it isn't applied to an argument matcher.

tests/Moq.Tests/UnmatchableMatchersFixture.cs
```diff
@@ -0,0 +1,48 @@
...
+			var mock = new Mock<IX>();
+			mock.Setup(x => x.USeNullableLong(It.IsAny<long>()));
+		}
+
+		public interface IX
+		{
+			void UseDateTimeOffset(DateTimeOffset arg);
+			void UseInt(int arg);
+			void UseLong(long arg);
+			void USeNullableLong(long? arg);
```
stakx: Typo: `USe` &rarr; `Use`

tests/Moq.Tests/UnmatchableMatchersFixture.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Moq.Tests
+{
+	public class UnmatchableMatchersFixture
+	{
+		[Fact]
+		public void Matchers_that_are_unmatchable_due_to_implicit_conversion_operator_cause_setup_failure()
+		{
+			var mock = new Mock<IX>();
+			var ex = Assert.Throws<ArgumentException>(() => mock.Setup(x => x.UseDateTimeOffset(It.IsAny<DateTime>())));
```
stakx: For the sake of these tests' independence from external factors, perhaps they shouldn't rely on external types (even ones from the BCL). It might be wiser to define our own types with a `implicit operator`. This would make the operator directly visible and thus make the problem at hand easier to understand.

PR #902, opened 2019-08-22
Fix `SetupAllProperties` indexer regression & refactor accessor check methods for clarify

### LINKED ISSUE #901: Regression in 4.12.0: `SetupAllProperties` resets indexer setups
### Code to reproduce:

```csharp
public interface IX
{
    string this[int index] { get; }
}

[Fact]
void SetupAllProperties_should_not_reset_indexer_setups()
{
    var mock = new Mock<IX>();
    mock.SetupGet(x => x[1]).Returns("value from setup");
    Assert.Equal("value from setup", mock.Object[1]);  // passes
    mock.SetupAllProperties();
    Assert.Equal("value from setup", mock.Object[1]);  // fails
}
```

This is caused by this line:

https://github.com/moq/moq4/blob/81d1fb24b5a9c5ceb443114b1657137501c4f9e5/src/Moq/SetupCollection.cs#L62

The filter condition should be more specific and rule out indexer accessors.

stakx:
Fixes #901.

### DIFF
--- a/src/Moq/SetupCollection.cs
+++ b/src/Moq/SetupCollection.cs
@@ -32,7 +32,7 @@ namespace Moq
 		{
 			lock (this.setups)
 			{
-				if (setup.Method.LooksLikeEventAttach() || setup.Method.LooksLikeEventDetach())
+				if (setup.Method.IsEventAddAccessor() || setup.Method.IsEventRemoveAccessor())
 				{
 					this.hasEventSetup = true;
 				}
@@ -59,7 +59,7 @@ namespace Moq
 
 			lock (this.setups)
 			{
-				this.setups.RemoveAll(x => x.Method.IsPropertyAccessor() && !x.Method.IsPropertyIndexerAccessor());
+				this.setups.RemoveAll(x => x.Method.IsPropertyAccessor());
 				this.overridden = 0U;
 			}
 		}
--- a/src/Moq/SetupCollection.cs
+++ b/src/Moq/SetupCollection.cs
@@ -59,7 +59,7 @@ namespace Moq
 
 			lock (this.setups)
 			{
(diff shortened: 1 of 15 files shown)


src/Moq/StringBuilderExtensions.AppendExpression.cs
```diff
@@ -327,39 +328,49 @@ private static StringBuilder AppendExpression(this StringBuilder builder, Method
...
-					builder.Append('.')
-					       .Append(expression.Method.Name, 4);
-					if (expression.Arguments.Count > paramFrom)
+					if (expression.Method.IsPropertyAccessor())
+					{
+						builder.Append('.')
+						       .Append(expression.Method.Name, 4);
+						if (expression.Arguments.Count > paramFrom)
+						{
+							builder.AppendCommaSeparated("[", expression.Arguments.Skip(paramFrom), AppendExpression, "]");
```
stakx: Not introduced by this refactoring, but weird nonetheless: Why do angle brackets appear in the formatting of a property getter expression?

src/Moq/SetupCollection.cs
```diff
@@ -32,7 +32,7 @@ public void Add(Setup setup)
 		{
 			lock (this.setups)
 			{
-				if (setup.Method.LooksLikeEventAttach() || setup.Method.LooksLikeEventDetach())
+				if (setup.Method.IsEventAddAccessor() || setup.Method.IsEventRemoveAccessor())
```
stakx: For consistency, add a `.IsEventAccessor()` check method.

stakx: Left as is for now, as the new method would be used in only two places, and not introduce any opportunities for optimization.

PR #904, opened 2019-08-24
New algorithm for matching invoked methods against expected methods

### LINKED ISSUE #657: Moq does not mock 'Explicit Interface Implementation' and 'protected virtual' correctly

### LINKED ISSUE #903: Parameter types are ignored when matching an invoked generic method against setups
### Code to reproduce:

```csharp
public interface IX
{
	void M<T>(Exception arg);
	void M<T>(Uri arg);
}

[Fact]
public void Test()
{
    var exceptionSetupInvoked = false;
    var mock = new Mock<IX>();
    mock.Setup(m => m.M<object>((Exception)null)).Callback(() => exceptionSetupInvoked = true);
    mock.Setup(m => m.M<object>((Uri)null)).Callback(() => throw new Exception("Wrong setup matched."));

    mock.Object.M<int>((Exception)null);

    Assert.True(exceptionSetupInvoked);
}
```

### Expected outcome:

The test should succeed.

### Actual outcome:

The test throws the `"Wrong setup matched."` exception.

### Further information:

While the concrete generic arguments in the above test are irrelevant (since Moq matches generic arguments by assignment compatibility), the presence of generic type parameters are central to this issue as they disable any parameter type checks that would otherwise happen. Note the `else` in the following:

https://github.com/moq/moq4/blob/3868424bac43b2c08342b363eb6c511fdbc982ea/src/Moq/InvocationShape.cs#L113-L126

As a side note, setup order also matters for this test.

stakx:
This replaces the machinery that was responsible for determining whether an invoked method matches that encoded in a setup or verification expression. This previous machinery is very ad-hoc and performs a bunch of tests, such as method name equality, assignment compatibility, and parameter list equality. 

Instead of all that, why not let the runtime do that work by leveraging `method.GetBaseDefinition()`, then simply comparing `MethodInfo`s? (There are some finer points, such as converting generic methods to their (open) generic method definitions; mapping interface methods to the concrete proxy type's methods that provide their implementations; special handling for delegate invocations; etc.)

Fixes #657. Supercedes #658. Fixes #903.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,15 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Moq will throw when it detects that an argument matcher will never match anything due to the presence of an implicit conversion operator. (@michelcedric, #897, #898)
 
+* New algorithm for matching invoked methods against methods specified in setup/verification expressions. (@stakx, #904)
+
 #### Added
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
 
 #### Fixed
+
+* Moq does not mock explicit interface implementation and `protected virtual` correctly. (@oddbear, #657)
 
 * `Invocations.Clear()` does not cause `Verify` to fail (@jchessir, #733)
 
@@ -34,6 +38,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 * `Verify` throws `TargetInvocationException` instead of `MockException` when one of the recorded invocations was to an async method that threw. (@Cufeadir, #883)
 
 * Regression in 4.12.0: `SetupAllProperties` removes indexer setups. (@stakx, #901)
+
+* Parameter types are ignored when matching an invoked generic method against setups. (@stakx, #903)
 
 
 ## 4.12.0 (2019-06-20)
--- a/src/Moq/Extensions.cs
+++ b/src/Moq/Extensions.cs
@@ -26,6 +26,44 @@ namespace Moq
 			return type.IsValueType ? Activator.CreateInstance(type) : null;
 		}
 
+		/// <summary>
+		///   Gets the least-derived <see cref="MethodInfo"/> in the given type that provides
+		///   the implementation for the given <paramref name="method"/>.
+		/// </summary>
+		public static MethodInfo GetImplementingMethod(this MethodInfo method, Type proxyType)
+		{
+			Debug.Assert(method != null);
+			Debug.Assert(proxyType != null);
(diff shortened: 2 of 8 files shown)


stakx:
One thing that isn't explicitly mentioned above is that Moq no longer considers the return type as part of a method's signature like it used to (and that is generally a good thing). This can have some consequences in certain scenarios, see e.g. #922. Marking this PR as _breaking change_ to make this more visible.

PR #905, opened 2019-08-24
Remove `memberExpression.GetReboundProperty()`

stakx:
This helper method has apparently become a no-op with the recent rewrite of the method matching algorithm in `InvocationShape`. As the code being removed explains in a code comment, the Roslyn compilers put base definitions in their emitted LINQ expression trees. Moq's new method mapping algorithm now also favors least-derived methods aka base definitions, which might explain why `.GetReboundProperty()` is now no longer needed.


stakx:
Hah! Not so redundant, after all! Interestingly, VS 2019's Test Runner didn't report any failing tests... besides being dog-slow, blocking the compiler, and crashing all the time. 💩 I think it's time for an uninstall.

PR #908, opened 2019-08-25
Add support for generic type argument matchers (It.IsAnyType and custom matchers)

### LINKED ISSUE #343: Open Generic Method Callback

stakx:
These commits add support for generic type argument matching. 😎

For starters, the only built-in type matcher is `It.IsAnyType`:

```csharp
mock.Setup(m => m.Method<It.IsAnyType>(...))...
```

### Creating a custom type matcher:

The type matcher infrastructure is designed to be fully extensible:

```csharp
class Exceptions : ITypeMatcher
{
    public bool Matches(Type typeArgument) => typeof(Exception).IsAssignableFrom(typeArgument);
}

handler.Setup(h => h.Handle<Exceptions>())
       .Callback(() => Console.WriteLine("Handled an exception.");
```

This allows matchers to satisfy generic type contraints:

```csharp
public interface IJuggler
{
    void Juggle<TBall>() where TBall : Ball
}

class AnyBall : Ball, ITypeMatcher
{
    public bool Matches(Type typeArgument) => true;
}

new Mock<IJuggler>().Setup(j => j.Juggle<AnyBall>())...;
```

When the type is constrained to something that won't allow you to implement `ITypeMatcher`, there's `[TypeMatcher]`:

```csharp
public interface IRedirector
{
    void Redirect<TDelegate>(TDelegate handler) where TDelegate : Delegate;
}

[TypeMatcher(typeof(It.IsAnyType))]
public delegate void AnyDelegate();

var redirector = new Mock<IRedirector>();
redirector.Setup(r => r.Redirect<AnyDelegate>(...)))...
```

### Accessing the type arguments in `Callback` and `Returns`:

(@michal-ciechan, this one is for you: You requested that some "invocation context" be injected into `Callback` and `Returns` callbacks, it turns out Moq already has a suitable type for that: `IInvocation`.)

In order for `Callback` and `Returns` to be able to know which type arguments were used, there are two new overloads that accept a `InvocationAction` and `InvocationFunc`, respectively:

```csharp
public interface IFactory
{
    T Create<T>();
}

var factory = new Mock<IFactory>();
factory.Setup(f => f.Create<It.IsAnyType>())
       .Returns(new InvocationFunc(invocation =>
                {
                    var type = invocation.Method.GetGenericTypeArguments()[0];
                    return Activator.CreateInstance(type);
                });

var something = factory.Object.Create<Something>();
```

These two types are designed as delegate type look-alikes, however they aren't actually delegates. This was necessary to prevent the C# compiler from starting to pick the wrong `Callback` and `Returns` overloads in everyday use cases.

### Matching arguments of a generic type:</summary>

```csharp
interface IFrobbler
{
    void Frobble<T>(T arg);
}

var frobbler = new Mock<IFrobbler>();
frobbler.Setup(f => f.Frobble(It.IsAny<It.IsAnyType>()))
        .Callback((object arg) => ...);
```

### Things left to do:

* [X] Add support for `It.IsAny<It.IsAnyType>`. Or, more generally speaking: `It.IsAny<TTypeMatcher>`, `It.IsNotNull<TTypeMatcher>`, etc.

* [X] If some or all of the above is implemented, add tests for the `Capture` code paths. (More specifically, I am worrying about the type conversions `(T)` hidden inside the `Match` class. Those will generally fail when `T` is a type matcher.)

   **Update:** This is only partially fulfilled. While `Capture.*` shouldn't throw, `CaptureMatch<T>` won't work just yet. This is left as a TODO for another time.

* [ ] ~~Add support for composite types making use of type matchers, e.g. `It.IsAnyType?`, `It.IsAnyType[]`, `It.IsAnyType[,]`, `IEnumerable<It.IsAnyType>`, `ref`/`in`/`out It.IsAnyType`, etc. Same for custom matcher types. It's possible that this cannot be made to work with custom argument matchers.~~

   **Update:** While that would be great, it would also negatively affect performance as every type parameter would have to be decomposed just to discover whether it includes a type matcher. For now, we're going the opposite direction and try to make type matcher discovery as fast as possible.

Closes #343.

/cc @michal-ciechan, @oddbear, @kzu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,36 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Added support for lambda expressions while creating a mock through `new Mock<SomeType>(() => new SomeType("a", "b"))` and `repository.Create<SomeType>(() => new SomeType("a", "b"))`. This makes the process of mocking a class without a parameterless constructor simpler (compiler syntax checker...). (@frblondin, #884)
 
+* Support for matching generic type arguments: `mock.Setup(m => m.Method<It.IsAnyType>(...))`. (@stakx, #908)
+
+   The standard type matchers are:
+
+   - `It.IsAnyType` &mdash; matches any type
+   - `It.IsSubtype<T>` &mdash; matches `T` and proper subtypes of `T`
+   - `It.IsValueType` &mdash; matches only value types
+
+   You can create your own custom type matchers:
+
+   ```csharp
+   [TypeMatcher]
+   class Either<A, B> : ITypeMatcher
+   {
+       public bool Matches(Type type) => type == typeof(A) || type == typeof(B);
+   }
+   ```
+
+* In order to support type matchers (see bullet point above), some new overloads have been added to existing methods:
+
+   - `setup.Callback(new InvocationAction(invocation => ...))`,  
+     `setup.Returns(new InvocationFunc(invocation => ...))`:
+
+      The lambda specified in these new overloads will receive an `IInvocation` representing the current invocation from which type arguments as well as arguments can be discovered.
+
+   - `Match.Create<T>((object argument, Type parameterType) => ..., ...)`,  
+     `It.Is<T>((object argument, Type parameterType) => ...)`:
+
+      Used to create custom matchers that work with type matchers. When a type matcher is used for `T`, the `argument` received by the custom matchers is untyped (`object`), and its actual type (or rather the type of the parameter for which the argument was passed) is provided via an additional parameter `parameterType`. (@stakx, #908)
+
 #### Fixed
 
 * Moq does not mock explicit interface implementation and `protected virtual` correctly. (@oddbear, #657)
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -208,7 +208,7 @@ namespace Moq
 							{
 								if (typeArgument.IsTypeMatcher(out var typeMatcherType))
 								{
-									Guard.HasDefaultConstructor(typeMatcherType);
+									Guard.CanCreateInstance(typeMatcherType);
 								}
 							}
 						}
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -208,7 +208,7 @@ namespace Moq
 							{
 								if (typeArgument.IsTypeMatcher(out var typeMatcherType))
 								{
-									Guard.CanCreateInstance(typeMatcherType);
+									Guard.ImplementsTypeMatcherProtocol(typeMatcherType);
 								}
 							}
 						}
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -206,9 +206,9 @@ namespace Moq
 						{
 							foreach (var typeArgument in methodCallExpression.Method.GetGenericArguments())
 							{
-								if (typeArgument.IsTypeMatcher())
+								if (typeArgument.IsTypeMatcher(out var typeMatcherType))
 								{
-									Guard.HasDefaultConstructor(typeArgument);
+									Guard.HasDefaultConstructor(typeMatcherType);
 								}
 							}
 						}
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -201,6 +201,18 @@ namespace Moq
 					case ExpressionType.Call:  // regular method call
 					{
 						var methodCallExpression = (MethodCallExpression)e;
+
+						if (methodCallExpression.Method.IsGenericMethod)
+						{
+							foreach (var typeArgument in methodCallExpression.Method.GetGenericArguments())
+							{
+								if (typeArgument.IsTypeMatcher())
+								{
+									Guard.HasDefaultConstructor(typeArgument);
+								}
+							}
+						}
+
 						if (!methodCallExpression.Method.IsStatic)
 						{
 							r = methodCallExpression.Object;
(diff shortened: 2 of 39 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,13 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 * Added support for setup and verification of the event handlers through `Setup[Add|Remove]` and `Verify[Add|Remove|All]` (@lepijohnny, #825) 
 
+* Support for matching generic type arguments: `mock.Setup(m => m.Method<It.IsAnyType>(...))`. Create your own type matchers by defining a type that implements `ITypeMatcher`: `class ExceptionsOnly : ITypeMatcher { public bool Matches(Type type) => typeof(Exception).IsAssignableFrom(type); }`. If your type matcher needs to be a delegate or enum, decorate it with a `[TypeMatcher]` custom attribute. (@stakx, #908)
+
+* New overloads for `setup.Callback` and `setup.Returns` where the callback receives the current `IInvocation`:  
+   - `setup.Callback(new InvocationAction(invocation => ...))`
+   - `setup.Returns(new InvocationFunc(invocation => ...))`
+   These are indended to be used in scenarios where type matchers are used as a `IInvocation` allows discovery of the used type arguments (besides the regular arguments). (@stakx, #908)
```
stakx: * missing line break between this line and the preceding one
* typo: `indended` &rarr; `intended`

tests/Moq.Tests/ItIsAnyTypeFixture.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD.
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
+using System;
 using Xunit;
```
stakx: Add blank line between `System` and `Xunit` namespace imports.

src/Moq/ExpressionExtensions.cs
```diff
@@ -201,6 +201,18 @@ void Split(Expression e, out Expression r /* remainder */, out InvocationShape p
 					case ExpressionType.Call:  // regular method call
 					{
 						var methodCallExpression = (MethodCallExpression)e;
+
+						if (methodCallExpression.Method.IsGenericMethod)
+						{
+							foreach (var typeArgument in methodCallExpression.Method.GetGenericArguments())
+							{
+								if (typeArgument.IsTypeMatcher())
```
stakx: Since this loop is going to run for every generic method Moq encounters, it might be a good idea to cache information about whether or not some `Type` is a type matcher.

src/Moq/Guard.cs
```diff
@@ -16,6 +16,18 @@ namespace Moq
 	[DebuggerStepThrough]
 	internal static class Guard
 	{
+		public static void HasDefaultConstructor(Type type)
```
stakx: This might be useful in `MockDefaultValueProvider`, too. Today, `DefaultValue.Mock` will simply attempt to mock types even if they don't have a default constructor, resulting in a DynamicProxy exception (which could be hidden).

An alternative would be to instead add a `type.HasDefaultConstructor()` check inside `type.IsMockable()`. That way, `MockDefaultValueProvider` would no longer attempt to mock default ctor-less types and delegate to `DefaultValue.Empty`.

tests/Moq.Tests/ItIsAnyTypeFixture.cs
```diff
@@ -94,6 +94,37 @@ public void Setup_with_It_IsAnyType_default_return_value()
...
+				typeArgument = invocation.Method.GetGenericArguments()[0];
+			}));
+
+			_ = mock.Object.Method<string>();
+
+			Assert.Equal(typeof(string), typeArgument);
+		}
+
+		[Fact]
+		public void In_order_to_discover_type_arguments_Returns_accepts_a_single_IInvocation_parameter()
```
stakx: Should rename: `a_single_IInvocation_parameter` &rarr; `an_InvocationFunc_parameter`.

tests/Moq.Tests/ItIsAnyTypeFixture.cs
```diff
@@ -94,6 +94,37 @@ public void Setup_with_It_IsAnyType_default_return_value()
 			Assert.Empty((int[])result);
 		}
 
+		[Fact]
+		public void In_order_to_discover_type_arguments_Callback_accepts_a_single_IInvocation_parameter()
```
stakx: Should rename: `a_single_IInvocation_parameter` &rarr; `an_InvocationFunc_parameter`.

michal-ciechan:
Nice! Thanks for caring about this issue so long after! Will take a look at the code and take it out for a spin tomorrow!

stakx:
It just occurred to me that we could special-case `It.IsAny<It.IsAnyType>()` to be roughly equivalent to @helloserve's `It.Ignore()`. Some thoughts regarding this idea:

* Since `It.IsAny` gets used a lot, perhaps we should require that all type matchers carry a `[TypeMatcher]` attribute so that they can be identified more quickly.

* Consider special-casing other built-on matchers, too, where that makes sense.

* What to do about custom argument matchers? Those would have to deal with `It.IsAnyType` et al. themselves.

* Ideally, argument matchers wouldn't support just `It.IsAnyType`, but *any* type matcher.

Issue #925, opened 2019-09-07
Ability to specify proxy generator or invalidate it


bclothier:
[This was originally opened here](https://github.com/castleproject/Core/issues/472).

Basically, the issue is that Moq uses a single `ProxyGenerator`, which is fine for 99% of the scenarios. However, I have a scenario where I'm dealing with types that get created at runtime and consequently, can cause collisions with the proxy type cache because the type can change. 

The ideal solution would be to be able to use a `MockRepository` and provide a proxy generator or at least specify that it should use its own proxy generator, separate from the rest of proxy generator, so that any mocks based on those volatile types won't get cached for very long and can be easily invalidated.

[![Back this issue](https://raw.githubusercontent.com/devlooped/SponsorLink/main/docs/assets/img/separator.png "Back this issue")](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)
[![Back this issue](https://img.shields.io/badge/backed-%240-EA4AAA?logo=githubsponsors "Back this issue")](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)

stakx:
We'll have to be very careful not to create a leaky abstraction.

* Allowing users to state that they want a repository to use a separate proxy generator only makes sense if Moq has an explanation for why one would ever want to do that. Right now, the explanation would be, "because that will cause the repository to have its own proxy type cache". That explanation would very much be tied to DynamicProxy, because...

* Any notion of "proxy type cache" isn't valid in the general case as there could be proxy factory implementations that don't cache the generated types.

* Any notion of "invalidate the proxy generator" also doesn't make much sense to me in the general case... what exactly does that mean?

The above points lead me to believe that we need to let client code choose a proxy factory. That way, we can expose the DynamicProxy implementation and document its exact semantics, and keep them separate from the more general proxy factory concept.

I also mentioned in the Gitter chat that Moq v4 isn't quite ready to fully expose its proxy factory abstraction. That is mainly because it, as well as other parts of Moq's design, have been heavily influenced by the capabilities of DynamicProxy, e.g. the fact that you cannot mock static or sealed types, and that you cannot set up non-virtual methods. If we open up Moq to other proxy factory implementations (that may be more powerful, e.g. those based on the CLR's profiler API), we'd have to rewrite large parts of Moq so it is able to deal with statics, non-virtuals, etc. For now, that is out of the question due to the very large amount of work involved.

All of the above leads me to the following proposal:

[...]

PR #938, opened 2019-10-15
Create separate generator for each type

ishatalkin:
There is a known problem of .net that multiple types in one dynamic assembly is way slower than multiple dynamic assemblies with one type each (see [stackoverflow](https://stackoverflow.com/questions/47295780/multiple-types-in-one-dynamic-assembly-is-way-slower-than-multiple-dynamic-assem/47297416)). Moq uses Castle Dynamic Proxy to create proxy-types. Unfortunately, it takes more and more time to create each next proxy-type. There is a simple solution - create a dynamic assembly for each proxy type.

The simple test shows that Moq needs 32.9 sec to create 735 mocks of different types. If using one assembly for each proxy type it's necessayr 5.7 sec to create the same 735 mocks, so the speed increases in 5-6 times.

Test:
`        [Test]
        public void MultipleObjectCreationTest()
        {
            var sw = new Stopwatch();
            sw.Start();
            int i = 1;
            foreach (var type in AppDomain.CurrentDomain.GetAssemblies()
                .SelectMany(a => a.GetTypes())
                .Where(t => !t.ContainsGenericParameters)
                .Where(t => !t.IsSealed)
                .Where(t => t.IsPublic)
                .Where(t => t.GetConstructor(Type.EmptyTypes) != null))
            {
                try
                {
                    var sw2 = new Stopwatch();
                    sw2.Start();
                    var mockType = typeof(Mock<>).MakeGenericType(type);
                    var mock = Activator.CreateInstance(mockType);
                    var obj = mockType.GetProperties().Where(p => p.Name == "Object").First().GetValue(mock);
                    sw2.Stop();
                    Console.WriteLine($"{i}. Type = {type.Name}. Time = {sw2.ElapsedMilliseconds} ms");
                    i++;
                    if (sw.Elapsed.Minutes > 2)
                        break;
                }
                catch (Exception e)
                {
                    Console.WriteLine($"{i}. Type = {type.Name}. Exception = {e}");
                }
            }
            Console.WriteLine($"Time = {sw.Elapsed.TotalSeconds} s, Mocks created = {i - 1}");
        }`

### DIFF
--- a/src/Moq/ProxyFactories/CastleProxyFactory.cs
+++ b/src/Moq/ProxyFactories/CastleProxyFactory.cs
@@ -2,6 +2,7 @@
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
@@ -30,7 +31,7 @@ namespace Moq
 	internal sealed class CastleProxyFactory : ProxyFactory
 	{
 		private ProxyGenerationOptions generationOptions;
-		private ProxyGenerator generator;
+		private readonly ConcurrentDictionary<string, ProxyGenerator> generators = new ConcurrentDictionary<string, ProxyGenerator>();
 
 		#if FEATURE_CAS || FEATURE_COM
 		static CastleProxyFactory()
@@ -52,12 +53,13 @@ namespace Moq
 		public CastleProxyFactory()
 		{
 			this.generationOptions = new ProxyGenerationOptions { Hook = new IncludeObjectMethodsHook(), BaseTypeForInterfaceProxy = typeof(InterfaceProxy) };
-			this.generator = new ProxyGenerator();
 		}
 
 		/// <inheritdoc />
 		public override object CreateProxy(Type mockType, Moq.IInterceptor interceptor, Type[] interfaces, object[] arguments)
 		{
+			var generator = generators.GetOrAdd(mockType.Name, key => new ProxyGenerator());
+
 			// All generated proxies need to implement `IProxy`:
 			var additionalInterfaces = new Type[1 + interfaces.Length];
 			additionalInterfaces[0] = typeof(IProxy);


ishatalkin:
Time necessary to create one proxy (all proxy types in one assembly - this is the current Moq behaviour)
X - the number of proxy
Y - time to create the proxy (ms)

EXCEL_Bg6tMkljqv

ishatalkin:
Time necessary to create one proxy (one assembly for one proxy type approach)
X - the number of proxy
Y - time to create the proxy (ms)

EXCEL_MMANIGwyif

stakx:
Hi @ishatalkin, and thank you for contributing! This is intriguing, however I am not going to merge this PR. While I agree that proxy generator creation might currently be sub-optimal in certain scenarios, I don't think newing up new `ProxyGenerator`s by default is the way to go. Reusing a single generator sgould be perfectly fine for most regular use cases.

What's lacking in all of these measurements is statements about memory consumption. Dynamic assemblies don't come for free, and currently DynamicProxy lacks the ability to create collectible ones (see https://github.com/castleproject/Core/issues/473), so I would have to assume that once created, they're going to remain in memory (unless you unload a whole app domain, which may not even be possible on .NET Core <3). If your unit test suite ran for 19 seconds and that time was mostly because of proxy creation, I'm estimating that you must be creating tens of thousands of proxies there. Switching to one dynamic assembly per proxy type would likely yield a similarly staggering amount of assemblies loaded into the app domain... that can't be good with regard to memory usage. (But I admit that measurements would be required to back that up.) Nonetheless, note that DynamicProxy explicitly recommends using a single generator (see https://github.com/castleproject/Core/blob/master/docs/dynamicproxy.md).

The more important reason for rejecting this PR, however, is my belief that we can do better by giving user code more explicit control over proxy generator instantiation; see #925. The solution proposed there should work for your scenario, too, while avoiding the cons of increased memory consumption & cache misses in DP.

stakx:
> If #925 will give the ability to create a proxy generator for each type, it's OK.

Yes, but not automatically. Given that the proposal gets merged (I have a draft implementation lying around & waiting btw.) it would be your task to decide whether a mock should use a non-default factory, and if so, which one. Selecting a (or newing up a fresh) proxy factory per mocked type would only be one out of many possible policies. 

The only built-in policy would be that new mocks use a factory that's been set as the default (either globally, or at the `MockRepository` level), unless told otherwise.

Btw. two `Mock<T>()` with the same `T` don't necessarily end up using the same proxy type. You'd probably want to consider additional interface types (specified using `mock.As<TInterface>()`) along with the main mocked type.

Issue #930, opened 2019-09-20
Problem when mocking an interface with hidden property of different type


angelaki:
I get an "AmbiguousMatchException" when mocking this interface:

```
public interface A { string Prop { get; } }
public interface B : A { new int Prop { get; } }

[Fact]
public void TestMock()
{
	var mock = new Mock<B>();
	var p = mock.Object.GetType().GetProperty("Prop");
}
```

If the "Prop" types match it works. But what to do in this case? In my real scenario the implementation of the interface explicitly describes the value of the two properties. Their names need to match.

stakx:
This is not directly related to Moq, it's a basic question about .NET reflection. If you have two properties with the same name, you need to specify its parameter & return types or use `GetProperties()` instead.

angelaki:
Yes, I know. But for example EF uses reflection to set values of an entity based on a poco:

Entry(paramEntity).CurrentValues.SetValues(mock.Object);

Would throw this exception. How to deal with it? I know it is not absolutely moq-related. But doesn't moq need to offer a solution for this?

stakx:
No, Moq is not a library for doing reflection. 

Your issue is with a .NET reflection method call, so you need to learn how to use *that* API before anything else.

angelaki:
No, the problem is in your Proxy instance using the DynamicProxy. You should be able to override a method that choses the property that need to be taken.

E.g. in this case I could just say "GetProperties().First(p => p.Name == name)" or I'd know what property of what interface interests me. This fix can only be done in and by Moq I guess?

stakx:
> the problem is in your Proxy instance using the DynamicProxy.

(In that case, you're at the wrong repository. If you feel DynamicProxy is misbehaving, you should make a report over at castleproject/Castle.Core.)

DynamicProxy will create a proxy type similar to the following:

```csharp
partial class BProxy : B
{
    public string Prop { get { ... } }
    public int Prop { get { ... } }
}
```

which isn't valid C#, because C# does not allow you to overload members that differ only by their return types. But that practice is perfectly legal and possible at the IL / runtime level, and DynamicProxy happens to take advantage of that.

DynamicProxy makes no promises about how the generated proxy types will be implemented except that they're valid CLR types and that they can be used as proxies for the proxied type(s) (which in practice means that the proxy types are either subclasses of the proxied class type, and/or they implement the requested interfaces).

[...]

PR #939, opened 2019-10-16
AmbiguousMatchException when setting up the property, that hides another one

ishatalkin:
mock.Setup(m => m.Prop) throws AmbiguousMatchException in case that:
* Prop hides the property with *new* keyword;
* the hidden and the derived properties have different types;
* mocked object has the child class of the class, where the property was hidden.

You can repropduce the exception with the following code:

	public class HidePropertyFixture
	{
		public class A
		{
			public string Prop { get; }
		}

		public class B : A
		{
			// consider new keyword and the different type of property
			public new virtual int Prop { get; }
		}

		public class C : B
		{

		}

		[Fact]
		public void SetupsDerivedProperty()
		{
			// create the mock of C; mock of B works correct
			var mock = new Mock<C>();
			// the next line will throw AmbiguousMatchException 
			mock.Setup(m => m.Prop);
		}
	}

Stacktrace:

    System.Reflection.AmbiguousMatchException
       в System.RuntimeType.GetPropertyImpl(String name, BindingFlags bindingAttr, Binder binder, Type returnType, Type[] types, ParameterModifier[] modifiers)
       в System.Type.GetProperty(String name, BindingFlags bindingAttr)
       в Moq.ExpressionExtensions.GetReboundProperty(MemberExpression expression) в C:\Projects\github_moq\src\Moq\ExpressionExtensions.cs:строка 323
       в Moq.ExpressionExtensions.<Split>g__Split|4_1(Expression e, Expression& r, InvocationShape& p) в C:\Projects\github_moq\src\Moq\ExpressionExtensions.cs:строка 286
       в Moq.ExpressionExtensions.Split(LambdaExpression expression) в C:\Projects\github_moq\src\Moq\ExpressionExtensions.cs:строка 107
       в Moq.Mock.SetupRecursive[TSetup](Mock mock, LambdaExpression expression, Func`3 setupLast) в C:\Projects\github_moq\src\Moq\Mock.cs:строка 590
       в Moq.Mock.Setup(Mock mock, LambdaExpression expression, Condition condition) в C:\Projects\github_moq\src\Moq\Mock.cs:строка 527
       в Moq.Mock`1.Setup[TResult](Expression`1 expression) в C:\Projects\github_moq\src\Moq\Mock.Generic.cs:строка 407

Here is pull request that fixes the bug.

This bug is somewhat similar with #930, but here is the problem of Moq.

### DIFF
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -321,8 +321,9 @@ namespace Moq
 			if (property.DeclaringType != expression.Expression.Type)
 			{
 				var derivedProperty = expression.Expression.Type
-					.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
-					.SingleOrDefault(p => p.Name == property.Name && p.PropertyType == property.PropertyType);
+					.GetMember(property.Name, MemberTypes.Property, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+					.Cast<PropertyInfo>()
+					.SingleOrDefault(p => p.PropertyType == property.PropertyType);
 				if (derivedProperty != null && derivedProperty.GetMethod.GetBaseDefinition() == property.GetMethod)
 				{
 					return derivedProperty;
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -320,7 +320,9 @@ namespace Moq
 			// we "upgrade" to the derived property.
 			if (property.DeclaringType != expression.Expression.Type)
 			{
-				var derivedProperty = expression.Expression.Type.GetProperty(property.Name, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+				var derivedProperty = expression.Expression.Type
+					.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+					.SingleOrDefault(p => p.Name == property.Name && p.PropertyType == property.PropertyType);
 				if (derivedProperty != null && derivedProperty.GetMethod.GetBaseDefinition() == property.GetMethod)
 				{
 					return derivedProperty;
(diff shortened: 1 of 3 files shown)


src/Moq/ExpressionExtensions.cs
```diff
@@ -320,7 +320,9 @@ internal static PropertyInfo GetReboundProperty(this MemberExpression expression
 			// we "upgrade" to the derived property.
 			if (property.DeclaringType != expression.Expression.Type)
 			{
-				var derivedProperty = expression.Expression.Type.GetProperty(property.Name, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+				var derivedProperty = expression.Expression.Type
+					.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+					.SingleOrDefault(p => p.Name == property.Name && p.PropertyType == property.PropertyType);
```
stakx: Could you please rewrite this query as follows?:

```csharp
var derivedProperty = expression.Expression.Type
        .GetMember(property.Name, MemberTypes.Property, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
        .Cast<PropertyInfo>()
        .SingleOrDefault(p => p.PropertyType == property.PropertyType);
```

This should be more efficient as the filtering by name happens earlier (inside System.Reflection).

ishatalkin:
Ok! I'll do the stuff tomorrow.

ср, 16 окт. 2019 г., 20:23 stakx <notifications@github.com>:

ishatalkin:
@stakx, I've made the changes. Is it OK now?

ishatalkin:
> Btw. it appears that you are using a different email address in your commits than on GitHub, so your commits won't be linked to your profile when viewed on GitHub. If you want to fix that, now would be a good time.

Added secondary e-mail address to my profile. Feel free to merge the PR

stakx:
:rocket:

Thank you for contributing, @ishatalkin!

mujdatdinc:
What about `mock.Setup(m => m.Prop).Returns("string value");` ? Is there a way to setup base class property (string version) ?

PR #947, opened 2019-10-18
Naive first attempt to fix #946.

### LINKED ISSUE #946: Recursive mocks with different paths
Just a general question: Should this scenario be working at all, or is it not how recursive mocks are meant to be?

    var mock = Mock.Of<IRecursiveRoot>(MockBehavior.Loose);

    mock.Setup(m => m.Branch.GetLeaf(1).Name).Returns("foo");
    mock.Setup(m => m.Branch.GetLeaf(2).Name).Returns("bar");

    Assert.Equal("foo", mock.Branch.GetLeaf(1).Name);
    Assert.Equal("bar", mock.Branch.GetLeaf(2).Name);

Currently, it seems the second setup of m.Branch completely erases all previous setups, even if previous setups might be reused because actually, there's different paths being setup.

danielcweber:
Compares arrays to match a previous recursive mock setup where params have been used.

Fixes #946.

### DIFF
--- a/src/Moq/ExpressionComparer.cs
+++ b/src/Moq/ExpressionComparer.cs
@@ -143,6 +143,17 @@ namespace Moq
 
 		private static bool EqualsConstant(ConstantExpression x, ConstantExpression y)
 		{
+			if (x.Value is Array array1 && y.Value is Array array2 && array1.Length == array2.Length)
+			{
+				for (var index = 0; index < array1.Length; index++)
+				{
+					if (!object.Equals(array1.GetValue(index), array2.GetValue(index)))
+						return false;
+				}
+
+				return true;
+			}
+
 			return object.Equals(x.Value, y.Value);
 		}
 
--- a/tests/Moq.Tests/RecursiveMocksFixture.cs
+++ b/tests/Moq.Tests/RecursiveMocksFixture.cs
@@ -315,6 +315,19 @@ namespace Moq.Tests
 			Assert.Equal(5, fooMock.Object.Bar.GetBaz("foo").Value);
 		}
 
+		[Fact]
+		public void ParamsAreMatchedInSecondSetup()
+		{
+			string expected = "1";
+
+			var target = new Mock<IFoo>();
+			target.Setup(x => x.Params(1).Do("1")).Returns("1");
+			target.Setup(x => x.Params(1).Do("2")).Returns("2");
+
+			string ret = target.Object.Params(1).Do("1");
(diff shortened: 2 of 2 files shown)


stakx:
Seems we've been working on a bug fix at the same time! 💯

At a first glance, I thought that your solution has one big disadvantage: it changes LINQ expression tree comparisons such that all array values will be compared by their contents, whether they are `params` arrays or not.

At a second glance, and while working on my own solution (#948), I noticed that [other parts of Moq (`ConstantMatcher`) do something similar](https://github.com/moq/moq4/blob/08b1e685b0b110391f4b319a904962104071b546/src/Moq/Matchers/ConstantMatcher.cs#L27-L34).

Intuitively, my PR feels more correct to me, as it takes care to target *only* `params` arrays, and not all arrays&mdash;this by applying a change inside `InvocationShape`, where we have a little bit more context of the compared values than is available in `ExpressionComparer`. But it's somewhat at odds with other parts in Moq. This becomes visible e.g. in the following example:

```csharp
public interface IX
{
    int Method(params int[] args);
}

var one = new[] { 1 };
var alsoOne = new[] { 1 };

var mock = new Mock<IX>();
mock.Setup(m => m.Method(one)).Returns(1);
mock.Setup(m => m.Method(alsoOne)).Returns(11);

Assert.Equal(1, mock.Object.Method(one));
Assert.Equal(11, mock.Object.Method(alsoOne));
```

In current versions of Moq, the first assertion fails due to the structural equality comparison of all enumerables in `ConstantMatcher`.

If we decide that this is wrong, we should go with my solution. If we decide that this is correct behavior, we should perhaps go with yours.

danielcweber:
That's why I called it a "naive first attempt". I don't have too much insight into Moq, just wanted to offer something that might maybe have a chance to get into the next release which, as I have seen, is due Oct. 31st. You know the codebase, so I think we should definitely go with your solution.

danielcweber:
Is there a CI-feed?

stakx:
> Is there a CI-feed?

You mean a CI feed for NuGet? That would be https://ci.appveyor.com/nuget/moq4, I think.

stakx:
> I think we should definitely go with your solution

OK, will do. I'll close your PR then&mdash;thank you for contributing, nonetheless!

P.S. I ran a quick test regarding `ConstantMatcher`'s behavior mentioned above, I think it should be possible to improve consistency between its behavior and the one we're about to introduce via my PR. But I'll leave that for another time and PR.

PR #948, opened 2019-10-18
Match `params` arrays in setup/verification expressions using structural equality

### LINKED ISSUE #946: Recursive mocks with different paths

stakx:
... which is how Moq treats these arrays in other parts, but currently not in recursive setup/verification expressions.

Fixes #946.

### DIFF
--- a/src/Moq/InvocationShape.cs
+++ b/src/Moq/InvocationShape.cs
@@ -170,6 +170,11 @@ namespace Moq
 				// not array reference equality:
 				if (i == li && lastParameterIsParamArray)
 				{
+					// In the following, if we retrieved the `params` arrays via `partiallyEvaluatedArguments`,
+					// we might see them either as `NewArrayExpression`s or reduced to `ConstantExpression`s.
+					// By retrieving them via `Arguments` we always see them as non-reduced `NewArrayExpression`s,
+					// so we don't have to distinguish between two cases. (However, the expressions inside those
+					// have already been partially evaluated by `MatcherFactory` earlier on!)
 					if (this.Arguments[li] is NewArrayExpression e1 && other.Arguments[li] is NewArrayExpression e2 && e1.Expressions.Count == e2.Expressions.Count)
 					{
 						for (int j = 0, nj = e1.Expressions.Count; j < nj; ++j)
--- a/src/Moq/InvocationShape.cs
+++ b/src/Moq/InvocationShape.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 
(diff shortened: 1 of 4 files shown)


src/Moq/InvocationShape.cs
```diff
@@ -160,8 +161,29 @@ public bool Equals(InvocationShape other)
 				other.partiallyEvaluatedArguments = PartiallyEvaluateArguments(other.Arguments);
 			}
 
-			for (int i = 0, n = this.partiallyEvaluatedArguments.Length; i < n; ++i)
+			var lastParameter = this.Method.GetParameters().LastOrDefault();
+			var lastParameterIsParamArray = lastParameter != null && lastParameter.ParameterType.IsArray && lastParameter.IsDefined(typeof(ParamArrayAttribute));
+
+			for (int i = 0, li = this.partiallyEvaluatedArguments.Length - 1; i <= li; ++i)
```
stakx: Changing the upper bound and `<` to `<=` so we don't have to `- 1` inside the loop.

src/Moq/InvocationShape.cs
```diff
@@ -160,8 +161,29 @@ public bool Equals(InvocationShape other)
...
 			{
+				// Special case for final `params` parameters, which need to be compared by structural equality,
+				// not array reference equality:
+				if (i == li && lastParameterIsParamArray)
+				{
+					if (this.Arguments[li] is NewArrayExpression e1 && other.Arguments[li] is NewArrayExpression e2 && e1.Expressions.Count == e2.Expressions.Count)
```
stakx: I suspect we should really be comparing partially-evaluated arguments. The reason why this code accesses the original `Arguments` instead is to avoid having to deal with two distinct cases:

* `partiallyEvaluatedArguments[li]` may have been evaluated to a `ConstantExpression` wrapping an actual array&mdash;we'd no longer be able to use `ExpressionComparer` for its elements, but we'd have to do something akin to what `ConstantMatcher` does (i.e. comparing arrays using `SequenceEqual`).

* `partiallyEvaluatedArguments[li]` could have been left as a `NewArrayExpression`, meaning we'd use `ExpressionComparer` to compare its contained sub-`Expressions`.

❗️ Ideally, add more tests to see if the current approach is insufficient.

src/Moq/InvocationShape.cs
```diff
@@ -160,8 +161,29 @@ public bool Equals(InvocationShape other)
...
+					{
+						for (int j = 0, nj = e1.Expressions.Count; j < nj; ++j)
+						{
+							if (!ExpressionComparer.Default.Equals(e1.Expressions[j], e2.Expressions[j]))
+							{
+								return false;
+							}
+						}
+
+						continue;
```
stakx: We could just as well `break` as we're in the last loop iteration... `continue` seems more correct, though.

tests/Moq.Tests/InvocationShapeFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+			var (expr1, method1, args1) = Deconstruct<B>(b => b.Method(1, 2, 3));
+			var (expr2, method2, args2) = Deconstruct<B>(b => b.Method(1, 2, 3));
+
+			Assert.NotSame(expr1.Body, expr2.Body);  // let's make sure the compiler didn't optimize the above
+
+			var expectation1 = new InvocationShape(expr1, method1, args1);
+			var expectation2 = new InvocationShape(expr2, method2, args2);
+
+			Assert.Equal(expectation1, expectation2);
+		}
```
stakx: ❗️ These two tests are almost identical. Perhaps extract what's common to both into a private method.

PR #949, opened 2019-10-19
Let `InnerMockSetup` (i.e. internal setups for cached return values) match generic args exactly instead of by assignment compatibility.

### LINKED ISSUE #932: Regression in 4.13.0
After updating from 4.12.0 one of our tests fails, consistently every time I run it.
I've got a really hard time reproducing it, though, because call sequence, type definitions, inheritance and maybe even thread switches may play a role.

**Before reading through all the complicated blabla below jump right to [this comment](https://github.com/moq/moq4/issues/932#issuecomment-535509249)**

I've got a class:

    internal class ViewModelCache
    {
        private readonly List<object> viewModels = new List<object>();

        public void Register(object viewModel)
        {
            this.viewModels.Add(viewModel);
        }

        public virtual IReadOnlyList<TViewModel> Retrieve<TViewModel>()
        {
            return this.viewModels.OfType<TViewModel>().ToList();
        }
    }

A test which sets up the mock like:

    ViewModelCache = new Mock<ViewModelCache> { DefaultValue = DefaultValue.Mock };

No further setups.
Tthe test performs quite a lot of steps (it's not a class-level unit test), and at one point it fails here:
image

Inspecting the moq invocations I can see the reason:
Moq has actually created a mock of `IReadOnlyList<IHaveCellComponet>` instead of `IReadOnlyList<StorageViewModel>`:

image

image

## Trying to Reproduce
Calling `.Retrieve<StorageViewModel>` immediately after the creation of the moq succeeds.
image

Also, **the exception that originally occurred now doesn't happen anymore**. All further calls to `.Retrieve<StorageViewModel>` **succeed**.

The test still fails, further down, when retrieving an IReadOnlyList of another type, `MachineBaseViewModel`:
image

"further down" is proven by there being more invocations than the first time:
image

When I add another `.Retrieve<MachineBaseViewModel>()` right after creation of the moq, we get even further.
This means, in the time between creation of the Moq and calling `Resolve` something is happening which get's Moq to behave erratic. But I don't have a clue to what that is. I can't find any other access to `Mock<ViewModelCache>`, only to `Mock<ViewModelCache>.Object`.
Within our solution this reproduces reliably, but I haven't found the key to creating a repro yet.

stakx:
Resolves #932.

As mentioned in one commit message, I am not convinced this is a 100% correct bug fix... it could even introduce new problems. It would benefit from some additional thinking about how Moq should generally handle co-/contravariance when matching setups for generic methods.

/cc @BrunoJuchli


stakx:
Goind to merge this despite my reservations that I mentioned above. In the worst case, the change being made here means some additional "cache" misses, i.e. Moq will at worst produce new return values instead of returning the previous one. Other parts of Moq shouldn't be affected by this change.

BrunoJuchli:
@stakx 
Thanks a lot!

PR #952, opened 2019-10-19
Fix `SetupProperty` for partially overridden properties

### LINKED ISSUE #951: `mock.SetupProperty` throws `NullReferenceException` when called for partially overridden property
Noticed while working on a bug fix for #886 that `SetupProperty` will also fail for partially overridden properties, albeit for different reasons.

### Steps to reproduce:

```csharp
public class WithAutoProperty
{
    public virtual object Property { get; set; }
}

public class OverridesOnlySetter : WithAutoProperty
{
    public override object Property { set => base.Property = value; }
}

[Fact]
public void SetupAllProperties_retains_value_of_derived_read_write_property_that_overrides_only_setter()
{
    var mock = new Mock<OverridesOnlySetter>();
    mock.SetupProperty(m => m.Property);
    mock.Object.Property = "value";
    Assert.Equal("value", mock.Object.Property);
}
```

### Expected outcome:

The test passes.

### Actual outcome:

> ```
> System.NullReferenceException : Object reference not set to an instance of an object.
>    at Moq.ExpressionExtensions.GetReboundProperty(MemberExpression expression) in src\Moq\ExpressionExtensions.cs:line 327
>    at Moq.ExpressionExtensions.ToPropertyInfo(LambdaExpression expression) in src\Moq\ExpressionExtensions.cs:line 343
>    at Moq.Mock`1.SetupProperty[TProperty](Expression`1 property, TProperty initialValue) in src\Moq\Mock.Generic.cs:line 659
>    at Moq.Mock`1.SetupProperty[TProperty](Expression`1 property) in src\Moq\Mock.Generic.cs:line 621
>    at ...
> ```

stakx:
Fixes #951.

### DIFF
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -326,8 +326,8 @@ namespace Moq
 					.SingleOrDefault(p => p.PropertyType == property.PropertyType);
 				if (derivedProperty != null)
 				{
-					if ((derivedProperty.CanRead && derivedProperty.GetGetMethod(true).GetBaseDefinition() == property.GetGetMethod(true)) ||
-						(derivedProperty.CanWrite && derivedProperty.GetSetMethod(true).GetBaseDefinition() == property.GetSetMethod(true)))
+					if ((derivedProperty.CanRead(out var getter) && getter.GetBaseDefinition() == property.GetGetMethod(true)) ||
+						(derivedProperty.CanWrite(out var setter) && setter.GetBaseDefinition() == property.GetSetMethod(true)))
 					return derivedProperty;
 				}
 			}
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -284,8 +284,8 @@ namespace Moq
 						r = memberAccessExpression.Expression;
 						var parameter = Expression.Parameter(r.Type, r is ParameterExpression ope ? ope.Name : ParameterName);
 						var property = memberAccessExpression.GetReboundProperty();
-						var method = property.CanRead ? property.GetGetMethod(true) : property.GetSetMethod(true);
-						//                    ^^^^^^^                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+						var method = property.CanRead(out var getter) ? getter : property.GetSetMethod(true);
+						//                    ^^^^^^^                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 						// We're in the switch case block for property read access, therefore we prefer the
 						// getter. When a read-write property is being assigned to, we end up here, too, and
 						// select the wrong accessor. However, that doesn't matter because it will be over-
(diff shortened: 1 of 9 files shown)


src/Moq/ExpressionExtensions.cs
```diff
@@ -284,8 +284,8 @@ void Split(Expression e, out Expression r /* remainder */, out InvocationShape p
 						r = memberAccessExpression.Expression;
 						var parameter = Expression.Parameter(r.Type, r is ParameterExpression ope ? ope.Name : ParameterName);
 						var property = memberAccessExpression.GetReboundProperty();
-						var method = property.CanRead ? property.GetGetMethod(true) : property.GetSetMethod(true);
-						//                    ^^^^^^^                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+						var method = property.CanRead(out var getter) ? getter : property.GetSetMethod(true);
```
stakx: Using `property.GetSetMethod(true)` directly should be safe here, since there must be a setter if the property isn't readable. (Properties must have at least one accessor.)

src/Moq/ExpressionExtensions.cs
```diff
@@ -324,8 +324,10 @@ internal static PropertyInfo GetReboundProperty(this MemberExpression expression
 					.GetMember(property.Name, MemberTypes.Property, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
 					.Cast<PropertyInfo>()
 					.SingleOrDefault(p => p.PropertyType == property.PropertyType);
-				if (derivedProperty != null && derivedProperty.GetMethod.GetBaseDefinition() == property.GetMethod)
+				if (derivedProperty != null)
 				{
+					if ((derivedProperty.CanRead && derivedProperty.GetGetMethod(true).GetBaseDefinition() == property.GetGetMethod(true)) ||
+						(derivedProperty.CanWrite && derivedProperty.GetSetMethod(true).GetBaseDefinition() == property.GetSetMethod(true)))
```
stakx: This is what causes the `NullReferenceException`. When upgrading from a base property to a more derived one, the more derived one may not have the same accessors as the base one did.

src/Moq/Extensions.cs
```diff
@@ -250,24 +250,12 @@ public static bool CanOverride(this MethodBase method)
...
 		public static bool CanOverrideGet(this PropertyInfo property)
 		{
-			if (property.CanRead)
-			{
-				var getter = property.GetGetMethod(true);
-				return getter != null && getter.CanOverride();
-			}
-
-			return false;
+			return property.CanRead(out var getter) && getter.CanOverride();
```
stakx: This should be safe even when a base class derives a `virtual` readable property and the derived class decides to `sealed` the getter&mdash;in such a case, `property.CanRead()` will retrieve the more derived `sealed` accessor.

src/Moq/Extensions.cs
```diff
@@ -250,24 +250,12 @@ public static bool CanOverride(this MethodBase method)
...
 		public static bool CanOverrideSet(this PropertyInfo property)
 		{
-			if (property.CanWrite)
-			{
-				var setter = property.GetSetMethod(true);
-				return setter != null && setter.CanOverride();
-			}
-
-			return false;
+			return property.CanWrite(out var setter) && setter.CanOverride();
```
stakx: As above.

src/Moq/Mock.cs
```diff
@@ -813,7 +813,7 @@ internal void AddInnerMockSetup(MethodInfo method, IReadOnlyList<Expression> arg
 				var property = expression.ToPropertyInfo();
 				Guard.CanRead(property);
 
-				Debug.Assert(method == property.GetGetMethod(true));
+				Debug.Assert(property.CanRead(out var getter) && method == getter);
```
stakx: It's hard to tell whether this change is correct in all situations&mdash;but it doesn't break any tests, so let's go with it for now.

PR #964, opened 2019-12-01
Added support for Mock.Of<T> with not default constructor

### LINKED ISSUE #963: Support for Mock.Of<T> with not default constructor
At the moment, new Mock<T> and Mock.Of<T> are similar but if you need use non generic constructor you must use new Mock<T>. If you are using the  Mock.Of<T> syntax in the project, change all for this will be a problem.
Add support for this is a very little change and give more functionality to the Mock.Of<T> syntax.

JorTurFer:
This closes #963


tests/Moq.Tests/MockOf.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Xunit;
```
peterdew, line 1: I feel that instead of adding a new file for this tests you should extend the tests in Moq.Tests\linq\QueryableMocksFixture and extend the tests named 'shouldOneOf....CreateNewMock'. 
A test 'ShouldOneOfWithContructorParameterAndPredicateCreateNewMock ' would be a good additional test that is missing now.

src/Moq/Linq/Mock.cs
```diff
@@ -45,6 +56,29 @@ public partial class Mock
...
+		/// <summary>
+		/// Creates a mock object of the indicated type.
+		/// </summary>
+		/// <param name="behavior">Behavior of the mock.</param>
+		/// <param name="args">Optional constructor arguments if the mocked type is a class.</param>
+		/// <typeparam name="T">The type of the mocked object.</typeparam>
+		/// <returns>The mocked object created.</returns>
+		public static T Of<T>(MockBehavior behavior,params object[] args) where T : class
+		{
+			// This method was originally implemented as follows:
```
peterdew, line 68: This is copied from the overload above but I doubt if this historical improvement-insight should have his place in code. I would leave this at least in this copy imho

tests/Moq.Tests/MockOf.cs
JorTurFer: You are right @peterdew, when I added the tests I didn't find where place them but I will change now.
Thanks

src/Moq/Linq/Mock.cs
stakx: Agreed, that comment is a bit out of place since the method it appears in didn't even exist at the time that the comment refers to.

It's a detail and not super important, but if you're going to edit anyway consider removing this comment.

tests/Moq.Tests/MockOf.cs
JorTurFer: I am reviewing the code and I didn't add support for default constructor with predicate so is not necesary those tests

src/Moq/Linq/Mock.cs
JorTurFer: You are right @peterdew and @stakx. When I copied the original method I forgot to remove the comment, sorry. I will remove it

PR #967, opened 2019-12-20
add support for mocking protected generic methods

JmlSaul:


src/Moq/Protected/IProtectedMock.cs
```diff
@@ -38,6 +39,16 @@ public interface IProtectedMock<TMock> : IFluentInterface
 		/// remember to use <see cref="ItExpr"/> rather than <see cref="It"/>.</param>
 		ISetup<TMock> Setup(string voidMethodName, params object[] args);
 
+		/// <summary>
+		/// Specifies a setup for a void method invocation with the given 
+		/// <paramref name="voidMethodName"/>, optionally specifying arguments for the method call.
+		/// </summary>
+		/// <param name="voidMethodName">The name of the void method to be invoked.</param>
+		/// <param name="genericTypeArguments">An array of types to be substituted for the type parameters of the current generic method definition, null for non-generic method.</param>
```
stakx: > `null for non-generic method`

I'd prefer `null` and an empty `Type[0]` array to be invalid argument values for this new parameter (`genericTypeArguments`). If you want to set up a non-generic method, simply use the existing overloads.

That is, please remove this comment and add a `Guard.NotNull(genericTypeArguments, nameof(genericTypeArguments))` to the new public overloads. (You may of course still use `null` in the internal implementation.)

src/Moq/Protected/ProtectedMock.cs
```diff
@@ -35,16 +35,24 @@ public ProtectedMock(Mock<T> mock)
 
 		public ISetup<T> Setup(string methodName, params object[] args)
 		{
-			Guard.NotNullOrEmpty(methodName, nameof(methodName));
+			return Setup(methodName, null, false, args);
```
stakx: Detail: Please change `Setup` &rarr; `this.Setup` here and in other places where you made a similar change, in order to stay consistent with (the majority of) the code base.

stakx:
Thanks for your contribution, @JmlSaul!

Issue #937, opened 2019-10-14
Provide a way to opt out of VerifyAll()


ashmind:
My goal is similar to #747, so I am using `VerifyAll()` to detect unused setups. However there are a few very common setups done in common helpers that I don't want to detect.

It would be great to have some way to opt-out of `VerifyAll` in specific cases, e.g. by using `Verifiable(false)`.

stakx:
@ashmind, suppose there were a new `setup.Verifiable(bool)` method overload where:

* `.Verifiable(true)` is equivalent to today's `.Verifiable()`, and
* `.Verifiable(false)` is a no-op, i.e. equivalent to omitting `.Verifiable()` altogether.

Would that be sufficient for you to be able to parameterize your helpers? You'd then use `.Verify()` instead of `.VerifyAll()` and therefore test only setups marked as `.Verifiable([true])`, but exclude those marked with `[.Verifiable(false)]` (I'm using `[]` to show parts that may be omitted).

ashmind:
Would this require me to opt-in with `Verifiable(true)` on most setups?
I'll elaborate on my situation a bit.

#### My Project Scenario 1
(most common, any developer in the project)
```
public void SomeTest() {
     GetMock<X1>().Setup(...);
     GetMock<X2>().Setup(...);

     // Act
     // Assert
}
```

No `Verifiable` anywhere (and developer might not even know it exists). Since "ambient" mocks are retrieved from test scope, base test class can call `VerifyAll()` on those mocks in after-test method, to ensure there are no unnecessary setups unused by the test.

Most of the time developers might not even be aware base class does that.

#### My Project Scenario 2
(rare, only a few most widely used shared helpers)

```
public void SomeSetupHelper() {
     GetMock<X1>().Setup(...).Verifiable(false);
     GetMock<X2>().Setup(...).Verifiable(false);
}
```

Can explicitly opt out of verification -- not perfect, but makes helper simpler and it's too widely used to guarantee both setups will be used (would be cool to guarantee at least one was used, but well).

There are only a few, and the authors of those will be more experienced developers, aware of Verifiable(false) if added.

stakx:
> Would this require me to opt-in with `Verifiable(true)` on most setups?
>  I'll elaborate on my situation a bit. [...]

Yes, and based on your two scenarios, I can see how this wouldn't help a lot. My immediate thoughts:

[...]

PR #976, opened 2020-03-13
WIP: Expose `Mock.Setups` collection

stakx:
This has been in the plans for a while now, and might go a long way towards resolving #937. Work in progress. What's still missing:

* [x] If a setup resulted from splitting up a composite setup expression, a means to get back at the original expression ("navigate upward"). The plan here is to have an `ISetup` implementation that represents the full, original setup expression; say `ICompositeSetup`. Any partial setup is associated with that original setup and can be queried for it. (When verified, such an original setup would verify only the exact call described by it, and not other invocations on any inner mocks along the way.)
* [x] If a setup resulted from splitting up a composite setup expression, a means to get at the inner mock ("navigate forward") so user code can do recursive verification.
* [x] Unit tests.
* [x] XML documentation comments.
* [ ] Check whether exposing verification methods on `Setup` means that we should also expose a static factory method on `MockException` to produce custom verification errors. (Likely only the case if we *also* expose a new `mock.Verify(setup => ...)` method.

Some additional minor flaws that we might need improvement (now or in a future PR):

* [ ] When looking at original setups created by `Mock.Of<T>()`, they won't include the operator (`==`) and the right-hand side (i.e. the configured return value). They will only consist of the left-hand side (i.e. the member being set up).

/cc @ashmind


src/Moq/ISetup.cs
```diff
@@ -14,5 +14,15 @@ public interface ISetup
...
+
+		/// <summary>
+		///   Gets whether this setup is conditional.
+		/// </summary>
+		bool IsConditional { get; }
+
+		/// <summary>
+		///   Gets whether this setup is disabled.
+		/// </summary>
+		bool IsDisabled { get; }
```
stakx: The reason for renaming this to `IsOverridden` in a later commit is that the new name better maintains the abstraction: Noone should have to know (outside of the Moq code base) that overridden setups are implemented using some kind of "disabled" flag.

(Also, it will be much more likely that users will request a `void Disable()` method once they see a `IsDisabled` property, which would be handy (essentially giving users the means to "remove" single setups) but hard to implement properly at this time. Imagine disabling a setup that overrides an earlier setup with the same expression; the latter would then have to be automatically re-enabled. Given today's implementation, that would be fairly tricky to achieve efficiently. Let's solve that problem at a later time.)

src/Moq/ISetup.cs
```diff
@@ -24,5 +24,25 @@ public interface ISetup
...
 		bool IsDisabled { get; }
+
+		/// <summary>
+		///   Gets whether this setup is verifiable.
+		/// </summary>
+		bool IsVerifiable { get; }
+
+		/// <summary>
+		///   Verifies this setup.
+		///   That is, checks whether it has been matched by at least one invocation on the mock.
```
stakx: For future reference, this is actually slightly inaccurate: an `AutoImplementedProperty[GS]etterSetup` can pass verification even if it has never been invoked.

src/Moq/Setup.cs
```diff
@@ -31,10 +34,12 @@ protected Setup(InvocationShape expectation)
 
 		public bool IsDisabled => (this.flags & Flags.Disabled) != 0;
 
-		public virtual bool IsVerifiable => false;
+		public bool IsVerifiable => (this.flags & Flags.Verifiable) != 0;
 
 		public MethodInfo Method => this.expectation.Method;
 
+		public Mock Mock => this.mock;
```
stakx: Could be exposed via `ISetup`. The only reason why I opted not to do so right away is because we don't link back to the owning `Mock` in `IInvocation`, either (in order to keep that type and its implementation slim)... nor do we have back links to the `Mock` in the `mock.Invocations` and `mock.Setups` collections. It's nice to have some consistency across all of these types.

src/Moq/CompositeSetupBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+			var expression = this.Expression;
+			var mockedType = expression.Parameters[0].Type;
+
+			var builder = new StringBuilder();
+			builder.AppendNameOf(mockedType)
+			       .Append(' ')
+			       .Append(expression.PartialMatcherAwareEval().ToStringFixed());
+
+			return builder.ToString();
+		}
```
stakx: It would be nice to someday not have this code duplicated across `CompositeSetupBuilder` and `Setup`.

src/Moq/Mock.cs
```diff
@@ -621,12 +626,13 @@ private static TSetup SetupRecursive<TSetup>(Mock mock, LambdaExpression express
 								Resources.UnsupportedExpression,
 								expr.ToStringFixed() + " in " + expression.ToStringFixed() + ":\n" + Resources.TypeNotMockable));
 					}
-					setup = new InnerMockSetup(mock, expectation: part, returnValue);
+					setup = new InnerMockSetup(mock, expectation: part, returnValue, originalSetup);
+					originalSetup.AddPart(setup);
```
stakx: We might want to add a `Debug.Assert(originalSetup != null)` here.

stakx:
@ashmind, if you have any time during the next few days, I'd value some input on this. Is the public API behind `Mock.Setups` and `ISetup` clear enough? In particular, is it clear what `ISetup.IsMatched`, `ISetup.IsPartOfCompositeSetup`, and `ISetup.ReturnsInnerMock` do? Do you find these names confusing, and if so, do you have suggestions for alternate names?

What's still missing here (but let's perhaps add that in a subsequent PR) is the ability of setups to carry random user-provided state. It would perhaps be easiest to simply add a writable state bag to `ISetup` (essentially something like a `ConcurrentHashSet<object> State { get; }`). Given that you don't have any multi-threading nastiness going on during your setups, you could mark a setup immediately after creating it using `mock.Setups.Last().State.Add(<tag>)`, then later filter your setups using `mock.Setups.Where(setup => setup.State.Contains(<tag>)`. Meaning we wouldn't strictly need any addition to the fluent API, that would be more of a convenience.

stakx:
I'm going to split up this PR into a few, where the first one will expose only basic (but undisputed) stuff about setups. Subsequent PRs will look at exposing more advanced functionality on top of that.

ashmind:
Sorry, I didn't have time to take a look at it until now. Are there any open subsequent PRs or should I provide feedback for this one?

I'll do some here.

Btw, as a note
```
mock.Setups.Last().State.Add(<tag>)
```
I think this one can be fluent, `mock.Setup(...).Tag(...)`?

src/Moq/ISetup.cs
```diff
@@ -0,0 +1,89 @@
...
+		///   If this setup is part of a composite setup,
+		///   this <see langword="out"/> parameter will be set to the latter.
+		/// </param>
+		/// <returns>
+		///   <see langword="true"/> if this setup is a part of a composite setup;
+		///   otherwise, <see langword="false"/>.
+		/// </returns>
+		/// <seealso cref="ICompositeSetup"/>
+		/// <seealso cref="ReturnsInnerMock(out Mock)"/>
+		bool IsPartOfCompositeSetup(out ICompositeSetup originalSetup);
```
ashmind, line 58: I think this one is a bit complicated because it both gives you boolean and a value.
What makes `originalSetup` `original` -- is it more like `baseSetup`?

Would be keen the have this as two properties (or two methods) unless there are large performance implications.

src/Moq/ISetup.cs
```diff
@@ -0,0 +1,89 @@
...
+		/// </summary>
+		/// <param name="innerMock">
+		///   If this setup is known to return a mocked object,
+		///   this <see langword="out"/> parameter will be set to the corresponding <see cref="Mock"/> instance.
+		/// </param>
+		/// <returns>
+		///   <see langword="true"/> if this setup is known to return a mocked object;
+		///   otherwise, <see langword="false"/>.
+		/// </returns>
+		bool ReturnsInnerMock(out Mock innerMock);
```
ashmind, line 72: Same note on two properties -- naming makes sense, but why not `Mock? ReturnedInnerMock` or something similar?

src/Moq/ISetup.cs
```diff
@@ -0,0 +1,89 @@
...
+
+		/// <summary>
+		///   Gets whether this setup is conditional.
+		/// </summary>
+		bool IsConditional { get; }
+
+		/// <summary>
+		///   Gets whether this setup was matched by at least one invocation on the mock.
+		/// </summary>
+		bool IsMatched { get; }
```
ashmind, line 26: Just to clarify, if this is `false` and I want to convert it into a nice exception in a specific place, I can call `Verify()`?

ashmind:
This is a huge effort, even if not merged yet -- thanks for doing it!

src/Moq/ISetup.cs
stakx: Say you have a setup with an expression that involves method chaining; `parentMock.Setup(parent => parent.Child.Member)`. Moq will split this up and create, not one, but two setups:

* a setup on `parentMock` for `parent => parent.Child` which returns some `Mock<TChild>.Object`
* a setup on that child mock for `child => child.Member`.

These are the setups you will see in the `Setups` collections of these two mocks; the original expression is gone.

This method here gives you a means to get back at the full, original, composite setup expression.

I haven't pushed this yet, but I'm most likely going to turn this method into a simple `ISetup OriginalSetup { get; }` method, which may return the very same `ISetup` instance, or a `IFluentSetup` (the new name for `ICompositeSetup`&mdash;it's somewhat closer to Moq's traditional terminology).

src/Moq/ISetup.cs
stakx: A simple property `Mock InnerMock { get; }` (which can also return `null`) would be simplest, I'm undecided whether it's a good idea to make this a property since there's possibly a bit of computation going on.

src/Moq/ISetup.cs
stakx: `Verify` will *typically* throw if `IsMatched` is `false`&mdash;but there are some internal setup types where that isn't necessarily the case (among other reasons, for backward compatibility).

For instance, if a `get`/`set` property gets setup by `SetupAllProperties`, two setups are created: one for the getter, one for the setter. Both of these setups will verify successfully even if they were never matched by an invocation.

TL;DR: Even if `IsMatched` is `false`, `Verify` may succeed and give you no error message. If you want to verify, call `Verify`. :-)

I'll eventually try to document the exact mode of operation of `Verify` et al., but I'm not quite there yet.

stakx:
Thanks for your comments @ashmind. I've since merged 4 PRs (see the changelog), I've juggled around and renamed a few things since this PR, and I'm going to simplify a bit more&mdash;especially those `out`-parameter methods you commented on. [This](https://github.com/moq/moq4/blob/master/src/Moq/ISetup.cs) is the current `ISetup` interface. Changes to come:

* `bool BelongsToFluentSetup(out IFluentSetup)` is likely going to change into `ISetup OriginalSetup { get }`
* `bool? ReturnsMock(out Mock innerMock)` may change to simply `Mock InnerMock { get; }`

(The reason for these changes being that `ISetup` is mostly about introspection, and `out`-parameter methods don't play nice with e.g. a Watch window.)

src/Moq/ISetup.cs
ashmind: Thanks, I get it now! Maybe `FullSetup` would be clearer for me, but either way it's clear now.

src/Moq/ISetup.cs
ashmind: Sure -- `Mock? GetInnerMock()` also sounds fine/clear.

ashmind:
Yep I think it's also way simpler than out-parameter methods, thanks!

PR #978, opened 2020-03-22
Retire `FluentMockVisitor`

stakx:
`FluentMockVisitor` was the original component used to automatically set up inner mocks for multi-dot setup/verification expressions. For the regular `new Mock<T>`, `mock.Setup`, `mock.Verify` APIs, we began replacing this component with new infrastructure based on `Mock.SetupRecursive` and `expression.Split()` &mdash; see approx. #765. However, `FluentMockVisitor` has continued to be used for LINQ to Mocks.

This PR converts LINQ to Mocks to use the new infrastructure so we don't have to maintain two parallel mechanisms for dealing with multi-dot expressions.

(This will also simplify things over in #976.)


src/Moq/Linq/MockSetupsBuilder.cs
```diff
@@ -212,24 +191,89 @@ private static Expression ConvertToSetupProperty(Expression targetObject, Expres
...
+			{
+				// An expression may contain nested expressions, such as:
+				//
+				//     m => m.GetObjectById(It.IsAny<int>(id => id > 0))
+				//          ^                                   ^^
+				//   We're only interested            ...but not in this one!
+				//   in this parameter...
+				//
+				// Therefore, for the nested (quoted) expression, we short-circuit:
+				return node.NodeType == ExpressionType.Quote ? node : base.VisitUnary(node);
```
stakx: It *may* be safer to replace this guard with a different algorithm:

1. Implement a new visitor that travels along the "left spine" of an expression until it arrives at the root `ParameterExpression`, then it stores a reference to it.

2. Run that visitor on the `left` expression and make the identified `ParameterExpression` available to the present visitor.

2. Add checks in the present visitor so it only transforms expressions involving the same `ParameterExpression` that was identified in the earlier steps.

src/Moq/Linq/MockSetupsBuilder.cs
```diff
@@ -116,33 +116,10 @@ private static Expression ConvertToSetup(Expression left, Expression right)
 		{
 			switch (left.NodeType)
 			{
-				case ExpressionType.MemberAccess:
-					var member = (MemberExpression)left;
-					Guard.NotField(member);
```
stakx: This check are removed because `expression.Split` and `InvocationShape` should already perform their own checks.

PR #984, opened 2020-03-29
Expose `Mock.Setups`, part 1: Basics

stakx:
This exposes a mock's `Setups` collection in the public API. Setups accessible via that (for now read-only) collection are of type `ISetup` and (for now) have three properties that describe some basic aspects of each setup:

* `Expression`
* `IsConditional`
* `IsOverridden`

This PR will be followed by a couple more that add more advanced functionality, e.g. to navigate the mock-setup or setup-setup object graph, and to perform custom verification.

### DIFF
--- a/src/Moq/ISetupList.cs
+++ b/src/Moq/ISetupList.cs
@@ -9,7 +9,7 @@ namespace Moq
 	///   A list of setups that have been configured on a mock,
 	///   in chronological order (that is, oldest setup first, most recent setup last).
 	/// </summary>
-	public interface ISetupList : IEnumerable<ISetup>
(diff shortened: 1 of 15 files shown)


src/Moq/ISetupList.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Moq
+{
+	/// <summary>
+	///   A list of setups that have been configured on a mock,
+	///   in chronological order (that is, oldest setup first, most recent setup last).
+	/// </summary>
+	public interface ISetupList : IReadOnlyList<ISetup>
+	{
+	}
```
stakx: Should `void Clear()` be exposed? Now, or in a later PR?

PR #985, opened 2020-03-31
Expose `Mock.Setups`, part 2: Match status of setups & invocations

stakx:
This allows you...

* to distinguish between unmatched setups and those that were matched by at least one invocation via the `ISetup.WasMatched { get; }` boolean property; and

* to find out whether an invocation matched a setup (and which one) via the `IInvocation.WasMatched(out ISetup matchingSetup)` method.

### DIFF
--- a/src/Moq/Setup.cs
+++ b/src/Moq/Setup.cs
@@ -114,13 +114,13 @@ namespace Moq
 		}
 
 		/// <summary>
-		///   Verifies this setup and/or those of its inner mock (if present and known).
+		///   Verifies this setup and those of its inner mock (if present and known).
 		/// </summary>
 		/// <param name="predicate">
 		///   Specifies which setups should be verified.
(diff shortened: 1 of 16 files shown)


src/Moq/Setup.cs
```diff
@@ -35,7 +35,25 @@ protected Setup(InvocationShape expectation)
...
+		{
+			// update this setup:
+			this.flags |= Flags.Matched;
+
+			// update invocation:
+			invocation.MarkAsMatchedBy(this);
+			this.SetOutParameters(invocation);
+
+			// update condition (important for `MockSequence`) and matchers (important for `Capture`):
+			this.Condition?.SetupEvaluatedSuccessfully();
```
stakx: Interesting to note that `Condition` no longer needs to remain `public`; it is now only ever queried from within the same type.

PR #987, opened 2020-03-31
Expose `Mock.Setups`, part 3: Individual setup verification

stakx:
This adds `void Verify(bool recursive = true)` and `void VerifyAll()` methods to `ISetup`.

### DIFF
--- a/src/Moq/ISetup.cs
+++ b/src/Moq/ISetup.cs
@@ -46,5 +46,39 @@ namespace Moq
 		///   Gets whether this setup was matched by at least one invocation on the mock.
 		/// </summary>
 		bool WasMatched { get; }
+
+		/// <summary>
+		///   Verifies this setup and optionally all verifiable setups of its inner mock (if present and known).
+		///   <para>
+		///     If <paramref name="recursive"/> is set to <see langword="true"/>,
+		///     the semantics of this method are essentially the same as those of <see cref="Mock.Verify()"/>,
+		///     except that this setup (instead of a mock) is used as the starting point for verification,
+		///     and will always be verified itself (even if not flagged as verifiable).
+		///   </para>
+		/// </summary>
+		/// <param name="recursive">
+		///   Specifies whether recursive verification should be performed.
+		/// </param>
+		/// <exception cref="MockException">
+		///   Verification failed due to one or more unmatched setups.
+		/// </exception>
+		/// <seealso cref="VerifyAll()"/>
+		/// <seealso cref="Mock.Verify()"/>
+		void Verify(bool recursive = true);
+
+		/// <summary>
+		///   Verifies this setup and all setups of its inner mock (if present and known),
+		///   regardless of whether they have been flagged as verifiable.
+		///   <para>
+		///     The semantics of this method are essentially the same as those of <see cref="Mock.VerifyAll()"/>,
+		///     except that this setup (instead of a mock) is used as the starting point for verification.
+		///   </para>
+		/// </summary>
+		/// <exception cref="MockException">
+		///   Verification failed due to one or more unmatched setups.
+		/// </exception>
+		/// <seealso cref="Verify(bool)"/>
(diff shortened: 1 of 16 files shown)


src/Moq/ISetup.cs
```diff
@@ -29,9 +29,55 @@ public interface ISetup
...
 		///   Gets whether this setup was matched by at least one invocation on the mock.
 		/// </summary>
 		bool WasMatched { get; }
+
+		/// <summary>
+		///   Verifies this setup and optionally all verifiable setups of its inner mock (if present and known).
+		///   <para>
+		///     If <paramref name="recursive"/> is set to <see langword="true"/>,
+		///     the semantics of this method are essentially the same as those of <see cref="Mock.Verify()"/>,
+		///     except that this setup (instead of a mock) is used as the starting point for verification.
```
stakx, line 55: "[...] and will always be verified itself."

src/Moq/ISetup.cs
```diff
@@ -29,9 +29,55 @@ public interface ISetup
...
+		/// <exception cref="MockException">
+		///   Verification failed due to one or more unmatched setups.
+		/// </exception>
+		/// <seealso cref="VerifyAll()"/>
+		/// <seealso cref="Mock.Verify()"/>
+		void Verify(bool recursive = true);
+
+		/// <summary>
+		///   Verifies this setup and all setups of its inner mock (if present and known),
+		///   regardless of whether they have been flagges as verifiable.
```
stakx, line 70: `flagges` &rarr; `flagged`

src/Moq/Setup.cs
```diff
@@ -116,6 +126,9 @@ public override string ToString()
 		/// <summary>
 		///   Verifies this setup and those of its inner mock (if present and known).
 		/// </summary>
+		/// <param name="recursive">
+		///   Specifies whether verification
```
stakx, line 130: ... yes?

src/Moq/Setup.cs
```diff
@@ -169,11 +179,35 @@ protected virtual void ResetCore()
...
+		{
+			this.Verify(recursive: true, setup => !setup.IsOverridden && !setup.IsConditional);
+		}
+
+		private void Verify(bool recursive, Func<ISetup, bool> predicate)
+		{
+			foreach (Invocation invocation in this.mock.MutableInvocations)
+			{
+				invocation.MarkAsVerifiedIfMatchedBy(setup => setup == this);
+			}
```
stakx, line 197: Should this only happen if verification actually succeeds? Same question goes for the corresponding loop in `Mock.Verify`.

tests/Moq.Tests/SetupFixture.cs
```diff
@@ -77,6 +77,212 @@ public void IsOverridden_becomes_true_as_soon_as_another_setup_with_an_equal_exp
...
+			var mock = new Mock<IX>();
+			mock.Setup(m => m.Inner.Property);
+			var setup = mock.Setups.First();
+
+			var innerMock = mock.Object.Inner;
+			var innerMockSetup = Mock.Get(innerMock).Setups.First();
+
+			Assert.True(setup.WasMatched);
+			Assert.False(innerMockSetup.IsVerifiable);  // which means that the inner mock setup will be ignored
+			Assert.False(innerMockSetup.WasMatched);  // this Would make verification fail if that setup were not ignored
```
stakx, line 162: `Would` &rarr; `would`

src/Moq/Setup.cs
stakx: While the current logic isn't perfect, it's generally good enough: if verification fails, an exception gets thrown anyway; it's unlikely someone catches that exception (Moq documentation explicitly recommends not doing that!) and then goes on to inspect the mock's state.

There may possibly be some odd corner cases regarding the verification status of invocations on a setup's inner mock; let's deal with those when they become an actual problem.

PR #1008, opened 2020-04-25
Add `Await` operator for use in async setup expressions

### LINKED ISSUE #1007: Easier async setups through a new `Await(...)` operator

```csharp
public interface IX
{
    IX Child { get; }
    Task<IX> GetChildAsync();

    string Name { get; }
}

```

## Problem:

```csharp
          .Returns("Alice");
```

```csharp
```

```csharp
```

## Proposed solution:

```csharp
          .Returns("Alice");
```

```csharp
          .Returns("Alice");
```

```csharp
```

could be rewritten as:

```csharp
```

or `ThrowsAsync`:

```csharp
```

could be rewritten as:

```csharp
```

```csharp
```

stakx:
This is a proof of concept demonstrating the general feasibility of an `Await` operator for use in async setup expressions as described in #1007. This would still need to be...

* [ ] properly tested (including `Verify`!?)
* [ ] cleaned up

Closes #1007.

### DIFF
--- a/src/Moq/ActionObserver.cs
+++ b/src/Moq/ActionObserver.cs
@@ -8,9 +8,10 @@ using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-
+using System.Threading.Tasks;
 using Moq.Expressions.Visitors;
 using Moq.Internals;
+using Moq.Linq.Expressions;
 using Moq.Properties;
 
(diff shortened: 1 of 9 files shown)


stakx:
~~Closing because this approach won't work with delegate-based setup methods like `SetupSet`, `SetupAdd`, and `SetupRemove`.~~ (I forgot: it _does_ work with those, too!)

PR #1009, opened 2020-04-25
Add raise async event methods

shadowchamber:
Hello. For some my project it was required to raise event which has async event 

```
public interface IView
    {
        event AfterAppearingAsyncEventHandler AfterAppearingAsync;
        object BindingContext { get; set; }
    }

    public delegate Task AfterAppearingAsyncEventHandler(object sender);
```

So I had to raise this event in test but i need to wait the event for checking values. after:

```
mockPairListView.Raise(i => i.AfterAppearingAsync += null, this);
// here i should wait for event
Assert.AreEqual(Pairs.Count, presenter.Items.Count);
```

It is not possible to create an extension method cause RaiseEvent method is internal,
so I created a local copy of moq library and made changes for my project which I suggest in this pull request as a result now I'm able to use it like this

await mockPairListView.RaiseAsync(i => i.AfterAppearingAsync += null, this);

It would be great to have it in next nuget package

Thank you

### DIFF
--- a/src/Moq/Mock.Generic.cs
+++ b/src/Moq/Mock.Generic.cs
@@ -10,7 +10,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Text;
 using System.Threading;
-
+using System.Threading.Tasks;
 using Moq.Language;
 using Moq.Language.Flow;
 using Moq.Properties;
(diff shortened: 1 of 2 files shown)


stakx:
Sorry, but I'm not going to merge this one: Events with a non-`void` return type are generally not recommended (what return value do you get back when there are several subscribers, and what happens with the remaining return values?), and we shouldn't encourage that anti-pattern by adding support for it.

See e.g.:

* [Why do event handlers always have a return type of void?](https://stackoverflow.com/questions/3325396/why-do-event-handlers-always-have-a-return-type-of-void) (question on Stack Overflow)

* [Custom Event Handler Design](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/event#custom-event-handler-design) (.NET Framework design guidelines)

Hint, if you want to keep using events where subscribers return `Task`s, you should perhaps [fetch the event's invocation list](https://docs.microsoft.com/en-us/dotnet/api/system.delegate.getinvocationlist), call all subscribers manually and collect their return values, then do a `await Task.WhenAll` (or similar).

shadowchamber:
You might be right but it is a question of testing, testing is required even if design is bad.
However originally it came from IAsyncCommand [https://johnthiriet.com/mvvm-going-async-with-async-command/](https://johnthiriet.com/mvvm-going-async-with-async-command/)
And if we have such request on stackoverflow [https://stackoverflow.com/questions/47785263/moq-raise-for-an-async-event-listener](https://stackoverflow.com/questions/47785263/moq-raise-for-an-async-event-listener) then it is useful not only for me.
Of course it's up to you merge or not but at least it would be great to be able to extend your classes for my case.
Thanks for the hint normally it should be only one handler in my case but i think it will be good to be able to process even in cases with multiple handlers so I'll do like you suggested.

stakx:
@shadowchamber, I understand the argument you are making. Nonetheless, I think we're free to choose *not* to support a particular programming practice, especially when it's a generally recommended-against one that also isn't very widespread (which is my assumption in this case). It's not like you'll be stuck after all, you can always code up a mock object manually for such edge cases.

Thanks for your understanding. :-)

shadowchamber:
ok, thanks for the explanation.

Issue #972, opened 2020-02-13
Interface Default methods are ignored


hahn-kev:
I'm not sure if this would be considered a bug, but I'd think it would work the same as mocking an abstract class.

Example
```C#
interface ITest
{
    string CallMe();

    string CallMeDefault()
    {
        return CallMe();
    }
}

[Fact]
public void DefaultFails()
{
    var mock = new Mock<ITest>(MockBehavior.Loose);
    mock.Setup(t => t.CallMe()).Returns("hello");
    Assert.NotNull(mock.Object.CallMeDefault());
}
```

My interface has the method `CallMeDefault` which has an implementation, however when I mock that interface it gets overridden and calling that method returns null instead of what `CallMe` returns as I'd expect. I think ideally this would follow the CallBase setting, however it doesn't seem like it does.

stakx:
Before Moq can support default interface methods, DynamicProxy must add support for them; see https://github.com/castleproject/Core/issues/447.

stakx:
@hahn-kev: OK, I think this can be made to work. You'll have to configure `CallBase` behavior for your `CallMeDefault` interface method, either at the setup level:

```diff
 var mock = new Mock<ITest>(MockBehavior.Loose);
 mock.Setup(t => t.CallMe()).Returns("hello");
+mock.Setup(t => t.CallMeDefault()).CallBase();
 Assert.NotNull(mock.Object.CallMeDefault());
```

or at the mock level (whatever is a better fit for your tests):

```diff
-var mock = new Mock<ITest>(MockBehavior.Loose);
+var mock = new Mock<ITest>(MockBehavior.Loose) { CallBase = true };
 mock.Setup(t => t.CallMe()).Returns("hello");
 Assert.NotNull(mock.Object.CallMeDefault());
```

In terms of default interface implementations, `.CallBase()` shall call [the most specific override](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods#the-most-specific-override-rule).

PR #1042, opened 2020-07-30
Multitarget .NET Standard 2.1

### LINKED ISSUE #1041: Add .NET Standard 2.1
Since .NET Core 3.1 is an LTS version, support for .NET Standard 2.1 should definitly be added.

twsl:
Since v5 is not released as an official nuget package yet I'm pushing this minor update to the v4 repo instead.

Contains:
* Added netstandard2.1 to TargetFrameworks

Additional:
* Make unit test language invariant

In order to test my changes I ran the unit tests and noticed a language depending unit test. Instead it is now based on the stack trace which makes this language invariant.

Fixes #1041 

I'm not too sure if this minor fix actually requires an entry to the changelog. If you want me to add one, please say so.

### DIFF
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -6,7 +6,7 @@
 	<Import Project="$(BuildDirectory)SourceLink.props" />
 
 	<PropertyGroup>
-		<TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
+		<TargetFrameworks>net45;netstandard2.0;netstandard2.1</TargetFrameworks>
 		<AssemblyName>Moq</AssemblyName>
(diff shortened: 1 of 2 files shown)


stakx:
@twsl, I'd like to merge this, but it might be better to wait until Castle.Core has a `netstandard2.1` target, too. So this could remain open for a while.

But we *could* fix that language-dependent test now. Would you be willing to submit that second commit as a separate PR and remove it from this one?

stakx:
@twsl I've merged this even though there still isn't a DynamicProxy release that will directly support .NET Standard 2.1. Turns out this is still useful in order to implement a workaround for #972.

Thank you for contributing!

PR #1061, opened 2020-09-27
Fix setting of nested non-overridable properties via `Mock.Of`

### LINKED ISSUE #1039: Mock.Of<T>((m) => ...) doesn't play well with Microsoft IOptions<T>
repro repo: https://github.com/mariotee/Moq-MockOf-Issue

I have two identical unit tests. The difference is the version of Moq. `UnitTestProject1` uses Moq 4.14.5 and fails to load IOptions into the properties. `UnitTestProject2` uses Moq 4.13.1 and passes

UPDATE: can confirm this is also broken in 4.14.0. so it might have been somewhere within that minor version change

stakx:
Fixes #1039.

### DIFF
--- a/src/Moq/Mock.cs
+++ b/src/Moq/Mock.cs
@@ -548,6 +548,38 @@ namespace Moq
 			return Mock.Setup(mock, expression, condition);
 		}
 
+		// This specialized version of `SetupSet` exists to let `Mock.Of` support properties that are not overridable.
+		// Note that we generally prefer having a setup for a property's return value, but in this case, that isn't possible.
+		internal static void SetupSet(Mock mock, LambdaExpression expression, PropertyInfo propertyToSet, object value)
+		{
+			Guard.NotNull(expression, nameof(expression));
+
+			Mock.SetupRecursive<MethodCall>(mock, expression, setupLast: (targetMock, _, __) =>
+			{
+				// Setting a mock's property through reflection will only work (i.e. the property will only remember the value
+				// it's being set to) if it is being stubbed. In order to ensure it's stubbed, we temporarily enable
+				// auto-stubbing (if that isn't already switched on).
+
+				var temporaryAutoSetupProperties = targetMock.AutoSetupPropertiesDefaultValueProvider == null;
+				if (temporaryAutoSetupProperties)
+				{
+					targetMock.AutoSetupPropertiesDefaultValueProvider = targetMock.DefaultValueProvider;
+				}
+				try
+				{
+					propertyToSet.SetValue(targetMock.Object, value, null);
+				}
+				finally
+				{
+					if (temporaryAutoSetupProperties)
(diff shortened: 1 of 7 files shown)


src/Moq/Mock.cs
```diff
@@ -548,6 +548,41 @@ internal static MethodCall SetupSet(Mock mock, LambdaExpression expression, Cond
...
+				// the LINQ query such that a final disable of auto-stubbing happens is much more difficult!)
+
+				var temporaryAutoSetupProperties = targetMock.AutoSetupPropertiesDefaultValueProvider == null;
+				if (temporaryAutoSetupProperties)
+				{
+					targetMock.AutoSetupPropertiesDefaultValueProvider = targetMock.DefaultValueProvider;
+				}
+				try
+				{
+					propertyToSet.SetValue(targetMock.Object, value, null);
```
stakx, line 573: Perhaps we could go one step further here and add support for indexers by passing through the index arguments instead of hard-coding `null` as the last argument.

src/Moq/Mock.cs
```diff
@@ -548,6 +548,41 @@ internal static MethodCall SetupSet(Mock mock, LambdaExpression expression, Cond
...
+			Guard.NotNull(expression, nameof(expression));
+
+			Mock.SetupRecursive<MethodCall>(mock, expression, setupLast: (targetMock, _, __) =>
+			{
+				// For strict mocks, we haven't called `SetupAllProperties` on the mock being set up.
+				// Therefore, whenever a property is being initialized, we quickly need to enable auto-stubbing.
+				//
+				// (One would think that it would be simpler to perform `SetupAllProperties` at the beginning
+				// and leave it enabled until the initialized mock is returned to the user. However, transforming
+				// the LINQ query such that a final disable of auto-stubbing happens is much more difficult!)
```
stakx, line 564: This comment no longer makes too much sense after being moved here; perhaps we can simplify things a little further now?

stakx:
Note to self: It may be worth reconsidering whether `InvocationShape` should do any validation at all with regards to method overridability.

It may be better to extract validation from `InvocationShape`, and move it somewhere else. (I was going to say `[Castle]ProxyFactory`, but that wouldn't be totally right, since it isn't just the proxy factories that require methods to be overridable, but lots of other parts of Moq, too.)

Each party that currently builds up `InvocationShape`s would then have to explicitly validate them as needed. This way, the `allowNonOverridable…` parameters being added in this PR could be dropped again.

src/Moq/Mock.cs
stakx: Leaving this unresolved for now. Supporting indexers for non-overridable properties would only work partly anyway, e.g. you could only use fixed values (no matchers) for the index arguments. Not sure if it's worth doing the work to support this scenario as far as is possible.

PR #1082, opened 2020-10-26
Always record calls to += and -= event accessors

### LINKED ISSUE #1058: Remove mandatory SetupAdd & SetupRemove for eventhandler subscription verification.
What is the reason that having to add this setup:
`_mock.SetupRemove(d => d.Event -= It.IsAny<EventHandler>());`

To verify this...
`_mock.VerifyRemove(d => d.Event -= It.IsAny<EventHandler>(), Times.Once);`

... is mandatory?

Normal function calls do not require to be 'registered' beforehand either. I think it will help if these setups are not required and act more like any other `Verify()` call.

stakx:
Resolves #1058.

### DIFF
--- a/tests/Moq.Tests/EventHandlersFixture.cs
+++ b/tests/Moq.Tests/EventHandlersFixture.cs
@@ -1,0 +1,153 @@
+// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD, and Contributors.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
+using System;
+
+using Xunit;
+
+namespace Moq.Tests
+{
+	public class EventHandlersFixture
+	{
+		[Fact]
+		public void Raising_event__directly_on_mock_object__does_not_trigger_handler__if_CallBase_false()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.Object.Event += () => handled = true;
+			mock.Object.RaiseEvent();
+			Assert.False(handled);
+		}
+
+		[Fact]
+		public void Raising_event__directly_on_mock_object__triggers_handler__if_CallBase_true()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>() { CallBase = true };
+			mock.Object.Event += () => handled = true;
+			mock.Object.RaiseEvent();
+			Assert.True(handled);
+		}
+
+		[Fact]
+		public void Raising_event__directly_on_mock_object__does_not_trigger_handler__if_setup_without_CallBase_present()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.SetupAdd(m => m.Event += It.IsAny<Action>());
+			mock.Object.Event += () => handled = true;
+			mock.Object.RaiseEvent();
+			Assert.False(handled);
+		}
+
+		[Fact]
+		public void Raising_event__directly_on_mock_object__triggers_handler__if_setup_with_CallBase_present()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.SetupAdd(m => m.Event += It.IsAny<Action>()).CallBase();
+			mock.Object.Event += () => handled = true;
+			mock.Object.RaiseEvent();
+			Assert.True(handled);
+		}
+
+		[Fact]
+		public void Raising_event__using_Raise__triggers_handler__if_CallBase_false()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.Object.Event += () => handled = true;
+			mock.Raise(m => m.Event += null);
+			Assert.True(handled);
+		}
+
+		[Fact]
+		public void Raising_event__using_Raise__does_not_trigger_handler__if_CallBase_true()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>() { CallBase = true };
+			mock.Object.Event += () => handled = true;
+			mock.Raise(m => m.Event += null);
+			Assert.False(handled);
+		}
+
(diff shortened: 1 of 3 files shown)


tests/Moq.Tests/EventHandlersFixture.cs
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		[Fact]
+		public void Raising_event__directly_on_mock_object__triggers_handler__if_setup_with_CallBase_present()
+		{
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.SetupAdd(m => m.Event += It.IsAny<Action>()).CallBase();
+			mock.Object.Event += () => handled = true;
+			mock.Object.RaiseEvent();
```
siprbaum, line 50: I have to admit it completely escapes me why the `Callbase()` makes a difference here,

tests/Moq.Tests/EventHandlersFixture.cs
```diff
@@ -0,0 +1,153 @@
...
+			Assert.True(handled);
+		}
+
+		[Fact]
+		public void Raising_event__using_Raise__triggers_handler__if_CallBase_false()
```
siprbaum, line 55: This test...

tests/Moq.Tests/EventHandlersFixture.cs
```diff
@@ -0,0 +1,153 @@
...
+			var handled = false;
+			var mock = new Mock<HasEvent>();
+			mock.Object.Event += () => handled = true;
+			mock.Raise(m => m.Event += null);
+			Assert.True(handled);
+		}
+
+		[Fact]
+		public void Raising_event__using_Raise__does_not_trigger_handler__if_CallBase_true()
+		{
```
siprbaum, line 66: ... and this test is also not clear to my why CallBase makes a difference. 
I stop now annotating the tests, as it seems that there is some fundamental difference with CallBase = {true, false} I don't understand.

siprbaum (commented):
Could you elaborate why CallBase makes such a difference for the event raising?
I looked around, e.g. the quick start guide but couldn't figure out why it makes a difference

stakx:
> Could you elaborate why CallBase makes such a difference for the event raising?

I'll try, but keep in mind that I didn't implement that part of Moq and, like you, had to try and make sense of it; so it's possible that the following explanations aren't entirely accurate.

First, consider how one raises an event from outside the class that defines it. The short version is, it can't be done; not without reflecting over private members. C#'s `event`s are a convenient shortcut that generates a delegate-typed backing field, as well as two accessors for modifying that field. But the field isn't accessible outside the class.

That's why Moq builds a completely separate event handling mechanism for its `mock.Raise(m => m.SomeEvent += null)` method. It creates its own delegate-typed backing field for `SomeEvent`. When you invoke `mock.Object.SomeEvent += ...` or `-=`, the handlers are subscribed to _that_ backing field, and `Raise` will be able to invoke those handlers.

Now you've got a problem, however. If you are mocking a class, and that class has some methods that raise `SomeEvent`, that will no longer trigger your subscribed handlers. So you might end up in situations where you can't properly test events / event handlers.

That's where `CallBase` comes it. It basically gives you the choice whether Moq should subscribe handlers to the actual event (i.e. delegate to the base class' `+=`, `-=` accessors, thus `CallBase`) or whether it should use its own mechanism. In the former case, you won't be able to use `Raise`, in the latter case, the base class won't see your event handlers. You need to decide between these two options based on what's desired / needed in your specific test situation.

siprbaum:
Thank you for the explanation. That clear things up for me.
But now I wonder how one is supposed to know that without reading the explanation, e.g. how to get it somehow documented so anyone not reading that detailed explanation has a chance to understand the implications.

stakx:
@siprbaum, that's a good question, and I don't have the answer.

Generally speaking, mocking classes tends to be more complicated than mocking interfaces because you have a base implementation to consider at all stages. For that reason, the first recommendation should be to write / refactor your code in such a way that you can work with (and mock) interfaces. That's generally helpful re: testability, and regarding mocked events, it makes things conceptually straightforward again (as you won't need nor have the choice whether to use `CallBase` or not).

But for those users mocking classes... after all these years, Moq still doesn't have proper documentation. I suspect the quickstart in the wiki was only ever meant to demonstrate the most common, straightforward usages; it probably isn't a good place for more in-depth looks at those aspects of Moq that aren't very intuitive.

That being said, the wiki can be edited by everyone. Feel free to cook up some documentation and add it as a new wiki page.

tests/Moq.Tests/EventHandlersFixture.cs
stakx: As explained below: By using `.Callbase()`, you make sure that when you subscribe to the event, the real `+=` accessor gets used (as opposed to mock's parallel mechanism). So when you raise the real event (through the base class' `RaiseEvent` method), the handler should be triggered. This wouldn't be the case if the handler were subscribed to the mock's "fake" event.

tests/Moq.Tests/EventHandlersFixture.cs
stakx: As explained below: Here the difference is that the event isn't raised through the base class' `RaiseEvent`, but through  `Mock.Raise`, which uses the "fake" event. If you had used `.CallBase()`, the handler would be subscribed in the wrong place and `Mock.Raise` wouldn't be able to trigger it.

tests/Moq.Tests/EventHandlersFixture.cs
stakx: Continuing from my explanation above, this test demonstrates how things go wrong when you use `.CallBase()` as well as `Mock.Raise`: the event handlers won't trigger.

PR #1092, opened 2020-11-01
Add support for nested type matchers

### LINKED ISSUE #919: Add support for nested type matchers
See e.g. the non-ticked-off bullet list item in #908, or #918.

stakx:
Closes #919.

### DIFF
--- a/src/Moq/Extensions.cs
+++ b/src/Moq/Extensions.cs
@@ -252,6 +252,26 @@ namespace Moq
 			}
 		}
 
+		public static bool IsOrContainsTypeMatcher(this Type type)
+		{
+			if (type.IsTypeMatcher())
+			{
+				return true;
+			}
+			else if (type.HasElementType)
+			{
+				return type.GetElementType().IsOrContainsTypeMatcher();
+			}
+			else if (type.IsGenericType)
+			{
+				return type.GetGenericArguments().Any(IsOrContainsTypeMatcher);
+			}
+			else
+			{
+				return false;
+			}
+		}
+
 		public static bool ImplementsTypeMatcherProtocol(this Type type)
 		{
 			return typeof(ITypeMatcher).IsAssignableFrom(type) && type.CanCreateInstance();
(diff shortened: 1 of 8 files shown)


src/Moq/Extensions.cs
```diff
@@ -369,6 +369,94 @@ private static MethodInfo GetInvokeMethodFromUntypedDelegateCallback(Delegate ca
...
+					var se = te.SubstituteTypeMatchers(oe);
+					return se == te ? type : se.MakePointerType();
+				}
+			}
+			else if (type.IsGenericType && other.IsGenericType)
+			{
+				var td = type.GetGenericTypeDefinition();
+				var od = other.GetGenericTypeDefinition();
+
+				if (td == od)
```
stakx, line 454: Never quite sure about the runtime's guarantees re: reference equality of `Type` instances; so `.Equals()` might be safer than `==`.

src/Moq/Extensions.cs
```diff
@@ -369,6 +369,94 @@ private static MethodInfo GetInvokeMethodFromUntypedDelegateCallback(Delegate ca
...
+				{
+					var ta = type.GetGenericArguments();
+					var oa = other.GetGenericArguments();
+					var changed = false;
+
+					for (int i = 0; i < ta.Length; ++i)
```
stakx, line 460: Might want to add a `Debug.Assert(oa.Length == ta.Length)`, explicitly stating the assumption that if generic type definitions match, so will the number of their type parameters (and thus arguments).

src/Moq/Extensions.cs
```diff
@@ -369,6 +369,94 @@ private static MethodInfo GetInvokeMethodFromUntypedDelegateCallback(Delegate ca
...
+					{
+						var sa = ta[i].SubstituteTypeMatchers(oa[i]);
+						if (sa != ta[i])
```
stakx, line 463: Same as above, perhaps use `.Equals() == false` instead of `!=`.

src/Moq/Extensions.cs
```diff
@@ -369,6 +369,94 @@ private static MethodInfo GetInvokeMethodFromUntypedDelegateCallback(Delegate ca
...
+						{
+							changed = true;
+							ta[i] = sa;
+						}
+					}
+
+					return changed ? type.GetGenericTypeDefinition().MakeGenericType(ta) : type;
```
stakx, line 470: Can reuse `td` here.

tests/Moq.Tests/NestedTypeMatchersFixture.cs
```diff
@@ -57,5 +57,23 @@ public void SubstituteTypeMatchers_may_substitute_only_some_parts(Type type, Typ
...
 		#endregion
+
+		#region Test cases
+
+		[Fact]
+		public void It_IsAnyType_used_as_generic_type_argument_of_method()
+		{
+			var mock = new Mock<IX>();
+			mock.Setup(m => m.Method<It.IsAnyType>(It.IsAny<IEnumerable<It.IsAnyType>>())).Verifiable();
+			mock.Object.Method(new string[] { "a", "b" });
```
stakx, line 68: I'm not actually sure I understand why this should work, given the previous tests for `SubstituteTypeMatchers`. Would this successfully match a `string[]` (runtime type) or a `IEnumerable<string>` against `IEnumerable<It.IsAnyType>`? Need to investigate.

Issue #1107, opened 2020-11-24
moq is not compatible with System.Threading.Tasks.Extensions 4.5.3 or 4.5.4


alejandromoncayo:
Nuget gallery dependency states it should be compatible:
https://www.nuget.org/packages/Moq/4.15.1
"System.Threading.Tasks.Extensions (>= 4.5.1)"

But on a project with System.Threading.Tasks.Extensions 4.5.3 or 4.5.4 it fails with the following error:
 Message: 
    Initialization method Trizetto.Cap.Application.Tests.Cache.CacheServiceTests.Setup threw exception. System.TypeInitializationException: The type initializer for 'Moq.DefaultValueProvider' threw an exception. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Threading.Tasks.Extensions, Version=4.2.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified..
  Stack Trace: 
    LookupOrFallbackDefaultValueProvider.ctor()
    EmptyDefaultValueProvider.ctor()
    DefaultValueProvider.cctor()
    --- End of inner exception stack trace ---
    DefaultValueProvider.get_Empty()
    Mock`1.ctor(MockBehavior behavior, Object[] args)
    Mock`1.ctor(MockBehavior behavior)
    Mock`1.ctor()
    CacheServiceTests.Setup() line 21

using the powershell, I can see that from 4.5.0 to 4.5.2 the "version" was 4.2.0.0, but on 4.5.3 it was changed to 4.2.0.1 as displayed by the following powershell script:
Write-Output "4.5.0"
([system.reflection.assembly]::loadfile("C:\Users\admin\.nuget\packages\system.threading.tasks.extensions\4.5.0\lib\netstandard1.0\System.Threading.Tasks.Extensions.dll")).FullName
Write-Output "4.5.1"
([system.reflection.assembly]::loadfile("C:\Users\admin\.nuget\packages\system.threading.tasks.extensions\4.5.1\lib\netstandard1.0\System.Threading.Tasks.Extensions.dll")).FullName
Write-Output "4.5.2"
([system.reflection.assembly]::loadfile("C:\Users\admin\.nuget\packages\system.threading.tasks.extensions\4.5.2\lib\netstandard1.0\System.Threading.Tasks.Extensions.dll")).FullName
write-output "4.5.3"
[...]

PR #1108, opened 2020-11-25
Updating System.Threading.Tasks.Extensions to 4.5.4

JeffAshton:
Fix for issue https://github.com/moq/moq4/issues/1107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Changed
+
+* Upgraded `System.Threading.Tasks.Extensions` dependency to version 4.5.4 (@JeffAshton, #1108)
+
 ## 4.15.1 (2020-11-10)
 
 #### Added
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -42,7 +42,7 @@
 		<PackageReference Include="Castle.Core" Version="4.4.0" />
 		<PackageReference Include="IFluentInterface" Version="2.1.0" PrivateAssets="All" />
 		<PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
-		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.1" />
+		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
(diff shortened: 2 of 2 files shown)


stakx:
Thanks. I'd like to wait a little and see if they push another release of that library in the next few days; increasing the build number might have been a mistake that they're possibly going to fix. If they won't, I'll merge this.

JeffAshton:
@stakx, so this strong version bump did not happen recently.  v4.5.3 was released on 7/9/2019, and v4.5.4 early 2020.

Up until now it's merely been an inconvenience.  Not many of the standard libraries depend on this package.  But with the release of net 5, consumers of this library's dependency `System.Runtime.CompilerServices.Unsafe` have been upgraded. For example, `System.Collections.Immutable` has a dependency chain back to the next strong version of `System.Runtime.CompilerServices.Unsafe`.

Do you have a sense of how long you want to wait?  I'm also surprised they didn't release a 5.0.0, but it looks like everything that was going to get updated had a few months of previews and RCs, but `System.Threading.Tasks.Extensions` has been quiet.

stakx:
I was under the impression that this was a relatively recent change since we only just got a report about it now; but if it's been that way for a while, there's probably no point in waiting.

JeffAshton:
> I was under the impression that this was a relatively recent change since we only just got a report about it now; but if it's been that way for a while, there's probably no point in waiting.

I'm surprised it never got updated.  Some packages above it did:

```
Microsoft.Bcl.AsyncInterfaces (5.0.0)
  -> System.Threading.Tasks.Extensions (4.5.4)
      -> System.Runtime.CompilerServices.Unsafe (4.5.3)
```

And there is a v5.0.0 of `System.Runtime.CompilerServices.Unsafe`, but `System.Threading.Tasks.Extensions` and `System.Memory` haven't been updated yet to consume it.

I think there is value in having a version of `Moq` that points at v4.5.4 of `System.Threading.Tasks.Extensions`, even if short lived, to help bridge this net5.0 gap.  Given examples like the one above and this one:

```
System.Collections.Immutable (5.0.0)
  -> System.Memory (4.5.4)
      -> System.Runtime.CompilerServices.Unsafe (4.5.3)
```

stakx (changes requested):
Almost good to go. Could you please also add a changelog entry at the top of `CHANGELOG.md`:

```diff
The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).


+## Unreleased
+
+#### Changed
+
+* Upgraded `System.Threading.Tasks.Extensions` dependency to version 4.5.4 (@JeffAshton, #1108)
+
+
 ## 4.15.1 (2020-11-10)
```

stakx (approved):

stakx:
Let's do this! 🚀 :wink:

JeffAshton:
Thanks @stakx for the quick turnaround.

PR #1111, opened 2020-11-28
Fix for slow down with many setups on single mock

### LINKED ISSUE #1110: Adding a setup looks like it became an O(n*n) operation since #984 was merged
I don't think this will be an issue for many users but in one of our tests we added 3000 setups to a mock and after updating from 4.13.1 to 4.15.2 the test time went from <10 seconds to over 14 minutes

CeesKaas:
I've played with the code a bit this morning and this change fixes the unittests I've added (at the cost of a bit of extra memory consumption and an extra operation always to save n operations in most cases)

fixes #1110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Fixed
+
+* Added Hashset of active invocations to SetupCollection to only call MarkOverridden when a duplicated expectation is added (@CeesKaas, #1110)
+
 ## 4.15.2 (2020-11-26)
 
 #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Fixed
 
-* Added Hashset of active invocations to SetupCollection to only call MarkOverridden when a duplicated expectation is added (@CeesKaas, #1110)
+* Performance regression: Adding setups to a mock becomes slower with each setup (@CeesKaas, #1110)
 
 ## 4.15.2 (2020-11-26)
 
--- a/src/Moq/SetupCollection.cs
+++ b/src/Moq/SetupCollection.cs
@@ -11,10 +11,12 @@ namespace Moq
 	internal sealed class SetupCollection : ISetupList
 	{
 		private List<Setup> setups;
+		private HashSet<InvocationShape> activeInvocationShapes;
 
 		public SetupCollection()
 		{
 			this.setups = new List<Setup>();
+			this.activeInvocationShapes = new HashSet<InvocationShape>();
 		}
 
 		public int Count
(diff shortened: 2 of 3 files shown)


tests/Moq.Tests/MockFixture.cs
```diff
@@ -1312,5 +1312,37 @@ public void Accessing_property_of_bad_serializable_type_after_SetupAllProperties
...
+			var sw = System.Diagnostics.Stopwatch.StartNew();
+			var mock = new Mock<ITestInterface>();
+			var id = Guid.NewGuid();
+			for (int i = 0; i < 1000; i++)
+			{
+				mock.Setup(m => m.Get(id));
+			}
+			sw.Stop();
+			Assert.True(sw.Elapsed < TimeSpan.FromSeconds(1), $"Setups were expected to take less than a second but it took {sw.Elapsed}");
+		}
```
stakx: Timed unit tests are a bad idea because we don't really have a way of determining an absolute time threshold. These tests may never fail to begin with on a really fast machine; and they may still fail (even after your proposed change) on a very slow machine.

Perhaps the only reference time we have is that of previous Moq versions... but those cannot be easily used in this test suite.

I can't think of a good way to write these unit tests, so unless you can, I suggest we remove them.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Fixed
+
+* Added Hashset of active invocations to SetupCollection to only call MarkOverridden when a duplicated expectation is added (@CeesKaas, #1110)
```
stakx: Most Moq users will probably have no idea what exactly this means, as you focus on Moq implementation details instead of what it means from a user's perspective. May I suggest that you rephrase this somewhere along these lines:

```suggestion
* Performance regression: Adding setups to a mock becomes slower with each setup (@CeesKaas, #1110)
```

stakx (changes requested):
Hi @CeesKaas, thanks for your patience.

I do agree that we should do something about this problem. I've thought about it for a bit, and your solution indeed seems to easiest.

Just a couple of details that need your attention. Then this should be good to merge!

tests/Moq.Tests/MockFixture.cs
CeesKaas: i fully agree with timed unittests being a bad idea but i didn't know a different way of testing the problem but i'm fine with removing them

src/Moq/SetupCollection.cs
```diff
@@ -101,6 +105,7 @@ public void Clear()
 			lock (this.setups)
 			{
 				this.setups.Clear();
+				this.activeInvocationShapes.Clear();
```
stakx, line 108: Just a remark for posteriority: in theory, we should also filter out removed setups in `RemoveAllPropertyAccessorSetups`. Not doing means that the first time a property setup is recreated after a call to `mock.SetupAllProperties()`, there may be a superfluous call to `MarkOverriddenSetups`. This is an edge case however, and the overhead of the additional `MarkOverriddenSetups` in this case is probably quite small, so let's just ignore this for now.

stakx (approved):

stakx:
All good. Thanks for contributing, @CeesKaas! 🚀

PR #1115, opened 2020-12-09
Added GetInvocationsOf extension method

Mrxx99:
This adds one extension method to `Mock<T>` called `GetInvocationsOf` which allows to get the invocations on one specific method instead of all methods like with `Mock.Invocations`.

This acts like the `GetArgumentsForCallsMadeOn` method from Rhino.Mocks.
It can be used like this:
```csharp
	var mock = new Mock<IProcessor>();
		
	... // exercise mock
		
	var invocationsOnExecute = mock.GetInvocationsOf(x => x.Execute(default(string)));
```

Like the Rhino.Mocks verison it ignores the values of the arguments on the specified method.
This is very useful when you need to inspect the arguments of a method call manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+#### Added
+
+* Added extension method `GetInvocationsOf` which returns only the invocations on the specified method (same lamda specification like in Setup/Verify).
+
 
 ## 4.15.2 (2020-11-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Added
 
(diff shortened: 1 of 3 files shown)


stakx:
Thanks for your interest in improving Moq. Unfortunately though, I'm probably not going to merge this because I think this extension method is too particular.

If you only want to filter invocations by method name:

1. We generally have no methods in the public API that target whole method groups. This would be a first, and therefore an exception.

2. Doing a simple LINQ query like `mock.Invocations.Where(i => i.Method.Name == "Execute")` might be good enough; no new API needed.

However, if you actually want to filter not just by method name, but also match method signatures, it makes sense to include parameters somehow. You could still go with (2) and write a LINQ query, but I'll admit it might get a little ugly / complex.

3. Having to specify arguments that are then ignored would probably not be sufficiently intuitive. I foresee users reporting problems because they expected e.g. matchers to work here.

A more generally useful addition to the public API might be (for example) a `invocation.Matches(callExpression)` helper method which you could then apply to your scenario as follows:

```csharp
fooMock.Invocations.Where(i => i.Matches(foo => foo.Execute(It.IsAny<...>, ...)))
```

In closing, I'd prefer not to add the proposed extension method to the core library's API &ndash; better to either keep it in a third-party contrib package; or to add a method that is more widely applicable.

Thanks again, however, for contributing!

PR #1116, opened 2020-12-10
Changed TypeNotMockable exception message

### LINKED ISSUE #1112: TypeNotMockable Exception
Can the Moq Gaurd class exception below also return the type name? it would be very helpful when mocking objects that take multiple constructor parameters.

		public static void IsMockable(Type type)
		{
			if (!type.IsMockable())
			{
				throw new NotSupportedException(Resources.TypeNotMockable);
			}
		}

jacker92:
Modified Exception message to include the type name. 

Fixes #1112

### DIFF
--- a/src/Moq/Guard.cs
+++ b/src/Moq/Guard.cs
@@ -204,7 +204,10 @@ namespace Moq
 		{
 			if (!type.IsMockable())
 			{
-				throw new NotSupportedException(Resources.TypeNotMockable);
+				throw new NotSupportedException(
+					string.Format(
+						Resources.TypeNotMockable,
+						type.Name));
 			}
 		}
 
--- a/src/Moq/Guard.cs
+++ b/src/Moq/Guard.cs
@@ -207,7 +207,7 @@ namespace Moq
 				throw new NotSupportedException(
 					string.Format(
 						Resources.TypeNotMockable,
-						type.Name));
+						type.GetFormattedName()));
 			}
 		}
 
--- a/src/Moq/Properties/Resources.Designer.cs
+++ b/src/Moq/Properties/Resources.Designer.cs
@@ -572,7 +572,7 @@ namespace Moq.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Mocked type {0} must be an interface, a delegate, or a non-sealed, non-static class..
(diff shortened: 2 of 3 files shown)


src/Moq/Guard.cs
```diff
@@ -204,7 +204,10 @@ public static void IsMockable(Type type)
 		{
 			if (!type.IsMockable())
 			{
-				throw new NotSupportedException(Resources.TypeNotMockable);
+				throw new NotSupportedException(
+					string.Format(
+						Resources.TypeNotMockable,
+						type.Name));
```
stakx: For consistency with existing code, please use the `type.GetFormattedName()` extension method:

```suggestion
						type.GetFormattedName()));
```

src/Moq/Properties/Resources.resx
```diff
@@ -121,7 +121,7 @@
     <value>Expression {0} involves a field access, which is not supported. Use properties instead.</value>
   </data>
   <data name="TypeNotMockable" xml:space="preserve">
-    <value>Type to mock must be an interface, a delegate, or a non-sealed, non-static class.</value>
+    <value>Mocked type {0} must be an interface, a delegate, or a non-sealed, non-static class.</value>
```
stakx: Since the type isn't actually mocked yet, _"mocked type"_ is less correct than the existing phrasing, _"type to mock"_. I'd suggest changing this to:

```suggestion
    <value>Type to mock ({0}) must be an interface, a delegate, or a non-sealed, non-static class.</value>
```

stakx (changes requested):
Just two small details, then we should be good to go!

src/Moq/Properties/Resources.Designer.cs
```diff
@@ -572,7 +572,7 @@ internal class Resources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Type to mock must be an interface, a delegate, or a non-sealed, non-static class..
+        ///   Looks up a localized string similar to Mocked type {0} must be an interface, a delegate, or a non-sealed, non-static class..
```
stakx: Please have VS regenerate this file so that it is in sync with `Resources.resx`.

stakx (changes requested):

stakx (approved):

PR #1126, opened 2021-01-01
Add ability to set up the `.Result` of (value) tasks

### LINKED ISSUE #384: Async Method Support Clean-up
Right now, Moq doesn't have a proper support for async methods. There are a few methods giving Moq some Async support. Those are spreaded through out Moq fluentApi like ReturnAsync() and ThrowAsync() but it's not developer friendly.
I suggest to implement FluentApi support for async methods starting from SetupAsync() and expose only async functionality.
At the same time, remove all the async methods from the current Setup().
Let me know your thoughts. Should I do a request pull and start working on it?
Regards

Example:
interface IComponent {
&nbsp;&nbsp;   int Process(int value);
&nbsp;&nbsp;   Task&lt;int&gt; ProcessAsync(int value);
}
componentMock.Setup(I => Process(It.Any&lt;int&gt;()).Return(1);                               // normal method 
componentMock.Setup(I => Process(It.Any&lt;int&gt;()).Throw(new Exception());        // normal method 
componentMock.Setup(I => Process(It.Any&lt;int&gt;()).Callback<int>(i => capt = i); // normal method 
componentMock.SetupAsync(I => ProcessAsync(It.Any&lt;int&gt;()).Return(1);                                // full async support
componentMock.SetupAsync(I => ProcessAsync(It.Any&lt;int&gt;()).Throw(new Exception());          // full async support
componentMock.SetupAsync(I => ProcessAsync(It.Any&lt;int&gt;()).Callback<int>(i => capt = i);   // full async support

### LINKED ISSUE #1007: Easier async setups through a new `Await(...)` operator

stakx:
This much more focused version of #1123 only covers setting up the `.Result` of (value) task-based async methods.

## Benefits

* The dedicated async setup methods `.ReturnsAsync`, `.ThrowsAsync` etc. are now largely superfluous:

   ```diff
   -mock.Setup(x => x.GetFooAsync()).ReturnsAsync(foo)
   +mock.Setup(x => x.GetFooAsync().Result).Returns(foo)
   ```

   From a user perspective, setups become more uniform in shape. From a maintainer's perspective, we have less work. More specifically, we no longer need to ensure that those async setup verbs are at feature parity with their non-async counterparts.

* It adds async support where such async verbs aren't currently available:

   ```diff
   -Mock.Of<X>(x => x.GetFooAsync() == Task.FromResult(foo))
   +Mock.Of<X>(x => x.GetFooAsync().Result == foo)
   ```

   but also in other places like in `Verify…(callExpression, …)`.

* It enables recursive setups across async calls in a single setup expression. For example, using just `Mock.Of`:

   ```diff
   -Mock.Of<X>(x => x.GetFooAsync() == Task.FromResult(Mock.Of<IFoo>(f => f.Bar == bar)))
   +Mock.Of<X>(x => x.GetFooAsync().Result.Bar == bar)
   ```

## Should `.ReturnsAsync`, `.ThrowsAsync` etc. be marked as `[Obsolete]`?

I am not marking those methods as `[Obsolete]` just yet, mainly for two reasons:

 * There are a few specialized overloads (e.g. for delayed task completion) that would need to be rewritable using just the regular setup verbs. I'd like to look at those separately.

 * `.PassAsync` for sequence setups doesn't have an easy replacement, because non-generic tasks do not have a `.Result` property that one could set up instead. Replacing `.PassAsync` would require an `await`-like operator substitute (as proposed in #1007 and implemented in #1123), which would likely be added together with support for custom awaitable types. Until that happens, it doesn't make sense to deprecate only some (but not all) async setup verbs.

## What is still missing?

As noted in the new changelog entry, support in `mock.Protected()` is still missing, or at least spotty. Also, unlike #1123, there is no support for custom awaitable types just yet; however, most of the necessary machinery is now in place, so this should be relatively easy to add in the future.

Resolves #384, closes #1007, closes #1123.

### DIFF
--- a/src/Moq/ExpressionExtensions.cs
+++ b/src/Moq/ExpressionExtensions.cs
@@ -10,6 +10,7 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
 
+using Moq.Async;
 using Moq.Properties;
 using Moq.Protected;
 
@@ -294,6 +295,16 @@ namespace Moq
 					{
 						var memberAccessExpression = (MemberExpression)e;
 						Debug.Assert(memberAccessExpression.Member is PropertyInfo);
+
+						if (IsResult(memberAccessExpression.Member, out var awaitableFactory))
+						{
+							Split(memberAccessExpression.Expression, out r, out p);
+							p.AddResultExpression(
+								awaitable => Expression.MakeMemberAccess(awaitable, memberAccessExpression.Member),
+								awaitableFactory);
+							return;
+						}
+
 						r = memberAccessExpression.Expression;
 						var parameter = Expression.Parameter(r.Type, r is ParameterExpression ope ? ope.Name : ParameterName);
 						var property = memberAccessExpression.GetReboundProperty();
@@ -326,6 +337,15 @@ namespace Moq
 						Debug.Assert(!CanSplit(e));
 						throw new InvalidOperationException();  // this should be unreachable
 				}
+			}
+
+			bool IsResult(MemberInfo member, out IAwaitableFactory awaitableFactory)
+			{
+				var instanceType = member.DeclaringType;
+				awaitableFactory = AwaitableFactory.TryGet(instanceType);
+				var returnType = member switch { PropertyInfo p => p.PropertyType,
+				                                 _              => null };
+				return awaitableFactory != null && object.Equals(returnType, awaitableFactory.ResultType);
 			}
 		}
 
--- a/src/Moq/Invocation.cs
+++ b/src/Moq/Invocation.cs
@@ -7,6 +7,8 @@ using System.Diagnostics;
 using System.Reflection;
 using System.Text;
 
+using Moq.Async;
+
 namespace Moq
 {
 	internal abstract class Invocation : IInvocation
@@ -86,6 +88,18 @@ namespace Moq
 			{
 				Debug.Assert(this.result == null);
 				this.result = new ExceptionResult(value);
+			}
+		}
+
+		public void ConvertResultToAwaitable(IAwaitableFactory awaitableFactory)
+		{
+			if (this.result is ExceptionResult r)
+			{
+				this.result = awaitableFactory.CreateFaulted(r.Exception);
+			}
+			else if (this.result != null && !this.method.ReturnType.IsAssignableFrom(this.result.GetType()))
+			{
+				this.result = awaitableFactory.CreateCompleted(this.result);
 			}
 		}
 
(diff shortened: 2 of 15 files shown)


src/Moq/ExpressionExtensions.cs
```diff
@@ -294,6 +294,16 @@ void Split(Expression e, out Expression r /* remainder */, out InvocationShape p
 					{
 						var memberAccessExpression = (MemberExpression)e;
 						Debug.Assert(memberAccessExpression.Member is PropertyInfo);
+
+						if (IsResult(memberAccessExpression.Member, out var awaitableFactory))
+						{
+							Split(memberAccessExpression.Expression, out r, out p);
+							p.AddResultExpression(
+								awaitable => Expression.MakeMemberAccess(awaitable, memberAccessExpression.Member),
+								awaitableFactory);
```
stakx, line 304: Once we add support for custom awaitable types, it will be better to preserve the original expression, instead of replacing the `.Result` access with a new result expression.

src/Moq/Invocation.cs
```diff
@@ -89,6 +91,18 @@ public Exception Exception
 			}
 		}
 
+		public void ConvertResultToAwaitable(IAwaitableFactory awaitableFactory)
+		{
+			if (this.result is ExceptionResult r)
+			{
+				this.result = awaitableFactory.CreateFaulted(r.Exception);
+			}
```
stakx, line 99: There may be circumstances where we may have to call the other `CreateFaulted(IEnumerable<Exception> exceptions)` overload instead (which, I'm assuming, is meant to cover `AggregateException`). I haven't yet looked into whether this is actually necessary. Perhaps it would be better to remove that other overload until that time?

src/Moq/Setup.cs
```diff
@@ -65,7 +65,25 @@ public void Execute(Invocation invocation)
...
+			if (this.expectation.HasResultExpression(out var awaitableFactory))
+			{
+				try
+				{
+					this.ExecuteCore(invocation);
+				}
+				catch (Exception exception)
+				{
+					invocation.Exception = exception;
+				}
```
stakx, line 77: We may have to change this some day in order to account for several / aggregate exceptions.

PR #1127, opened 2021-01-04
Allow InSequence to be used with protected mocks

FTWinston:
Added a couple of overloads of InSequence to work with IProtectedMock and IProtectedAsMock, in addition to working with Mock directly.

Exposed the underlying mock on (I)ProtectedMock and (I)ProtectedAsMock to support this.

If I've missed something obvious and there's an existing way of testing the sequence of calls to protected methods without getting into callback hell, then by all means close this off. But otherwise, this is (IMO) a quick and convienient minor addition to support that specific case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
    This should work in all principal setup methods (`Mock.Of`, `mock.Setup…`, `mock.Verify…`). Support in `mock.Protected()` and for custom awaitable types may be added in the future. (@stakx, #1126)
 
+* `InSequence` can now be used with protected mocks (@FTWinston, #1127)
+
 #### Changed
 
 * Attempts to mark conditionals setup as verifiable are once again allowed; it turns out that forbidding it (as was done in #997 for version 4.14.0) is in fact a regression. (@stakx, #1121)
--- a/src/Moq/MockSequence.cs
+++ b/src/Moq/MockSequence.cs
@@ -5,6 +5,7 @@ using System.ComponentModel;
 
 using Moq.Language;
 using Moq.Language.Flow;
+using Moq.Protected;
 
 namespace Moq
 {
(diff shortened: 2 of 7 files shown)


FTWinston:
If adding `UnderlyingMock` to `IProtectedMock` and `IProtectedAsMock` doesn't pass the smell test, the `InSequence` overloads could be changed to cast from `IProtectedMock` to `ProtectedMock` and `IProtectedAsMock` to `ProtectedAsMock`.

I don't see any way around doing one of these approaches without extensive refactoring, but I'm certainly open to suggestions.

stakx:
Thanks @FTWinston for your clean and nicely written PR.

> If adding `UnderlyingMock` to `IProtectedMock` and `IProtectedAsMock` doesn't pass the smell test[...]

It's the quickest route to where you want to go, but as you seem to suspect yourself, it definitely isn't the best (from an architectural design viewpoint).

First, the ability to transition from `mock.Protected()` back to `mock` is unlikely to be useful, so ideally `UnderlyingMock` shouldn't be there.

Second, you may have seen that sequences are implemented on top of conditional setups. As it happens, `mock.Protected()` is also missing `.When`. If we had _that_, `.InSequence` could likely be implemented on top of it without needing `UnderlyingMocks`... and we'd reach even better feature parity with the regular setup API. That would be the best fit with the API as it is today.

Unfortunately, implementing `.When` requires two interfaces for the fluent API with much of `IProtected[As]Mock` duplicated (and perhaps a separate backing implementation but I'm not entirely certain here).

I have been trying to refactor Moq internals in the direction of being more generic, hoping that we can someday reduce all the existing duplication in the public API. (Yes, that would be a breaking change!) That's why I am very reluctant to go in the opposite direction here.

My preference would be to not pursue this any further... but if it's pressing and you want to go on, I might consider merging a solution that adds `.When` to `mock.Protected().As<>` (and only there!), then layers `MockSequence.For` and `mock.Protected().As<>(...).InSequence` overloads on top of that.

What are your thoughts on this?

P.S.:

> If I've missed something obvious and there's an existing way of testing the sequence of calls to protected methods without getting into callback hell

You haven't missed anything, I suspect this isn't easily possible. But then it may be worth more if you refactored your code to be more testable. Moq's (limited) ability to test protected methods is convenient but it perhaps doesn't encourage the best possible code.

FTWinston:
Fair enough, thanks @stakx for that thorough explanation.

I was adding tests to existing code (can you tell?), so I've refactored it to make public everything that's being mocked.

Based on your answer, I've no objection to closing this PR.

stakx:
@FTWinston, :+1:. Great to hear you've found a way to test your code. Last but not least, thank you for contributing!

PR #1130, opened 2021-01-16
`.CallBase` for default interface implementations

### LINKED ISSUE #972: Interface Default methods are ignored

stakx:
As of version 4.4.0, DynamicProxy is not able to reliably `.Proceed()` to a base method implementation if it is provided by an interface. From a user perspective, one would likely try to invoke a default interface implementation using `.CallBase()`.

This is an attempt to work around this current DynamicProxy limitation by performing the non-virtual call to the interface method manually. This involves some System.Reflection.Emit code inspired by @thomaslevesque's approach mentioned in https://github.com/castleproject/Core/issues/447#issuecomment-536274305.

~~This works mostly, except for one important detail: Moq will always call the least specific default implementation, instead of the most specific one. (This matters when one overrides a default implementation in a derived interface, then mocks that derived interface or a type inheriting/implementing it.)~~ (Update: Most specific override identification is now implemented, too.)

The code being added here is rather horrible, and it would be very good to get rid of it as soon as DynamicProxy properly supports proceeding to default interface implementations.

Resolves #972.

### DIFF
--- a/src/Moq/ProxyFactories/CastleProxyFactory.cs
+++ b/src/Moq/ProxyFactories/CastleProxyFactory.cs
@@ -2,15 +2,17 @@
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
 using System;
+using System.Diagnostics;
+using System.Reflection;
+
+#if FEATURE_DEFAULT_INTERFACE_IMPLEMENTATIONS
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using System.Reflection.Emit;
 using System.Runtime;
 using System.Text;
+#endif
 
 using Castle.DynamicProxy;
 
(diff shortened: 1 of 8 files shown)


src/Moq/ProxyFactories/CastleProxyFactory.cs
```diff
@@ -136,6 +151,108 @@ public void DetachFromUnderlying()
...
+					for (var i = 0; i < n; ++i)
+					{
+						il.Emit(originalParameterTypes[i].IsByRef ? OpCodes.Ldloca : OpCodes.Ldloc, arguments[i]);
+					}
+					il.Emit(OpCodes.Call, method);
+
+					// Put the return value in a local variable for later retrieval.
+					if (method.ReturnType != typeof(void))
+					{
+						il.Emit(OpCodes.Box, method.ReturnType);
```
stakx: This would fail on a by-ref return type (which AFAIK neither we nor DynamicProxy currently support anyway).

PR #1140, opened 2021-02-02
Improve error message when verification fails

### LINKED ISSUE #1138: Improve error message when verification fails
I would like to improve the verification error message for these two specific scenarios:

1. Currently, if you have a variable of type `LambdaExpression`, and you pass it in to the `It.Is<TValue>` static method, and verification fails, then the formatting of the lambda expression in the exception message isn't as good as it should be. Specifically, the representation of the lambda expression in the message is poor because `ToString()` has been called on it instead of the friendly expression formatting routine.

    ```c#
    var mock = new Mock<IFoo>();

    // Declare a variable to hold the lambda expression.
    Expression<Func<string, bool>> nullOrEmpty = s => string.IsNullOrEmpty(s);

    // Pass the lambda expression variable to `It.Is`.
    mock.Setup(x => x.Execute(It.Is(nullOrEmpty)))
        .Returns("ack")
        .Verifiable();

    var mex = Assert.Throws<MockException>(() => mock.Verify());
    Assert.Contains(@".Execute(It.Is<string>(s => string.IsNullOrEmpty(s)))", mex.Message);
    ```

    The second assertion fails with the message (nobody wants to see "<>c__DisplayClass" - yuck!):

    ```
    This mock failed verification due to the following:

          VerifyFixture.IFoo x => x.Execute(It.Is<string>(s => IsNullOrEmpty(s))):
          This setup was not matched.
    ```

    Note the lambda expression `s => IsNullOrEmpty(s)`, when it should be `s => string.IsNullOrEmpty(s)`. There seems to be a difference in how a lambda expression is constructed depending on whether it is captured in a variable or not. When captured in a variable, the overall expression contain a `ConstantExpression` which has a value of type LambdaExpression - this is the actual lambda expression that we want. However, currently, the value of all constant expressions is basically formatted by calling `ToString()` on them.
     
2. Assuming scenario 1 (where we have a variable of type lambda expression passed to `It.Is`), if a lambda expression captures a local variable as a closure, then we should *not* include the ugly name of the closure class (beginning with "<>c__DisplayClass") in the verification failed message.

    ```c#
    var mock = new Mock<IFoo>();

    string password = "abc123";

    Expression<Func<string, bool>> matchingPassword = s => s == password;

    mock.Setup(x => x.Execute(It.Is(matchingPassword)))
        .Returns("ack")
        .Verifiable();

    var mex = Assert.Throws<MockException>(() => mock.Verify());
    Assert.Contains(@".Execute(It.Is<string>(s => s == password)", mex.Message);
    ```
    
    The second assertion fails with the message (nobody wants to see "<>c__DisplayClass" - yuck!):

    ```
    This mock failed verification due to the following:
    
       VerifyFixture.IFoo x => x.Execute(It.Is<string>(s => s == VerifyFixture.<>c__DisplayClass5_0.password)):
       This setup was not matched.
    ```
    
I have a [branch containing these tests, along with a fix for each](https://github.com/moq/moq4/compare/main...bfriesen:lambda_closure_support). If this is something you're interested in, I'll make a pull request. I do have one question though: for the purpose of the updating the changelog, is this an 'Added', 'Changed' or 'Fixed'?

bfriesen:
Fixes #1138.

### DIFF
--- a/tests/Moq.Tests/VerifyFixture.cs
+++ b/tests/Moq.Tests/VerifyFixture.cs
@@ -83,6 +83,24 @@ namespace Moq.Tests
 		}
 
 		[Fact]
+		public void ThrowsWithExpressionIfVerifiableExpectationWithLambdaMatcherVariableAndClosureAccessNotCalled()
+		{
+			var mock = new Mock<IFoo>();
+
+			string password = "abc123";
+
+			Expression<Func<string, bool>> matchingPassword = s => s == password;
+
+			mock.Setup(x => x.Execute(It.Is(matchingPassword)))
+				.Returns("ack")
+				.Verifiable();
+
+			var mex = Assert.Throws<MockException>(() => mock.Verify());
+			Assert.True(mex.IsVerificationError);
+			Assert.Contains(@".Execute(It.Is<string>(s => s == password)", mex.Message);
+		}
+
+		[Fact]
 		public void VerifiesNoOpIfNoVerifiableExpectations()
 		{
 			var mock = new Mock<IFoo>();
--- a/tests/Moq.Tests/VerifyFixture.cs
+++ b/tests/Moq.Tests/VerifyFixture.cs
@@ -2,6 +2,7 @@
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
 using System;
+using System.Linq.Expressions;
 using System.Threading.Tasks;
 
 using Moq;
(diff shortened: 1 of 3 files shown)


tests/Moq.Tests/VerifyFixture.cs
```diff
@@ -65,6 +66,40 @@ public void ThrowsWithExpressionIfVerifiableExpectationWithLambdaMatcherNotCalle
...
+
+			Expression<Func<string, bool>> matchingPassword = s => s == password;
+
+			mock.Setup(x => x.Execute(It.Is(matchingPassword)))
+				.Returns("ack")
+				.Verifiable();
+
+			var mex = Assert.Throws<MockException>(() => mock.Verify());
+			Assert.True(mex.IsVerificationError);
+			Assert.Contains(@".Execute(It.Is<string>(s => s == password)", mex.Message);
```
stakx, line 100: Small detail (but don't worry about it): missing closing parenthesis at the end of the string.

stakx (approved):
Looks good to me! :+1:

PR #1165, opened 2021-05-29
Add "set" methods to ProtectedAsMock

tonyhallett:
Fix #1164

PR #1174, opened 2021-06-22
InSequence for protected mocks

### LINKED ISSUE #1173: InSequence is not available for protected
Given the following types it is not currently possible to combine them in a sequence.

```
public abstract class Protected
{
	protected abstract int ProtectedDo(int arg);
	public int InvokeProtectedDo(int arg)
	{
		return ProtectedDo(arg);
	}
}

public interface ProtectedLike
{
       int ProtectedDo(int arg);
}

public interface IFoo
{
	int Do(int arg);
}
```

This is currently not possible
```

[Fact]
public void WorksWithProtectedAsSuccess()
{
	var a = new Mock<IFoo>(MockBehavior.Strict);
	var protectedAs = new Mock<Protected>().Protected().As<ProtectedLike>();
	var sequence = new MockSequence();
	a.InSequence(sequence).Setup(x => x.Do(100)).Returns(101);
	a.InSequence(sequence).Setup(x => x.Do(200)).Returns(201);
        // ************ this is not possible
	protectedAs.InSequence(sequence).Setup(y => y.ProtectedDo(300)).Returns(301);

	Assert.Equal(101, a.Object.Do(100));
	Assert.Equal(201, a.Object.Do(200));
        //protectedAs.Mocked added in pull request
	Assert.Equal(301, protectedAs.Mocked.Object.InvokeProtectedDo(300));
			
}
```

[Back this issue](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)
[Back this issue](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)

tonyhallett:
Will resolve #1173 when implementations are provided for two Setup methods as described below.

For discussion

There is common code with existing WhenPhrase and Mock methods that differ only in the condition - for instance
```
//When

		public ISetup<T> Setup(Expression<Action<T>> expression)
		{
			var setup = Mock.Setup(mock, expression, this.condition);
			return new VoidSetupPhrase<T>(setup);
		}
//Mock<T>

		public ISetup<T, TResult> Setup<TResult>(Expression<Func<T, TResult>> expression)
		{
			var setup = Mock.Setup(this, expression, null);
			return new NonVoidSetupPhrase<T, TResult>(setup);
		}
```

The code in this pull request follows suit.  WhenPhraseProtected differs to ProtectedAsMock only by condition.

---

Note that WhenPhraseProtected throws NotImplementedException for SetupSet methods
These can be added to match https://github.com/moq/moq4/pull/1165

### DIFF
--- a/src/Moq/Language/Flow/MockProtectedSequencePhrase.cs
+++ b/src/Moq/Language/Flow/MockProtectedSequencePhrase.cs
@@ -11,25 +11,17 @@ using Moq.Protected;
 
 namespace Moq.Language.Flow
 {
-	internal sealed class WhenPhraseProtected<T,TAnalog> : ISetupConditionResultProtected<T,TAnalog>
+	internal sealed class MockProtectedSequencePhrase
+		<T, TAnalog> : MockSequencePhraseBase<T, TAnalog>
 		where T : class
 		where TAnalog : class
 	{
 		private static DuckReplacer DuckReplacerInstance = new DuckReplacer(typeof(TAnalog), typeof(T));
 
-		private Mock<T> mock;
-		private Condition condition;
+		public MockProtectedSequencePhrase(IProtectedAsMock<T, TAnalog> protectedAsMock, IMockSequence mockSequence) : base(protectedAsMock.Mocked, mockSequence) { }
 
-		public WhenPhraseProtected(IProtectedAsMock<T, TAnalog> protectedAsMock, Condition condition)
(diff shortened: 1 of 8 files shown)


stakx:
@tonyhallett, same here as in your other PR, and as mentioned in the issue belonging to this PR: this isn't the right time to review code changes as long as no decision has been reached about the future direction for sequences. I'm not convinced at this time that the current `.InSequence` / `MockSequence` mechanism is the right way forward.

PR #1193, opened 2021-07-31
new mock sequence - for discussion

tonyhallett:
A base class that provides functionality for the concept of a mock sequence 

A new mock sequence type with the following behaviour

Captures invocations performed on the registered mocks to facilitate the concept of strict vs loose sequences.
Setups are performed on the mocks and not the sequence and you can now specify how many Times these "sequence setups" can have matching invocations.

Note that this implementation does not have the concept of Cyclic - lets discuss further.
When the invocations do not match the sequence an exception is thrown immediately and is not dependent on the MockBehavior of the mock with the matching setup. - let's discuss further.

Usage

```
var mock = new Mock<IFoo>();
var mocked = mock.Object;

var protectedMock = new Mock<Protected>();
var protectedAsMock = protectedMock.Protected().As<ProtectedLike>();
var protectedMocked = protectedMock.Object;

// loose sequence
var mockSequence = new NewMockSequence(false, mock, protectedMock);

// VerifiableSetup can be Verified - this will be worked on and may change and may be deemed unnecessary.
var verifiableSetup = mockSequence.Setup(() => mock.Setup(m => m.Do(1)), Times..AtLeast(1));
mockSequence.Setup(() => protectedAs.Setup(p => p.ProtectedDo(1))); // defaults to Times.Once()

//act - may throw exception

mockSequence.Verify()
```

How it works

Listens for changes to setups ( so no special code for protected mocks ) - adding a condition ( internal code change ) so that the sequence can be notified when setups are matched.
To faciliate using the same setup multiple times internal code changed so that "overridden" setups can be removed.

Listens for invocations on the registered mocks ( internal code change to register for additions - tbd behaviour change when invocations cleared )

Some internal information on Times has been surfaced internally to allow using Times in the context of a sequence.

### DIFF
--- a/src/Moq/Interception/Mock.cs
+++ b/src/Moq/Interception/Mock.cs
@@ -1,6 +1,8 @@
 // Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD, and Contributors.
 // All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
 
+using System;
+
 namespace Moq
 {
 	partial class Mock : IInterceptor
--- a/src/Moq/InvocationCollection.cs
+++ b/src/Moq/InvocationCollection.cs
@@ -65,6 +65,7 @@ namespace Moq
 
 				this.invocations[this.count] = invocation;
 				this.count++;
+				owner.AddedInvocation(invocation);
 			}
 		}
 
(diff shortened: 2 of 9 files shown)


tonyhallett:
Unfortunately this in its current form is unable to provide necessary behaviour

PR #1196, opened 2021-08-03
Ref out support for protected mock

### LINKED ISSUE #1195: `mock.Protected()` does not support methods with ref or out parameters
```
public class HasMethodWithRefParameter
{
	protected virtual void RefMethod(ref int arg) { }
}

[Fact]
public void ProtectedMockDoesNotSupportMethodsWithRefOrOutParameters()
{
	var mock = new Mock<HasMethodWithRefParameter>();
	
       // System.ArgumentException : No protected method HasMethodWithRefParameter.RefMethod found whose signature is compatible with the provided arguments (int).
        mock.Protected().Setup("RefMethod", 1);		
}

```

and this is not allowed 

```
int refInt = 1;
mock.Protected().Setup("RefMethod", ref refInt);
```

tonyhallett:
closes #1195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 #### Added
 
+* `mock.Protected()` does not support methods with ref or out parameters (@tonyhallett, #1196)
 * `SetupSet`, `VerifySet` methods for `mock.Protected().As<>()` (@tonyhallett, #1165)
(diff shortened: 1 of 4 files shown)


tonyhallett:
Note that I have removed the ItIsRefAny check as this is no necessary as is covered by 
`if (!type.IsAssignableFrom(expression.GetType()))` as the type is a by ref type and ItExpr.Ref is not

I added the changelog entry to Added - do you want it to be Fixed ?

stakx:
Closing this along with the issue; see my explication there.

Issue #1066, opened 2020-10-06
Property setups are ignored on mocks instantiated using `Mock.Of`


stakx:
This report goes back to a [Gitter chat post by @bluechrism on Sep 3 2020](https://gitter.im/moq/moq?at=5f51383159ac794e02e69756):

> ![image](https://user-images.githubusercontent.com/104481/95241454-0d432280-080e-11eb-9a96-4c4fc28ed07e.png)
>
> [...] can someone explain this difference between `new Mock<T>` and using `Mock.Of<T>` with `Mock.Get` to set up a property: Again, both the Tests would pass with Mock 4.7.8 but the second now fails with 4.14.5.

Moq 4.12.0 introduced a behavioral change that leads to property setups sometimes being ignored on mocks that were created with `Mock.Of<>` (or on mocks on whom `SetupAllProperties` was called):

#### Repro code:

```csharp
public interface IX
{
    int Property { get; set; }
}

[Fact]
public void Stubbed_property_set_before_SetupGet()
{
    var mock = Mock.Get(Mock.Of<IX>());
    mock.Object.Value = 4;
    mock.SetupGet(m => m.Property).Returns(3);
    Assert.Equal(3, mock.Object.Value);
}

[Fact]
public void Stubbed_property_set_after_SetupGet()
{
    var mock = Mock.Get(Mock.Of<IX>());
    mock.SetupGet(m => m.Property).Returns(3);
    mock.Object.Value = 4;
    Assert.Equal(3, mock.Object.Value);
}
```

#### Expected behavior:

Both tests should pass, since one would expect that the `SetupGet` setup will be invoked.

#### Actual behavior:

The second test fails.

#### Likely cause:

#826 made stubbed properties fully lazy. (`Mock.Of<T>` traditionally calls `SetupAllProperties`, which stubs all properties.) Setups for stubbed properties are now only added once that property is first queried (or being assigned to). That's the reason why the `SetupGet` setup in the second test gets overridden/shadowed by the stubbed property getter.

#### Possible resolution:

If we classify the new behavior (where the second test fails since 4.12.0) as a regression or bug, we could only register stubbed properties' lazy accessor setups if there isn't already a pre-existing setup for that accessor.

/cc @bluechrism @lepijohnny

lepijohnny:
Hi @stakx,

the proposal makes sense. However, I have 2 questions:

1. Would the change be as simple as adding, and respective setter?
[...]

PR #1203, opened 2021-08-24
Enable multi-method setups

stakx:
This refactoring enables setups that handle more than just a single method.

I'm have two use cases in mind:

 1. `StubbedPropertyGetterSetup` and `StubbedPropertySetterSetup` could be merged into a single setup type. This would be more efficient than what we have now, and also better capture the semantics of a call to `mock.SetupProperty()`.

 2. It would allow creation of a new setup type `StubbedPropertiesSetup` backing `mock.SetupAllProperties()`. This would be added on `mock.SetupAllProperties()` and keep track of any number of properties' values. Again, this would be more efficient than the current solution, allow us to get rid of some convoluted internal bits (such as [`SetupCollection.RemoveAllPropertyAccessorSetups`](https://github.com/moq/moq4/blob/eacd7571c063fb458072289dfd40f708a8a258bc/src/Moq/SetupCollection.cs#L75-L93)), and it may help solve #1066.

### DIFF
--- a/src/Moq/Expectation.cs
+++ b/src/Moq/Expectation.cs
@@ -1,0 +1,45 @@
+// Copyright (c) 2007, Clarius Consulting, Manas Technology Solutions, InSTEDD, and Contributors.
+// All rights reserved. Licensed under the BSD 3-Clause License; see License.txt.
+
+using System;
+using System.Linq.Expressions;
+
+using Moq.Async;
+
+namespace Moq
+{
+	/// <summary>
+	///   Represents a set (or the "shape") of invocations
+	///   against which concrete <see cref="Invocation"/>s can be matched.
+	/// </summary>
+	internal abstract class Expectation : IEquatable<Expectation>
+	{
+		public abstract LambdaExpression Expression { get; }
+
+		public virtual bool HasResultExpression(out IAwaitableFactory awaitableFactory)
+		{
+			awaitableFactory = null;
+			return false;
+		}
+
+		public override bool Equals(object obj)
+		{
+			return obj is Expectation other && this.Equals(other);
+		}
+
+		public abstract bool Equals(Expectation other);
+
+		public abstract override int GetHashCode();
+
+		public abstract bool IsMatch(Invocation invocation);
+
+		public virtual void SetupEvaluatedSuccessfully(Invocation invocation)
+		{
+		}
+
+		public override string ToString()
+		{
+			return this.Expression.ToStringFixed();
+		}
+	}
+}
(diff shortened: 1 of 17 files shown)


src/Moq/Expectation.cs
```diff
@@ -0,0 +1,45 @@
...
+	/// </summary>
+	internal abstract class Expectation : IEquatable<Expectation>
+	{
+		public abstract LambdaExpression Expression { get; }
+
+		public virtual bool HasResultExpression(out IAwaitableFactory awaitableFactory)
+		{
+			awaitableFactory = null;
+			return false;
+		}
```
stakx, line 23: This method doesn't really fit in very well here, as it doesn't have much to do with the core purpose of `Expectation`.

A possibly better design might be to create a custom LINQ expression tree node type `ResultExpression` (which would carry the `IAwaitableFactory` as data), then add such a node directly to `Expression`.

That can be done as a separate refactoring another time.

src/Moq/Expectation.cs
```diff
@@ -0,0 +1,45 @@
...
+
+		public abstract bool Equals(Expectation other);
+
+		public abstract override int GetHashCode();
+
+		public abstract bool IsMatch(Invocation invocation);
+
+		public virtual void SetupEvaluatedSuccessfully(Invocation invocation)
+		{
+		}
```
stakx, line 38: This also doesn't fit in very well here.

I'm not yet sure how to get rid of it. A few possible options off the top off my head:

 * Let `Expectation` expose its argument matchers through a property. When a setup is successfully matched, `Setup` could then skip over `Expectation.SetupEvaluatedSuccessfully` and call the matchers' `SetupEvaluatedSuccessfully` methods directly.

 * Add an event `Setup.Matched`, to which argument matchers may optionally subscribe their `SetupEvaluatedSuccessfully` methods.

The first option seems like the more plausible in the short term. It doesn't make much sense to add a `Matched` event just to get rid of this method here, however that solution may start to look more plausible if the event is going to get added to also support other features (such as better support for sequences).

src/Moq/SetupCollection.cs
```diff
@@ -82,7 +82,7 @@ public void RemoveAllPropertyAccessorSetups()
 
 			lock (this.setups)
 			{
-				this.setups.RemoveAll(x => x.Method.IsPropertyAccessor());
+				this.setups.RemoveAll(s => s is MethodSetup ms && ms.Method.IsPropertyAccessor());
```
stakx, line 85: This is a bad change, but acceptable if it means we'll soon be able to remove this whole method.

aysan0:
@stakx, thanks for completing this. What version of MoQ will this change be in?

stakx:
@aysan0, this change was made mainly to pave the way for fixing #1066. For now, it remains strictly an implementation detail, and this will probably not be exposed in any public API anytime soon. (The likely moment for making this public would be when we allow user code to add fully custom `Behavior` setups... I had a WIP pull request about that a while ago but it wasn't ready yet.)

PR #1204, opened 2021-08-24
Combine `StubbedProperty{G|S}etterSetup`

stakx:
This is a follow-up to #1203.


src/Moq/Interception/InterceptionAspects.cs
```diff
@@ -186,63 +187,25 @@ public static bool Handle(Invocation invocation, Mock mock)
...
 
-				Setup getterSetup = null;
-				if (property.CanRead(out var getter))
-				{
-					if (ProxyFactory.Instance.IsMethodVisible(getter, out _))
-					{
-						propertyValue = CreateInitialPropertyValue(mock, getter);
-						getterSetup = new StubbedPropertyGetterSetup(mock, expression, getter, () => propertyValue);
-						mock.MutableSetups.Add(getterSetup);
-					}
```
stakx, line 211: Those visibility checks are no longer required because we already know that either the getter or the setter can be intercepted by the proxy factory; otherwise we wouldn't end up here in the first place.

If the other accessor isn't interceptable, that amounts to the same thing as if it were completely absent: we'll still create a setup just for the present accessor (even if it won't be of much actual use beyond preventing this code from executing repeatedly for the same property).

src/Moq/Mock`1.cs
```diff
@@ -633,12 +633,19 @@ public Mock<T> SetupProperty<TProperty>(Expression<Func<T, TProperty>> property,
...
-			Mock.SetupSet(this, property.AssignItIsAny(), condition: null).SetCallbackBehavior(new Action<TProperty>(p => value = p));
+			if (!pi.CanRead(out var getter))
+			{
+				Guard.CanRead(pi);
+			}
+
+			if (!pi.CanWrite(out var setter))
+			{
+				Guard.CanWrite(pi);
+			}
```
stakx, line 645: This isn't pretty. Seems like we're missing a method that acts as both guard and for accessor(s) retrieval (or perhaps even better, an exception factory method).

src/Moq/SetupCollection.cs
```diff
@@ -82,7 +82,7 @@ public void RemoveAllPropertyAccessorSetups()
 
 			lock (this.setups)
 			{
-				this.setups.RemoveAll(s => s is MethodSetup ms && ms.Method.IsPropertyAccessor());
+				this.setups.RemoveAll(s => s is StubbedPropertySetup || (s is MethodSetup ms && ms.Method.IsPropertyAccessor()));
```
stakx, line 85: This line of code is getting really bad. It's even more important now to get rid of the whole method, if possible; see https://github.com/moq/moq4/pull/1203#discussion_r694890829.

src/Moq/StubbedPropertySetup.cs
```diff
@@ -0,0 +1,113 @@
...
+	{
+		private object value;
+
+		public StubbedPropertySetup(Mock mock, LambdaExpression expression, MethodInfo getter, MethodInfo setter, object initialValue)
+			: base(originalExpression: null, mock, new PropertyAccessorExpectation(expression, getter, setter))
+		{
+			Debug.Assert(mock != null);
+			Debug.Assert(expression != null);
+			Debug.Assert(expression.IsProperty());
+			Debug.Assert(getter != null || setter != null);
```
stakx: Those assertions are redundant, since `PropertyAccessorExpectation` and the base class' constructor already assert the same things.

Issue #1083, opened 2020-10-27
Slower performance after upgrading from 4.2.1510 to 4.14.7


vruss:
I'm trying to troubleshoot why the test duration has increased by several seconds on a lot of tests (which amounts to double-digit minutes total). 

Synthetic tests show that the newer Moq version is a lot faster with both `new Mock` and `Mock.Of`. But real testcases are significantly slower. 

I think the issue is somewhat related to this [Castle issue](https://github.com/castleproject/Core/issues/486). As we've also updated Castle from 4.3.3 to 4.4.1, I'm not sure if the issue is with Castle or Moq. 

Are there any known performance issues where mocked interfaces (and classes) with A LOT of properties and methods with multiple layers of inheritance with even more properties have become slower in newer updates?
#
I'm sorry I can't provide a repro as it's production code and I've yet to successfully reproduce it myself in a contained unit.

stakx:
There's quite frankly nothing we can do to help you without some code we can run and benchmark, so I am going to have to close this issue... sorry. A few remarks though:

* Is the Moq update the only thing you did? Or did you perhaps update the .NET Framework or switch platform? That could have a massive influence on runtime performance; .NET Core 3 is much, much faster than the .NET Framework, for example, so you may want to consider switching.

* To the best of my knowledge, DynamicProxy 4.x versions have been fairly stable. I wouldn't expect huge performance differences there. I don't know about earlier versions (3.x).

* That being said, Moq performance is mostly dictated by the time spent in DynamicProxy, which spends most of its time in System.Reflection[.Emit]. For that reason, it pays off to be aware of what types you're mocking. The more different types you mock, and the more members those types have, the worse performance you'll get. Favor small role interfaces over complex / monolithlic base classes in your tests.

* Where possible, don't reuse mocks between tests. Let each tests have its separate mock(s), and set up only what's necessary. Mocks with few setups should tend to be faster than mocks with a ton of setups to satisfy the needs of many diverse tests. Separate mocks are also beneficial because because mocks use locking for thread-safety. If you share the same mocks between tests, and you run tests in parallel, the test runner threads might end up having to wait their turn.

* Moq versions 4.8+ should have runtime performance characteristics similar to early 4.0-4.2 versions. They're a little slower (due to many new features I'd guess) but _not_ by an order of magnitude or more (last time I checked, anyway). 4.12 may very well be the fastest; I've focused more on consistency and correctness of features in newer versions and that may have somewhat brought down speed once again.

[...]

PR #1227, opened 2022-01-24
 Ability to specify proxy generator or invalidate it #925 

Znurre:
This is a draft for the changes as per our discussion in issue #1083.

I am aware that it might need cleaning up, documentation, and so on. In this first stage, I mainly want feedback on this approach, if it's the right way to go at all.

I opted to make `ProxyFactory` public, which I assume is desirable if we want external projects to provide their own implementations. That in turn cascaded a bit, and required me to make some more types public.

In order to not leak too much implementation, I decided to expose the interface types instead of the abstract base classes. This might require some more boilerplate for alternative proxy factory implementations, but I would argue that it's worth it.

I can also confirm that with these rather crude changes, we are now able to plug in our own proxy generator, and by utilizing Cecil, we have managed to cut our test execution times nearly 2 hours down to 15 minutes.

### DIFF
--- a/src/Moq/ActionObserver.cs
+++ b/src/Moq/ActionObserver.cs
@@ -107,7 +107,7 @@ namespace Moq
 				}
 			}
 
-			Expression[] GetArgumentExpressions(Invocation invocation, Match[] matches)
+			Expression[] GetArgumentExpressions(IInvocation invocation, Match[] matches)
 			{
 				// First, let's pretend that all arguments are constant values:
 				var parameterTypes = invocation.Method.GetParameterTypes();
@@ -239,7 +239,7 @@ namespace Moq
 		{
 			private readonly MatcherObserver matcherObserver;
 			private int creationTimestamp;
-			private Invocation invocation;
+			private IInvocation invocation;
 			private int invocationTimestamp;
 			private object returnValue;
 
@@ -251,7 +251,7 @@ namespace Moq
 				this.creationTimestamp = this.matcherObserver.GetNextTimestamp();
 			}
 
-			public Invocation Invocation => this.invocation;
+			public IInvocation Invocation => this.invocation;
 
 			public IEnumerable<Match> Matches
 			{
@@ -264,7 +264,7 @@ namespace Moq
 
 			public Recorder Next => (Awaitable.TryGetResultRecursive(this.returnValue) as IProxy)?.Interceptor as Recorder;
 
-			public void Intercept(Invocation invocation)
+			public void Intercept(IInvocation invocation)
 			{
 				var returnType = invocation.Method.ReturnType;
 
--- a/src/Moq/Async/IAwaitableFactory.cs
+++ b/src/Moq/Async/IAwaitableFactory.cs
@@ -7,18 +7,32 @@ using System.Linq.Expressions;
 
 namespace Moq.Async
 {
-	internal interface IAwaitableFactory
+	/// <summary>
+	/// </summary>
+	public interface IAwaitableFactory
 	{
+		/// <summary>
+		/// </summary>
 		Type ResultType { get; }
 
+		/// <summary>
+		/// </summary>
 		object CreateCompleted(object result = null);
 
+		/// <summary>
+		/// </summary>
 		object CreateFaulted(Exception exception);
 
+		/// <summary>
+		/// </summary>
 		object CreateFaulted(IEnumerable<Exception> exceptions);
 
+		/// <summary>
+		/// </summary>
 		Expression CreateResultExpression(Expression awaitableExpression);
 
+		/// <summary>
+		/// </summary>
 		bool TryGetResult(object awaitable, out object result);
 	}
 }
--- a/src/Moq/AutoImplementedPropertyGetterSetup.cs
+++ b/src/Moq/AutoImplementedPropertyGetterSetup.cs
@@ -24,7 +24,7 @@ namespace Moq
 			this.MarkAsVerifiable();
 		}
 
-		protected override void ExecuteCore(Invocation invocation)
+		protected override void ExecuteCore(IInvocation invocation)
 		{
 			invocation.ReturnValue = this.getter.Invoke();
 		}
--- a/src/Moq/AutoImplementedPropertySetterSetup.cs
+++ b/src/Moq/AutoImplementedPropertySetterSetup.cs
@@ -23,7 +23,7 @@ namespace Moq
 			this.MarkAsVerifiable();
 		}
 
-		protected override void ExecuteCore(Invocation invocation)
+		protected override void ExecuteCore(IInvocation invocation)
 		{
 			this.setter.Invoke(invocation.Arguments[0]);
 		}
--- a/src/Moq/Behavior.cs
+++ b/src/Moq/Behavior.cs
@@ -9,6 +9,6 @@ namespace Moq
 		{
 		}
 
-		public abstract void Execute(Invocation invocation);
+		public abstract void Execute(IInvocation invocation);
 	}
 }
--- a/src/Moq/Behaviors/Callback.cs
+++ b/src/Moq/Behaviors/Callback.cs
@@ -17,7 +17,7 @@ namespace Moq.Behaviors
 			this.callback = callback;
 		}
 
-		public override void Execute(Invocation invocation)
+		public override void Execute(IInvocation invocation)
 		{
 			this.callback.Invoke(invocation);
 		}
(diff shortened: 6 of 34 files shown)


stakx:
Thanks for your interest in working on this!

I haven't finished looking through all of your PR, however my first impression is that this exposes far too many Moq internals that shouldn't be made public. Exposing those would needlessly tie our hands in refactoring Moq in the future.

The main problem in this PR is that `ProxyFactory` isn't ready to be exposed publicly in its current form. (I mentioned this in https://github.com/moq/moq4/issues/1083#issuecomment-718474095. Much of Moq v4's internals are specifically geared towards DynamicProxy, and if we want to allow completely different proxy generators, we first need to cleanly decouple Moq's internals from DynamicProxy's capabilities. This would very likely affect the `ProxyFactory`'s contract. If we expose `ProxyFactory` as it is now, you might implement that seemingly straightforward contract but then run into weird malfunctions because under the hood, Moq still expects the behaviour and restrictions from your new proxy generator that DynamicProxy exhibits.)

Many other parts of your PR appear to be a direct consequence of exposing `ProxyFactory`'s methods, which IMHO is a no-go, so I won't go into more detail at this point.

stakx:
> I can also confirm that with these rather crude changes, we are now able to plug in our own proxy generator, and by utilizing Cecil, we have managed to cut our test execution times nearly 2 hours down to 15 minutes.

This is promising. Would you be willing to share your Cecil-based implementation of `ProxyFactory`? This might give us some further hints at how we'd go about decoupling Moq internals from DynamicProxy specifics.

Znurre:
Thank you for your reply :)

Our Cecil implementation is still very messy and needs cleaning up, but I can at least share the code of our `CecilProxyFactory`:

```cs
internal class CecilProxyFactory : ProxyFactory
{
	private readonly ConcurrentDictionary<CacheKey, Type> proxyTypes;

	public CecilProxyFactory()
	{
		this.proxyTypes = new ConcurrentDictionary<CacheKey, Type>();

	}

	public override object CreateProxy(Type mockType, IInterceptor interceptor, Type[] interfaces, object[] arguments)
	{
		var type = this.GetProxyType(mockType, interfaces, arguments.Length);
		var argsWithInterceptor = new object[arguments.Length + 1];
		argsWithInterceptor[0] = interceptor;
		arguments.CopyTo(argsWithInterceptor, 1);
		var instance = Activator.CreateInstance(type, argsWithInterceptor);
		if(mockType.IsDelegateType())
		{
			return Delegate.CreateDelegate(mockType, instance, instance.GetType().GetMethod("Invoke"));
		}
		return instance;
	}

	private Type GetProxyType(Type mockType, Type[] interfaces, int length)
	{
		var key = CacheKey.Create(mockType, interfaces);
		return this.proxyTypes.GetOrAdd(key, key => this.CreateProxyType(mockType, interfaces));
	}


	private Type CreateProxyType(Type mockType, Type[] interfaces)
	{
		using (var builder = new AssemblyGenerator())
		{
			if (mockType.IsInterface)
			{
				var typeName = builder.DefineInterfaceProxy(mockType, interfaces);
				var type = builder.BuildAssembly().GetType(typeName);
				return type;
			}
			else if (mockType.IsDelegateType())
			{
				var typeName = builder.DefineDelegateContainer(mockType, interfaces);
				var type = builder.BuildAssembly().GetType(typeName);
				return type;
			}
			else
			{
				// Class proxy
				var typeName = builder.DefineClassProxy(mockType, interfaces);
				var type = builder.BuildAssembly().GetType(typeName);
				return type;
			}
		}
		throw new NotImplementedException();
	}

	public override bool IsMethodVisible(System.Reflection.MethodInfo method, out string messageIfNotVisible)
	{
		messageIfNotVisible = null;
		return true;
	}

	public override bool IsTypeVisible(Type type)
	{
		return true;
	}

	private sealed class CacheKey
	{
		private readonly MemberInfo[] orderedTypes;
		private readonly int hashCode;

		private CacheKey(params System.Reflection.MemberInfo[] orderedTypes)
		{
			this.orderedTypes = orderedTypes;
			this.hashCode = this.CalculateHashCode();
		}

		public override bool Equals(object obj)
		{
			var other = obj as CacheKey;
			if (other == null || this.orderedTypes.Length != other.orderedTypes.Length)
			{
				return false;
			}
			for (var i = 0; i < this.orderedTypes.Length; i++)
			{
				if (!this.orderedTypes[i].Equals(other.orderedTypes[i]))
				{
					return false;
				}
			}
			return true;
		}

		public override int GetHashCode()
		{
			return this.hashCode;
		}

		private int CalculateHashCode()
		{
			// TODO: Improve whenever
			var hash = this.orderedTypes.Length;
			for (var i = 0; i < this.orderedTypes.Length; i++)
			{
				hash = (hash * 7) ^ this.orderedTypes[i].GetHashCode();
			}
			return hash;
		}

		internal static CacheKey Create(Type t1, Type[] moreTypes)
		{
			var h = new HashSet<MemberInfo>() { t1 };
			h.UnionWith(moreTypes);
			var orderedTypes = h.OrderBy(x => x.Name, StringComparer.Ordinal).ToArray();
			return new CacheKey(orderedTypes);
		}
	}
}
```
As you can see, we are using the provided `IInterceptor` directly, which pretty much forces these changes, but we also have our own implementation of `IInvocation`, which pretty much only changes the behavior of `CallBase`.

```cs
using Moq.Async;
using System;
using System.Linq;
using System.Reflection;

namespace Moq
{
	/// <summary/>
	public class CecilInvocation : Moq.IInvocation
	{
		/// <summary>
		/// Internal type to mark invocation results as "exception occurred during execution". The type just
		/// wraps the Exception so a thrown exception can be distinguished from an <see cref="System.Exception"/>
		/// return value.
		/// </summary>
		private readonly struct ExceptionResult
		{
			public ExceptionResult(Exception exception)
			{
				Exception = exception;
			}

			public Exception Exception { get; }
		}

		private readonly object proxyInstance;

		private object result;

		private MethodInfo methodImplementation;

		/// <summary/>
		public CecilInvocation(Type proxyType
			, System.Reflection.MethodInfo method
			, object proxyInstance
			, params object[] arguments
			)
		{
			this.proxyInstance = proxyInstance;

			this.ProxyType = proxyType;
			this.Method = method;
			this.Arguments = arguments;
		}

		/// <summary>
		/// 
		/// </summary>
		public MethodInfo Method { get; }
		/// <summary>
		/// 
		/// </summary>
		public MethodInfo MethodImplementation
		{
			get
			{
				if (this.methodImplementation == null)
				{
					this.methodImplementation = this.Method.GetImplementingMethod(this.ProxyType);
				}

				return this.methodImplementation;
			}
		}
		/// <summary>
		/// 
		/// </summary>
		public object[] Arguments { get; }
		/// <summary>
		/// 
		/// </summary>
		public ISetup MatchingSetup { get; private set; }
		/// <summary>
		/// 
		/// </summary>
		public Type ProxyType { get; }
		/// <summary>
		/// 
		/// </summary>
		public bool IsVerified { get; }
		/// <summary>
		/// 
		/// </summary>
		public object ReturnValue
		{
			get => this.result is ExceptionResult ? null : this.result;
			set
			{
				this.result = value;
			}
		}

		/// <summary>
		/// 
		/// </summary>
		public Exception Exception
		{
			get => this.result is ExceptionResult r ? r.Exception : null;
			set
			{
				this.result = new ExceptionResult(value);
			}
		}

		/// <summary/>
		public object CallBase()
		{
			if (this.Method.IsAbstract)
			{
				throw new NotSupportedException();
			}
			// call the base method
			var ftn = this.Method.MethodHandle.GetFunctionPointer();
			var delegateType = this.GetDelegateType();
			var d =(Delegate)Activator.CreateInstance(delegateType, this.proxyInstance, ftn);
			var result = d.DynamicInvoke(this.Arguments);
			return result;
		}

		/// <summary>
		/// 
		/// </summary>
		/// <param name="awaitableFactory"></param>
		public void ConvertResultToAwaitable(IAwaitableFactory awaitableFactory)
		{
			if (this.result is ExceptionResult r)
			{
				this.result = awaitableFactory.CreateFaulted(r.Exception);
			}
			else if (this.result != null && !this.Method.ReturnType.IsAssignableFrom(this.result.GetType()))
			{
				this.result = awaitableFactory.CreateCompleted(this.result);
			}
		}

		/// <summary>
		/// 
		/// </summary>
		/// <param name="setup"></param>
		public void MarkAsMatchedBy(ISetup setup)
		{
			this.MatchingSetup = setup;
		}

		/// <summary>
		/// 
		/// </summary>
		/// <returns></returns>
		private Type GetDelegateType()
		{
			var types = this.Method.GetParameters()
				.Select(p => p.ParameterType)
				.Concat(new[] { this.Method.ReturnType });
			return Expression.GetDelegateType(types.ToArray());
		}
	}
}
```
Perhaps it could be possible to create another abstraction on top of `IInterceptor` and `IInvocation` instead?

PR #1234, opened 2022-02-13
Create dedicated setup type for `SetupAllProperties`

### LINKED ISSUE #1066: Property setups are ignored on mocks instantiated using `Mock.Of`

### LINKED ISSUE #1231: SetupAllProperties causes mocks to become race-prone
Moq version: 4.16.1
Target Framework: Probably any, tested on net48 and net6.0

A test to reproduce the problem:
```
using System.Threading;
using System.Threading.Tasks;
using Moq;
using NUnit.Framework;

namespace NUnitTestProject2
{
  public interface IEntity
  {
    int Property { get; }
  }

  public class Tests
  {
    [Test]
    [Repeat(10000)]
    public void Using_SetupAllProperties_should_not_break_thread_safety()
    {
      TestContext.Progress.WriteLine("========== Test Started ===========");

      var barrier = new AutoResetEvent(false);
      var gate = new AutoResetEvent(false);
      var mock = new Mock<IEntity>();

      // Commenting the following line makes the test pass
      mock.SetupAllProperties();

      var task = Task.Factory.StartNew(() =>
      {
        barrier.Set();

        while (!gate.WaitOne(0))
        {
          TestContext.Progress.WriteLine($"Property = {mock.Object.Property}");
        }
      });

      barrier.WaitOne();
      gate.Set();

      TestContext.Progress.WriteLine("Changing Property to 5");
      mock.Setup(x => x.Property).Returns(5);
      TestContext.Progress.WriteLine("Changed Property to 5");

      Assert.That(mock.Object.Property, Is.EqualTo(5), () => "Property is not equal to 5");

      task.Wait();

      TestContext.Progress.WriteLine("========== Test Finished ===========");
    }
  }
}
```

stakx:
This is a continuation of #1204, where a dedicated setup type was created for single stubbed properties' accessors (i.e. `SetupProperty`). Thanks to #1203, the same can also be done for `SetupAllProperties`.

A few additional notes:

* This gets rid of the `RemoveAllPropertyAccessorSetups` method in `SetupCollection`, restoring it to an add-only collection, which in turn means its use of `lock`-ing could be optimized in a similar fashion as was done for `InvocationCollection`.

* Gets rid of the `AutoSetupPropertiesDefaultValueProvider` hack and simplifies the interception pipeline.

* Moq should become a little more efficient in principle, since there will be fewer setups created overall.

Fixes #1066, likely fixes #1231.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## Unreleased
+
+#### Fixed
+
+* Property setups are ignored on mocks instantiated using `Mock.Of` (@stakx, #1066)
+* `SetupAllProperties` causes mocks to become race-prone (@estrizhok, #1231)
+
+
 ## 4.17.0 (2022-02-13)
 
 #### Added
--- a/src/Moq/AsInterface.cs
+++ b/src/Moq/AsInterface.cs
@@ -43,12 +43,6 @@ namespace Moq
 			set => this.owner.DefaultValueProvider = value;
 		}
 
-		internal override DefaultValueProvider AutoSetupPropertiesDefaultValueProvider
-		{
-			get => this.owner.AutoSetupPropertiesDefaultValueProvider;
-			set => this.owner.AutoSetupPropertiesDefaultValueProvider = value;
-		}
-
 		internal override EventHandlerCollection EventHandlers => this.owner.EventHandlers;
 
 		internal override Type[] InheritedInterfaces => this.owner.InheritedInterfaces;
--- a/src/Moq/Interception/InterceptionAspects.cs
+++ b/src/Moq/Interception/InterceptionAspects.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Linq.Expressions;
 using System.Reflection;
 
 using Moq.Behaviors;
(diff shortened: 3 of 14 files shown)


mo-russo:
Hello @stakx, since we got version 4.18.1, our mocks' subclasses that call SetupAllProperties are crashing.
```c#
public class MyInterfaceMock : Mock<IMyInterface>
{
  public MyInterfaceMock ()
  {
    SetupAllProperties(); // Crashes here
  }
}
```

This isn't allowed anymore?

stakx:
@mo-russo, this is a regression. Could you please report the above again in a new issue, so that we can properly track it? Thanks.

Issue #962, opened 2019-11-27
How do you mock an IAsyncEnumerable<T> ?


TechWatching:
I want to setup a method that returns an IAsyncEnumerable<MyObject> but I don't understand how to do it.

poke:
(This was cross-posted to [Stack Overflow](https://stackoverflow.com/questions/59071994/how-do-you-mock-and-iasyncenumerable))

TechWatching:
I will close the issue as the Stack Overflow post answered the question, but I guess Moq should allow to use `RetunsAsync` instead of `Return` to setup methods retunring and `IASyncEnumerable`

PR #1250, opened 2022-03-30
IAsyncEnumerable ReturnsAsync support

cdonnellytx:
This adds `ReturnsAsync` overloads that can return `IAsyncEnumerable<T>` based on an input `IEnumerable<T>` value.

This was first requested in #962.

## Additional information

- Because IAsyncEnumerable is not supported in .NET 4.5, but is supported in .NET Framework 4.6.1 and later, Moq now additionally targets .NET 4.6.1 so that Framework users may also benefit from this feature.
- [System.Linq.Async](https://www.nuget.org/packages/System.Linq.Async) has also been added as an additional dependency.  
    - This could be reduced to just [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) but would need an implementation of the [ToAsyncEnumerable()](https://github.com/dotnet/reactive/blob/9f2a8090cea4bf931d4ac3ad071f4df147f4df50/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.cs) extension method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+#### Added
+
+* `IAsyncEnumerable` `ReturnsAsync` support (@cdonnellytx, #1250)
 
 ## 4.17.2 (2022-03-06)
 
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -134,12 +134,7 @@
 
 	<Target Name="Test" />
 
-	<Target Name="UpdatePackageMetadata"
-          BeforeTargets="PrepareForBuild;GetAssemblyVersion;GenerateNuspec;Pack"
-          DependsOnTargets="InitializeSourceControlInformation"
-          Condition="'$(SourceControlInformationFeatureSupported)' And
-                     '$(RepositoryUrl)' != '' And 
-                     '$(SourceRevisionId)' != ''">
+	<Target Name="UpdatePackageMetadata" BeforeTargets="PrepareForBuild;GetAssemblyVersion;GenerateNuspec;Pack" DependsOnTargets="InitializeSourceControlInformation" Condition="'$(SourceControlInformationFeatureSupported)' And&#xD;&#xA;                     '$(RepositoryUrl)' != '' And &#xD;&#xA;                     '$(SourceRevisionId)' != ''">
 		<PropertyGroup>
 			<Description>$(Description)
 
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -6,7 +6,7 @@
 	<Import Project="$(BuildDirectory)SourceLink.props" />
 
 	<PropertyGroup>
-		<TargetFrameworks>net45;netstandard2.0;netstandard2.1</TargetFrameworks>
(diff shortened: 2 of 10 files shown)


stakx:
Hi @cdonnellytx and thanks for contributing. Your PR is touching a few things that I was hoping to leave behind in the past, so I am not convinced this should get merged:

* I don't think adding new `netXY` target frameworks is the right way to go (I'd rather we keep them to a minimum) as the .NET Framework is now legacy tech.

* Even more so if it means adding conditional compilation symbols, too. Ideally we wouldn't have _any_ of those. I once made an effort to get rid of as many of them as I could, as they can make code harder to reason about.

* The `.ReturnsAsync`, `.ThrowsAsync` APIs have become mostly redundant, since you can now setup a `[Value]Task`'s `.Result` directly and just use `.Returns`, `.Throws`. I was hoping we could let those older APIs die off, since (as you discovered) they add a ton of repetitive method overloads.

Your PR's main goal (glossing over the time delay overloads) appears to be saving users from having to call `.ToAsyncEnumerable()` manually. Is that assessment correct?

cdonnellytx:
> Your PR's main goal (glossing over the time delay overloads) appears to be saving users from having to call `.ToAsyncEnumerable()` manually. Is that assessment correct?

Yes, that is correct.  I am hoping to avoid having to do that, especially since `[Value]Task` have the equivalent `ReturnsAsync` methods.  Though if those are going away, it doesn't makes sense to add them (back) for IAsyncEnumerable.
​
One difference though is that IAsyncEnumerable doesn't have a `.Result` or any other equivalent property; the nearest thing I could find is a [ToEnumerable()](https://fuqua.io/Rx.NET/ix-docs/html/M_System_Linq_AsyncEnumerable_ToEnumerable__1.htm) extension method:
​
```diff
- service.Setup(m => m.GetRelatedIdsAsync(1)).ReturnsAsync(new[] { 1, 2, 3 });
+ service.Setup(m => m.GetRelatedIdsAsync(1).ToEnumerable()).Returns(new[] { 1, 2, 3 });
```
​
Unfortunately, that doesn't work out-of-box today; Moq would have to be updated to support that expression -- and I'm not familiar enough with expression trees to know if that would work.
​
Regarding target frameworks, I went with `net461` because that is the [minimum supported target](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/1.0.0#supportedframeworks-tab) for IAsyncEnumerable, which meant adding a target to support it in later .NET Framework versions.  The alternative would be to raise Moq's minimum .NET Framework target to 4.6.1 (or possibly 4.6.2 since that will be the new minimum supported version after [April 26](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/ ".NET Framework 4.5.2, 4.6, 4.6.1 will reach End of Support on April 26, 2022")).
​
Regarding time delay overloads, are you planning on supporting some kind of delay testing in the next version of Moq?

stakx:
Finally found time to take a closer look at this. Sorry to say but I'm going to close this without merging.

* The `ReturnsAsync` methods produce `Task`s or `ValueTask`s &ndash; I don't think we need to have a `ReturnsAsync` method for `IAsyncEnumerable<>` (which is *not* a task, and thus doesn't really fit the existing pattern) just because there's "async" in the interface's name.
* Doing a `setup.Returns(enumerable.ToAsyncEnumerable())` isn't really _that_ much more complicated than `setup.ReturnsAsync(enumerable)`. I don't think we need to add a convenience helper method to the core library (that we'll then have to maintain) just so users can avoid calling that extension method themselves.

Issue #1244, opened 2022-03-17
moqthis.com is down


PorgJuice:
https://moqthis.com fails to load

![moqthis](https://user-images.githubusercontent.com/6478756/158832227-ad7c3578-61e7-4a26-9247-16491cd3707b.png)

stakx:
Hi @ProfessorSalty, thanks for reporting. Forwarding this to @kzu.

SeanKilleen:
When using `http` instead of `https`, it appears to forward to a parked Namecheap page and says it was recently registered: 

> ![image](https://user-images.githubusercontent.com/2148318/159701818-85b4b4dc-2657-4e51-a5de-e3c6ea48fcfc.png)

I hope that's intentional and that the domain wasn't lost.

If I can support in any way, let me know!

stakx:
Looks like https://moqthis.com is back online... closing.

mjstrom:
Both https://moqthis.com/ and http://moqthis.com/ are still serving landing pages like the ones in the screenshots earlier in this issue.   I'm assuming that isn't intentional and the domain was lost?

kzu:
Yeah, the domain has been entirely pointless for years, with no value added for the project. I let it expire. (not entirely intentionally, but I don't think it was really necessary anyway). We can live with http://moq.github.io/moq4/ I think.

mjstrom:
Thanks for the quick response.  Just a suggestion, you might want to update the places that still refer to the old site as it is confusing for someone (like myself) who was new to moq and trying to learn more.  I saw the link in at least two places -

- The `about` info section at the top right of the github page `https://github.com/moq/moq4` under the text "Repo for managing Moq 4.x"
- In the main README.md used for the git hub page `https://github.com/moq/moq4#where` in the last sentence in the link "documentation website".

stakx:
Agreed with both of you. Let's go ahead and remove all references to that site since it's now gone for good.

stakx:
@kzu could you please take care of the GitHub repo homepage link for us? (I don't have the necessary permissions to change / remove it myself in the repo's settings.)

SeanKilleen:
✅ Submitted the PR for the markdown
✅ Updated the wiki 

Once the link on the repo homepage is complete, this can likely be closed. 👍

kzu:
Done!

PR #1261, opened 2022-05-12
Update link to documentation website

SeanKilleen:
It is now officially at http://moq.github.io/moq4/

Supports #1244.


stakx (approved):

kzu:
should be https, but it will redirect automatically nevertheless

stakx:
> should be https

Good catch, thanks for mentioning it. I'll fix that.

PR #1274, opened 2022-08-02
add STTE for older target frameworks only

tibel:
Add System.Threading.Tasks.Extensions reference conditionally to reduce dependencies on newer frameworks.


stakx (approved):
Ooh, I like this! We've had so many versioning issues caused by this package in the past, I'm glad to see this go (at least for the newer TFMs). Thanks!

stakx:
@tibel, FYI, I've just pushed Moq version 4.18.2 to NuGet, which includes this change.

Issue #1210, opened 2021-09-21
Add Times Expectation to the mock setup


surovij45:
```csharp
mock
.Setup(foo => foo.Add("my string")
.Returns(true)
.Times(Times.Once)
```

Add `Times` expectation to the mock setup, and we can verify setup using single line `mock.VerifyAll()` or `mockRepo.VerifyAll()`

surovij45:
or 

```csharp
mock
.Setup(foo => foo.Add("my string"), Times.Once)
.Returns(true)
```

stakx:
I'll have to look up the issue where we discussed this topic last, but IIRC we ended up with `setup.Verifiable(times, [failMessage])` back then. `setup.Times(times)` also seems reasonable. `mock.Setup(..., times)` wouldn't be so great because we'd need to add a ton of method overloads to an already heavily overloaded method.

surovij45:
Implementation is up to you

Danielku15:
@stakx I am currently trying to migrate away from RhinoMocks and our codebase is heavily relying on the repeat functionality offered on setups. I have read through the past discussions [here](https://github.com/moq/moq4/issues/530) and [here](https://github.com/moq/moq4/issues/373) and all topics stayed unresolved due to alternatives. 

I wanted to share my bits on what we would actually need and maybe there is still a chance to get some solution to our issue: 

1. We have 606 setups with `Repeat.Once`
2. We have 30 setups with `Repeat.Any` + we have a thousands setups without any `Repeat` specified where it is expected that we allow any time of calles (exact number is hard to find out).
3. We have 26 setups with `Repeat.Twice` (we can ignore this, it would be fine to change on this one to some SetupSequence)
4. We have 18 setups with `Repeat.Never`
5 We have 5 setups with `Repeat.AsLeastOnce`

We are verifying our overall Mocks with `.VerifyAll()`. So our actual needs are typically: 

1. We want to allow a setup to only be called once. Any further attempts to use it should result in an error. Preferrably this error should already ocurr on the usage of the setup and not the verification only. It would lead to strange side effects in our tests if it continues to run without an exception raised due to unexpected logic. 

[...]

PR #1319, opened 2022-12-30
Add `setup.Verifiable(Times times, [string failMessage])` method

### LINKED ISSUE #373: Set Times expectation on Setup

stakx:
Better late than never. 😁

### Usage summary:

Specifying a maximum number of allowed calls:

```csharp
var bomb = new Mock<IBomb>();
bomb.Setup(b => b.Poke()).Verifiable(Times.AtMostOnce);
bomb.Object.Poke();
bomb.Object.Poke();  // throws `MockException` right away, without an explicit `mock.Verify()`
```

(This is mostly equivalent to the deprecated `setup.AtMostOnce()` and `setup.AtMost(n)` methods, except those didn't mark a setup as verifiable.)

Using `mock.Verify[All]()` to check whether the setup was called the correct number of times:

```csharp
var bomb = new Mock<IBomb>();
bomb.Setup(b => b.Poke()).Verifiable(Times.Never);
bomb.Verify();  // passes
```

Excluding a setup from `mock.VerifyAll()` using (e.g.) `Times.AtMost(int.MaxValue)`:

```csharp
var anyNumberOfTimes = Times.AtMost(int.MaxValue);
var bomb = new Mock<IBomb>();
bomb.Setup(b => b.Poke()).Verifiable(anyNumberOfTimes);
bomb.Object.Poke();
bomb.VerifyAll();  // passes, even if the above line of code were removed
```

(It would be nicer to have something like `Times.Any`, however that wouldn't make sense in most scenarios and might confuse users into doing strange things, so I chose not to add that to the library for now.)

Verifying the number of calls in situations where invocation arguments are modified after the invocation (but before verification) happened:

```diff
 var product = new Product();
 var inventory = new Mock<IInventory>();
+inventory.Setup(a => a.Add(It.Is<Product>(p => p.Type == "Flower"))).Verifiable(Times.Once);

 product.Type = "Flower";
 inventory.Object.Add(product);
 product.Type = "Bomb";

-inventory.Verify(i => i.Add(It.Is<Product>(p => p.Type == "Flower")), Times.Once);  // fails because product is now a bomb
+inventory.Verify();  // works because the setup triggered when product was still a flower
```

Closes #373, #937, #1210.


rishisolutionsbysms:
When this is going to released?

stakx:
I'd love to do a release but I'm currently blocked from doing so by #1340.

andy1547:
This feature has been monumental in allowing us to migrate from NMock to Moq without rewriting 3000+ tests to the AAA style. @stakx We'd be incredibly grateful if you could create a new release :')

stakx:
@andy1547, sorry for not replying earlier, I've been somewhat busy with my daytime life lately. Fortunately, @kzu has in the meantime cleared the road-blocks that have prevented us from doing releases.

Issue #1341, opened 2023-05-09
IProtectedMock fails on a new virtual methods


VladimirKhvostov:
This regression was introduced in v4.7.0 (tested with 4.5.30 - no issues).

Consider the following example:
```CSharp
namespace LibraryTests
{
    using System;
    using System.Collections.Generic;
    using Microsoft.VisualStudio.TestTools.UnitTesting;
    using Moq;
    using Moq.Protected;

    public class DataReader
    {
        protected virtual List<string> ExecuteReader() => throw new NotImplementedException();
    }

    public class AdvancedReader : DataReader
    {
        protected new virtual IList<string> ExecuteReader() => base.ExecuteReader();
    }

    [TestClass]
    public class ProtectedTests
    {
        [TestMethod]
        public void Test1()
        {
            Mock<AdvancedReader> mock = new Mock<AdvancedReader>();
            mock.Protected().Setup<IList<string>>("ExecuteReader").Returns(new List<string>() { "1", "2", "3" });
        }
    }
}
```
Setup fails with the following error:
`System.InvalidOperationException: 'Sequence contains more than one matching element'`
Stack:
```
System.Core.dll!System.Linq.Enumerable.SingleOrDefault<System.Reflection.MethodInfo>(System.Collections.Generic.IEnumerable<System.Reflection.MethodInfo> source, System.Func<System.Reflection.MethodInfo, bool> predicate)	Unknown
Moq.dll!Moq.Protected.ProtectedMock<LibraryTests.AdvancedReader>.InternalSetup<System.Collections.Generic.IList<string>>(string methodName, System.Type[] genericTypeArguments, bool exactParameterMatch, object[] args)	Unknown
Moq.dll!Moq.Protected.ProtectedMock<LibraryTests.AdvancedReader>.Setup<System.Collections.Generic.IList<string>>(string methodName, object[] args)	Unknown
```

Since ExecuteReader in the AdvancedReader class is a **new** virtual method, I believe reflection returned 2 MethodInfo objects and Moq expects one or zero.

[![Back this issue](https://raw.githubusercontent.com/devlooped/SponsorLink/main/docs/assets/img/separator.png "Back this issue")](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)
[![Back this issue](https://img.shields.io/badge/backed-%240-EA4AAA?logo=githubsponsors "Back this issue")](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)

PR #1342, opened 2023-05-10
mock.Protected().Setup<int>("Foo") fails base implementation of Foo is hidden in the derived class

VladimirKhvostov:
This PR fixes an issue described in the #1341


kzu:
Thanks for the PR! Could you rebase on top of main, since files moved around?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is loosely based on [Keep a Changelog](http://keepachangelog.com/en/1
 
 ## Unreleased
 
+#### Fixed
+
+* mock.Protected().Setup<int>("Foo") fails if base implementation of the method is hidden in the derived class (#1341)
+
```
stakx, line 13: The changelog is currently out of sync with the latest releases. (Whats's currently titled "Unreleased" became the last 4.18.* release, and the 4.20.* releases are completely missing.) It should first be updated (in a separate PR) so that your changelog entry will end up with the correct release instead of with the 4.18.* changes that were already published.

stakx (commented):
I haven't reviewed & tested this in detail TBH, but it generally looks good to me.

The changelog should be first updated with the latest few missing releases though, otherwise the changelog entry will end up with an incorrect (already published) version.

CHANGELOG.md
kzu: Should be up to date now.

PR #1356, opened 2023-07-31
Add Torutek as a sponsor

torutek-gh:
This is the project we'd like our sponsorship to go towards. Thank you for moq!
I couldn't get the sponsor image to work using the existing format, so I found another one.


kzu:
Not to worry, sponsorships are refreshed automatically, and you'll show up on every future release of not only Moq but all packages released by [devlooped](https://www.nuget.org/profiles/Devlooped). Thanks for your support!

PR #1357, opened 2023-08-02
Added `readonly` modifier to private fields which are never changed

Lehonti:
In order to better keep track of what is changed and not

### DIFF
--- a/src/Moq.Tests/Moq.Tests.csproj
+++ b/src/Moq.Tests/Moq.Tests.csproj
@@ -47,7 +47,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" $([System.Text.RegularExpressions.Regex]::IsMatch('$(DefineConstants)', 'FEATURE_EF'))">
-		<PackageReference Include="EntityFramework" Version="6.2.0" />
+		<PackageReference Include="EntityFramework" Version="6.4.4" />
 	</ItemGroup>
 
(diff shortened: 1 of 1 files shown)


stakx:
@Lehonti, you closed this already yourself so there's no need to do anything anytime soon, but just FYI, there was actually a time when I intentionally decided against marking some fields as `readonly` as that actually comes with a small runtime cost. These days I realise that omitting `readonly` was a micro-optimization that probably noone is ever going to notice. :-D

While I believe that marking constant fields as `readonly` is a good idea, there's a ton of other things that should be cleaned up also, such as: enabling C# nullability checking and annotating types with `?` accordingly; then getting rid of some related `Debug.Assert`s that check for `null`; converting some properties from `{ get; set; }` to `{ get; init; }`; etc.

"Modernizing" the code base that way would be quite an undertaking and I'm not yet sure which would be the best way to start it.

Lehonti:
@stakx thanks for replying! I usually start by doing small changes like these in order to see which projects are active and open to contributions.

To see how I modernize codebases little by little, take a look at this project's pull requests (both open and merged). The last few dozen are mine:

https://github.com/PintaProject/Pinta/pulls

... And tell me if there's some idea you like :) 

I like to work with clean code, so when I'm going to work on something I clean up a little. But if I'm just getting acquainted with the project, it takes time to decide what to work on, so I just look for certain patterns in the code and clean, and clean, and clean... a few non-conflicting pull requests at a time. 

I sometimes do refactorings of the same type "in bulk" so that they are easy to code and review.

PR #1358, opened 2023-08-02
Revamp structure, apply oss template, cleanup projects/imports

kzu:
This also switches to NuGetizer for packaging, standard @devlooped .editorconfig, and substantially reduced custom imports since we leverage the [devlooped/oss](https://github.com/devlooped/oss) template repo and shared assets.


kzu:
@stakx this moving around might cause some annoyance for pending PRs, but I tried to keep the rename clean so hopefully git rebase will figure things out

PR #1363, opened 2023-08-04
Add 💜 SponsorLink support

kzu:
See https://www.cazzulino.com/sponsorlink.html and https://github.com/devlooped/SponsorLink

Let's thank everyone who supports the project 💜

### DIFF
--- a/Moq.sln
+++ b/Moq.sln
@@ -10,6 +10,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		CONTRIBUTING.md = CONTRIBUTING.md
 		src\Directory.props = src\Directory.props
 		License.txt = License.txt
+		.github\workflows\publish.yml = .github\workflows\publish.yml
 		readme.md = readme.md
 	EndProjectSection
 EndProject
@@ -20,6 +21,8 @@ EndProject
 Project("{778DAE3C-4631-46EA-AA77-85C1314464D9}") = "Moq.Tests.VisualBasic", "src\Moq.Tests.VisualBasic\Moq.Tests.VisualBasic.vbproj", "{840A8B2E-3D4B-4521-A61A-0291562CDC8B}"
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Moq.Tests.FSharpTypes", "src\Moq.Tests.FSharpTypes\Moq.Tests.FSharpTypes.fsproj", "{2D9EE4E0-8433-4F9C-A330-C4D74B956E0B}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Moq.CodeAnalysis", "src\Moq.CodeAnalysis\Moq.CodeAnalysis.csproj", "{FB06ED30-2AFD-44AF-A378-99C8B13BF29D}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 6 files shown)


baynezy:
Now I have to replace Moq everywhere 😭😭

jamierobson:
This is the build that breaks trust, knowing that data is harvested without consent, and that this as the intent is established to execute code on the developers machine. Even if this change were reverted, it is clear that the developers have the willingness to take this step, and therefore trust is eroded.

My team will be moving away from Moq due to this, and I'd recommend others consider doing the same. NSubstitute is a strong alternitive, for those who need some ideas.

sja-schleupen:
This is a serious violation of GDPR - you should remove it

xiaoxiao921:
I don't see how it violate gdpr.
The big problem though is that garbage creates a compiler warning (warning as error flag and you can't suddenly build your project 🤣) and for some reason it also stop builds for a certain amount of seconds. The creator really didnt think that through, that alone will make a lot of people switch out.

CasHil:
> I don't see how it violate gdpr. The big problem though is that garbage creates a compiler warning (warning as error flag and you can't suddenly build your project 🤣) and for some reason it also stop builds for a certain amount of seconds. The creator really didnt think that through, that alone will make a lot of people switch out.

If you'd like to see how it violates GDPR, look at the discussion in https://github.com/moq/moq/issues/1372.

aradalvand:
"Add 💜 SponsorLink support"

Famous last words

msedi:
Yes, this is a big problem. There are clear names in our e-mails and I do not accept that the e-mails of our company are made visible.

Also if this is communicated here, sometimes people do a simple nuget update with being informed that something this severe happens.

dabaus:
> I don't see how it violate gdpr.
> The big problem though is that garbage creates a compiler warning (warning as error flag and you can't suddenly build your project 🤣) and for some reason it also stop builds for a certain amount of seconds. The creator really didnt think that through, that alone will make a lot of people switch out.

The problem with storing a hash of the email is that someone else can hash the users email, compare it to the hash in sponsor link and determine that it is the same person. Thus, the process is reversible and this is personal data, which means you need consent from the user.

Write:
Like, the audacity of saying thank you while doing that, lmao.

Misza13:
Nice library suicide

kzu:
So @dabaus if SponsorLink switches to a GUID generated when you install the GH app, which you then write in some file somewhere, everyone would be happy?

KeterSCP:
@kzu everyone would be happy if library they use will not send **any** data to untrusted third-party servers without any consent, moreover if this will also slow down build process intentionally

Misza13:
> @kzu everyone would be happy if library they use will not send **any** data to untrusted third-party servers without any consent, moreover if this will also slow down build process intentionally

On top of that, some corporate build servers are set up in such a way that they don't have internet access and rely on on-prem mirrors of NuGet, npm etc.

Write:
> So @dabaus if SponsorLink switches to a GUID generated when you install the GH app, which you then write in some file somewhere, everyone would be happy?

Dude your library is gone, forget about it.
No one will keep using it. You're not trustable. You did the dumbest thing, quite literally. 

I'm sorry to say that.

dabaus:
> So @dabaus if SponsorLink switches to a GUID generated when you install the GH app, which you then write in some file somewhere, everyone would be happy?

I am no GDPR expert, but generally speaking, any id (like a username, customer id, github account id, ip address) that you can use together with other information in order to identify a person is to be regarded as personal data. It does not matter if it's only you who have access to that information, it's still personal data.

So i think the only reasonable way to build your service would be to do it through a web-app where users need to manually sign up and opt-in. Adding what could be regarded as spyware to your users dev-environments is obviously, even if it is legal, not going to be acceptable.

AgentBlackout:
> So @dabaus if SponsorLink switches to a GUID generated when you install the GH app, which you then write in some file somewhere, everyone would be happy?

You have just torpedoed your entire reputation and probably fair damage to your future. I think the closest thing to a recovery would be an immediate revert and grovelling apology rather than trying to double down.

timothyeckert:
> 

No I believe removing the data farming you snuck in would be the best path forward, followed by a very large apology.

rwsp:
no coming back from this, trust permanently eroded.

TopSwagCode:
I know I am going be to downvoted for this.
But I think this was an honest mistake. I have used MOQ for many years and loved it. I don't like this has been added and would agree to this shouldn't have happened. I doubt there was any malicious intend. Try giving his blog post a read here: https://www.cazzulino.com/sponsorlink.html

Screenshot of sponsorlink in action.

VS-SL02

So the good. It's for telling people about sponsorships and they are needed. The bad slowing down builds costing companies money in build time in their CI / CD env.

This is a big topic currently how opensource developers and projects should earn money for the work they do. And I am all in for it. Which brings us back to why I think this was an honest mistake. It was an attempt to bring focus to the sponsorship side of things. Sponsorship is a young project trying to do the right thing, but in a wrong way. I am sure @kzu has learned a lot from this. I think the entire community should take a chill pill and try to understand in @kzu / MOQ's perspective.

DISCLAIMER: I have nothing to do with MOQ and I don't know @kzu - I know that we are all human and we all make mistakes. For me MOQ is not dead. There was a mistake. Killing a project over 1 mistake is insane. It was good it was caught and brought up. But it's also a reminder to check your dependencies and have automated tools to do so.

psimsa:
@TopSwagCode +1. Anyone who has never done anything stupid in their life raise hand. And reading the docs around it, it certainly doesn't seem too bad. From my perspective, 1) It should have been a major version bump bcz by default IDEs will often happily install latest minor versions while you could put it in license you have to accept in IDE when installing major version, 2) the lib itself should be open source - it's about transparency, 3) as others have pointed out, hash itself is not good if it isn't salted - it's still PII (which would be clear if open sourced) and 4) clear info and opt-out option. Like this for example: 
```
Telemetry
---------
The .NET tools collect usage data in order to help us improve your experience. The data is collected by Microsoft and shared with the community. You can opt-out of telemetry by setting the DOTNET_CLI_TELEMETRY_OPTOUT environment variable to '1' or 'true' using your favorite shell.

Read more about .NET CLI Tools telemetry: https://aka.ms/dotnet-cli-telemetry
```

remerle:
This was a terrible decision, but I understand the desire to try to monetize an OSS project. However, there are tons of OSS projects that are able to monetize without stealing data.

One way is to spin off a premium version with features that people want enough to pay for. I wouldn't expect monthly payments - nobody wants to "subscribe" to a mocking library. 

Or, charge for support. Or any of a myriad of other ways that don't export PII to an untrusted source.

Time to fork this repo, I guess.

bastetfurry:
Where is the fork?

bastetfurry:
> Time to fork this repo, I guess.

Seems like it.

PR #1375, opened 2023-08-09
Remove SponsorLink since it breaks MacOS restore

### LINKED ISSUE #1371: Upgrading to version 4.20.1 breaks the build
Hi there

I just upgraded our solution to the latest version of Moq (v. 4.20.1), and now all of our test projects fail to build on my Mac. I get this exception:

`Analyzer 'Moq.SponsorLinker' threw an exception of type 'System.UnauthorizedAccessException' with message 'Access to the path '/usr/local/share/dotnet/sdk/7.0.302/Roslyn/bincore/%TEMP%\1M5Ot' is denied.'.`

Looks like this was made for Windows, as `%TEMP%` looks like a Windows environment variable, which does not work on macOS.

Have to downgrade to 4.18.4 to get the build working again.

Hope you can look into this 🙂

kzu:
Fixes #1371

### DIFF
--- a/src/Moq/Moq.csproj
+++ b/src/Moq/Moq.csproj
@@ -110,8 +110,4 @@
     <InternalsVisibleTo Include="DynamicProxyGenAssembly2" PublicKey="0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7" />
   </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\Moq.CodeAnalysis\Moq.CodeAnalysis.csproj" ReferenceOutputAssembly="false" />
-  </ItemGroup>
-
 </Project>


PoteRii:
Hello, should we expect that SponsorLink is back after MacOS restore is fixed?

jakoss:
> Hello, should we expect that SponsorLink is back after MacOS restore is fixed?

It does not contain removal of the project, just the reference in the main csproj. So probably it will be back

LucHeart:
"This is fine" - so you are only removing it since it breaks MacOS?... Really?

How can you ignore all the voices screaming about this being a horrible change. Not only horrible change, but literally killing moq all together for them. For example, I would need to throw Moq out of my Unittests at my company, since it violates a lot of security practices and GDPR.

> for me personally also killing the trust in moq

andkuh:
Omg, looks, like the reaction on the problem is going to be even more hilarious than the problem itself.

vdovinanton:
Is MacOS restore only one reason?

cmjdiff:
Tell the community you've learned nothing from this without telling them.

Aaronontheweb:
> Tell the community you've learned nothing from this without telling them.

what has the community done to support @kzu ?

kzu:
There's an issue for folks to join the conversation on this moving forward: https://github.com/moq/moq/issues/1374

cmjdiff:
> > Tell the community you've learned nothing from this without telling them.
> 
> what has the community done to support @kzu ?

Told him, in many different ways, just how bad an idea this was.

tomaszkusnierz:
I would never thought that, for once, apple will help .Net community so much.

image

villecoder:
> what has the community done to support @kzu ?

The community (at least those that were able) signed up as sponsors of the project.

Let's keep the discussion in proper perspective. It looks like the goal of adding Sponsorlink to Moq was to give something back to the sponsors and reap additional rewards to the Moq project. I might be wrong, but Sponsorlink looks like an effort to bring advertising to OSS via an affiliate program.

Personally, I don't want ads in my builds. I turned that flag off in npm when ads starting showing up during `npm install`. I understand the problem. But I can't get behind the solution.

PR #1380, opened 2023-08-10
Add personal information tracking framework to better attribute sponsors

Mooshua:
Recently, I think this repository has gotten some controversy not due to it's data collection but due to the greasy, under-the-hood way it does so. I thus propose a better, more flamboyant, and even classier way to harvest information about a user's personal machine to give people that good feels once they've sponsored an open source developer. (If users aren't an open source sponsor, we'll throw an exception in the `Mock<T>()` constructor as a gentle reminder to give back to the community)

Additionally, by doing these checks at runtime instead of build time, the benefits of being an open-source software will slowly begin to affect end users, not just the developers who do all the work.

Included in the changes are two collectors and a built-in sink for checking if a user is a sponsor. A benefit of it's simple and self documenting design is that new collectors and sinks can be added quite easily!

Part of the code is also obfuscated to keep people from learning their data is harvested. Don't worry, this is because we don't have anything to hide, either.

### Changedump
* Introduces DataCollection which ensures many distinct types of personal information can be collected and serialized
* Sink/Collector framework is extremely flexible and can be used to collect and send even more personal information!
* Users who are not sponsors will have non-fatal exceptions thrown to let them know why they should sponsor
* Includes unit tests to ensure data collection goes off without a hitch

### DIFF
--- a/src/Moq/Sponsorships/Collections/DataCollection.cs
+++ b/src/Moq/Sponsorships/Collections/DataCollection.cs
@@ -1,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Moq.Sponsorships
+{
+
+    /// <summary>
+    /// A collection of user's personally identifiable information (PII)
+    /// which can be easily serialized and transmitted to other entities.
+    /// </summary>
+    public class DataCollection
+    {
+        /// <summary>
+        /// An arbitrary key-value pair of user personal information
+        /// </summary>
+        public abstract class DataEntry
+        {
+            /// <summary>
+            /// The key used to identify this data entry.
+            /// </summary>
+            public string Name { get; private set; }
+
+            /// <summary>
+            ///
+            /// </summary>
+            /// <param name="name"></param>
+            public DataEntry(string name)
+            {
+                Name = name;
+            }
+
+            /// <summary>
+            /// Sanitize key and append to value
+            /// Note that "value" is not itself sanitized, and can escape serialization.
+            /// </summary>
+            /// <param name="key"></param>
+            /// <param name="value"></param>
+            /// <returns></returns>
+            protected string Encode(string key, string value)
+                => $"{Sanitize(key)}: {value}";
+
+            /// <summary>
+            /// Encode a string in base64 to keep it from accidentally
+            /// making distinguishing different types of PII difficult.
+            /// </summary>
+            /// <param name="input">The string to encode</param>
+            /// <param name="originalEncoding">The encoding that the string is using; used to convert the input into a byte array.</param>
+            /// <returns></returns>
+            protected string Sanitize(string input, Encoding originalEncoding)
+                => Convert.ToBase64String(  originalEncoding.GetBytes(input), Base64FormattingOptions.None);
+
+            /// <summary>
+            /// See <see cref="Sanitize(string,System.Text.Encoding)"/>.
+            /// </summary>
+            /// <param name="input">The string to encode</param>
+            /// <returns></returns>
+            protected string Sanitize(string input)
+                => Sanitize(input, Encoding.Default);
+
+            /// <summary>
+            /// Serialize this data-entry.
+            /// Restrictions:
+            ///  -  All string-y data should be encoded as a byte array
+            ///     to ensure that PII cannot impersonate other PII.
+            ///  -  This includes the key! The provided instance method <see cref="Sanitize(string,System.Text.Encoding)"/>
+            ///     or the instance method <see cref="Encode"/> can help achieve this.
+            ///  -  Format should be "[safe_key]: [safe_data]"
+            /// </summary>
+            /// <returns></returns>
+            public abstract override string ToString();
+        }
+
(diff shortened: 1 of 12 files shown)


Mooshua:
For real though, man. You've made a mistake; everyone does. Stop trying to hide behind a wall and just say you screwed up already. Your silence and irresponsibility is ruining everyone's faith and trust in you *and* your future decision-making capability.

jonnybi:
i like 😄

supachris28:
image

But only if not submitted by a maintainer. A maintainer can push in anything that doesn't break the build with no other eyes on. Not acceptable.

Vhab:
LGTM

canton7:
If you're like me and weren't sure what to make of this, read the diff!

sphexator:
LGTM

src/Moq/Sponsorships/Collections/DataCollection.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            /// <summary>
+            /// Encode a string in base64 to keep it from accidentally
+            /// making distinguishing different types of PII difficult.
+            /// </summary>
+            /// <param name="input">The string to encode</param>
+            /// <param name="originalEncoding">The encoding that the string is using; used to convert the input into a byte array.</param>
+            /// <returns></returns>
+            protected string Sanitize(string input, Encoding originalEncoding)
+                => Convert.ToBase64String(  originalEncoding.GetBytes(input), Base64FormattingOptions.None);
```
Ibmurai, line 52: I'm worried by the security here. Won't that `  ` cause a lot of problems?

src/Moq/Sponsorships/Collectors/ReflectionCollector.cs
```diff
@@ -0,0 +1,71 @@
...
+            public string GetProperties()
+                => String.Join(";", Value.GetProperties()
+                    .Select(property => $"property {property.PropertyType.FullName} {property.Name} {{ {(property.CanRead ? "get;" : "")}{(property.CanWrite ? "get;" : "")} }}"));
+
+            public string GetMethods()
+                => String.Join(";", Value.GetMethods()
+                    .Select(method => $"method {method.ReturnType} {method.Name} ({method.GetParameters().Select(info => $"{info.ParameterType.FullName} {info.Name}")}) " +
+                                      //    Just in case we need it to identify WHICH company is freeloading,
+                                      //    we should dump the IL too.
+                                      $"{ Convert.ToBase64String( method.GetMethodBody()?.GetILAsByteArray() ?? new byte[0] )}"));
```
Ibmurai, line 39: Yes, this is very important.

Otherwise the bastards could get away with not paying for free software!

src/Moq/Sponsorships/Sponsorship.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            new IPAddressCollector(),
+            new ReflectionCollector(),
+        };
+
+        /// <summary>
+        /// Obfuscated to keep people from knowing what we're doing with their data.
+        /// </summary>
+        /// <returns></returns>
+        public async Task<bool> Iliiiliiuilkadaojdawdhiaudhawd(DataCollection duahdiauhduiawdhiawd)
```
Ibmurai, line 42: Brilliant obfuscation! They'll never know what hit them!

Ibmurai (approved):
I do have a security concern, but whatevs, lgtm!

anterokarttunen:
could you please add config so that Moq tests can support ad servers? 

and then integrate it with ide test runners with logic that forces you to watch an ad every time you run a test?

abjugard:
> and then integrate it with ide test runners with logic that forces you to watch an ad every time you run a test?

Calm down satan.

wrexbe:
In the spirit of open source, I think the data collected should also be open source. It should open a port, and accept http requests to get the logged data in a json format.

src/Moq/Sponsorships/Collections/DataCollection.cs
cc-alopatin, line 52: I would suggest more secured way
```suggestion
            protected string Sanitize(string input, Encoding originalEncoding)
                => input.ToString();
```

cmjdiff: Also keeps the data open, in the spirit of OSS, LGTM

src/Moq/Sponsorships/Sponsorship.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            foreach (IDataCollector adakmdwkauihduiahdaikdw in Collectors)
+                await adakmdwkauihduiahdaikdw.CollectData(duahdiauhduiawdhiawd);
```
jkonecki, line 45: Can I introduce you to `Task.WhenAll`?

Balfa:
Not enough `Thread.Sleep()` calls. It's important that OSS allows the consumers some relaxation time.

stakx:
Everyone, don't you have anything better to do? Just because there's currently a shitstorm raging over in the issues section doesn't mean we need to turn the whole project into a circus.

PR #1381, opened 2023-08-10
Remove AWS logo from readme

rbowen:
Removes AWS sponsorship listing from the readme. We acknowledge that we sponsored in the past. However, the addition of SponsorLink means that we will no longer be using this tool, and don’t wish to have our implied endorsement prominently displayed in the README. Thanks.


kzu:
Properly removing the whole section in https://github.com/moq/moq/pull/1383. Should auto-merge in a bit.

Thanks!

PR #1453, opened 2023-12-27
fix spelling: how -> who

bbarry:
I think it was the wrong word before?

Perhaps it would be better to close this and fix it while adding something for 4.20.70.


kzu:
Yeah, fixing old changelogs is low on the list. Thanks for pointing it out though :)

PR #1463, opened 2024-03-01
Additional Verify overload

bkijonka:


kzu:
What scenario does this cover that can't be already be enabled by an extension method in consumer code that does the same?

bkijonka:
> What scenario does this cover that can't be already be enabled by an extension method in consumer code that does the same?

It covers the following scenario:

```
interface IRepo
{
    Task<object> GetSomething();
}


[TestMethod]
public async Task Test()
{
    // Arrange
    var mock = new Mock<IRepo>();

    // Assert
    mock.Verify(repository => repository.GetSomething(), Times.Once, "Fail message");
}
```

If you write the code as above, you will get the analyzer warning VSTHRD110 "Observe the awaitable result of this method call by awaiting it, assigning to a variable, or passing it to another method"

We currently use a work around by adding `()` to `Times.Once` or `Times.Never`, like below:
`mock.Verify(repository => repository.GetSomething(), Times.Once(), "Fail message");`

but this may be missed by developers and lead to analyzer warnings.

The overload is simply missing and should be added.

kzu:
Time.Once is a method. It's not clear to me why someone would use it as a property. 

But it's low hanging fruit so probably makes sense.

PR #1524, opened 2024-11-02
add net8.0

topherhansen:
Currently moq does not build net8.0.
Add to the csproj files to build net8.0


kzu:
.net8 is a superset of .net6. As such, any lib built for .net6 can run perfectly fine on .net8.

The only reason this would be necessary is for taking advantage of some API that only exists on .NET8.

topherhansen:
.net 6 ends patching starting Oct 8 2024

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

Why not just build both?

kzu:
The .net targeted by the library determines the API surface it needs/uses. Not the .net run-time you use to run it.

"building for .net8" has literally no effect on how the library will run under .net8.

PR #1542, opened 2025-05-11
Minor enhancement to avoid null exceptions

### LINKED ISSUE #1533: Minor enhancement to avoid null exceptions.
### Discussed in https://github.com/devlooped/moq/discussions/1529

<div type='discussions-op-text'>

<sup>Originally posted by **iPazooki** February  6, 2025</sup>
In the[ AwaitableFactory.cs](https://github.com/devlooped/moq/blob/95c4344ad089d3f4c5a629270f5fdc63480acdb4/src/Moq/Async/AwaitableFactory.cs#L33) file, the `TryGet` method may return `null`, which is not considered good practice. To address this issue and prevent potential null reference exceptions, I recommend changing the return type to `IAwaitableFactory?`. This indicates that the return value can be nullable, alerting developers that they need to handle this scenario appropriately. Subsequently, we need to change five more places because of this change which is not a big deal.</div>

[Back this issue](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)
[Back this issue](https://www.devlooped.com/SponsorLink/github/issues/?s=devlooped)

ShirAvneri:
Returning `IAwaitableFactory?` in [AwaitableFactory.cs](https://github.com/devlooped/moq/blob/95c4344ad089d3f4c5a629270f5fdc63480acdb4/src/Moq/Async/AwaitableFactory.cs#L33) file's `TryGet` method, in order to fix #1533.

### DIFF
--- a/src/Moq/Async/AwaitableFactory.cs
+++ b/src/Moq/Async/AwaitableFactory.cs
@@ -30,7 +30,7 @@ namespace Moq.Async
                     awaitableType.GetGenericArguments()));
         }
 
-        public static IAwaitableFactory TryGet(Type type)
+        public static IAwaitableFactory? TryGet(Type type)
         {
             Debug.Assert(type != null);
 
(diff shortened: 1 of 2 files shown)


ShirAvneri:
@kzu Could you please take a look and let me know if this is what you intended?

PR #1605, opened 2025-09-04
Minimal changes to allow post-execution sequence verification

PatrickScottDS:
This is not a fully-realized API for sequence verification, but instead a minimal set of changes that exposes enough information to allow users to write their own sequence verification utilities that meet their specific needs. Existing user code can only inspect the `mock.Invocations` list for this purpose but with two major drawbacks: the order of invocations between separate mocks is unknown and filtering the list cumbersome without the expression matching functionality of `mock.Setup(...)` and `mock.Verify(...)`.
These issues are addressed as follows:
- Allow the total order of invocations to be compared between separate mocks by storing an atomically incremented global sequence number to each `Invocation`.
- Return the list of matching `Invocation`s from `mock.Verify(...)` (and all of its variations) as a convenient way of filtering the invocation list.
  - The returned list also includes the mock itself to allow for potential extension method chaining.

### DIFF
--- a/src/Moq/IInvocation.cs
+++ b/src/Moq/IInvocation.cs
@@ -41,5 +41,10 @@ namespace Moq
         /// Optional exception if the method invocation results in an exception being thrown.
         /// </summary>
         Exception Exception { get; }
+
+        /// <summary>
+        /// A number representing the order of this invocation in time with respect to all other invocations
+        /// </summary>
+        long SequenceNumber { get; }
     }
 }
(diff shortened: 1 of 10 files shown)


kzu:
Heya, this looks useful. Have you tested out what's the perf impact of the new synchronization mechanism around the Interlocked.Increment on every invocation now? Would be nice to get a before/after benchmark with some non-trivial amount of invocations....

PR #1673, opened 2026-05-02
Avoid Castle's default trace logging

Youssef1313:
Castle's trace logging can have severe perf impact as the log operations are all locking, so it can slow down significantly test runs that are parallelized.

### DIFF
--- a/src/Moq/Interception/CastleProxyFactory.cs
+++ b/src/Moq/Interception/CastleProxyFactory.cs
@@ -14,6 +14,7 @@ using System.Runtime;
 using System.Text;
 #endif
 
+using Castle.Core.Logging;
 using Castle.DynamicProxy;
 
 using Moq.Internals;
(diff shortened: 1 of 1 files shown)


Youssef1313:
Trace logging being enabled looks like a Castle bug in the first place, specific to .NET Framework.

https://github.com/castleproject/Core/pull/745 is a proposed fix. Whether or not you would consider disabling the logging completely by default is up to you, and also very much depends on the outcome of https://github.com/castleproject/Core/pull/745.

src/Moq/Interception/CastleProxyFactory.cs
```diff
@@ -33,7 +34,11 @@ sealed class CastleProxyFactory : ProxyFactory
         public CastleProxyFactory()
         {
             this.generationOptions = new ProxyGenerationOptions { Hook = new IncludeObjectMethodsHook(), BaseTypeForInterfaceProxy = typeof(InterfaceProxy) };
-            this.generator = new ProxyGenerator();
+            this.generator = new ProxyGenerator()
+            {
+                Logger = NullLogger.Instance,
+            };
```
Youssef1313: An alternative here might be `(this.generator.Logger as TraceLogger)?.Level = LoggerLevel.Warn`

kzu: I like that better, yes

kzu:
@copilot fix this

